PR #83, opened 2025-08-08
Fix vertical line height on organization profile page

rarar:
This was a fix to ensure the column border on the left side of the about panel extends to the bottom of the viewport.


apps/app/src/components/Profile/ProfileContent/index.tsx
```diff
@@ -205,7 +205,7 @@ const ProfileGrid = ({ profile }: { profile: Organization }) => {
           <ProfileFeed profile={profile} />
         </Suspense>
       </div>
-      <div className="col-span-6 border-l px-4 py-6">
+      <div className="col-span-6 border-l px-4 py-6 min-h-[50vh]">
```
scazan, line 208: Have you tested this out on an empty profile? It won't really work except in the case that the screen size happens to line up with this.

Image

rarar: Empty profiles were the main place that I tested. Are you noticing issues elsewhere?

scazan: I just pushed what I think should be a more reliable fix.

rarar:
Makes sense. Thanks for doing that!

PR #86, opened 2025-08-09
[Design System] Update `Calendar` styles

metonym:
- Update calendar styles (all cell types, factoring in different states)
- Stylize "current day"
- Add [@internationalized/date dependency](https://github.com/oneprojectorg/common/commit/60305ec7bcffa46926886b41e9d68404a84ef1c4), which is the recommended way to "obtain" the current date

---

Screenshot 2025-08-09 at 1 31 58 PM

PR #88, opened 2025-08-10
[Design System] Re-work `DatePicker`

metonym:
Stacked on #86, alternative to #70

- Depends on the re-worked `Calendar` and usage of `@internationalized/date` dependency

**Implementation Notes**

- TextField accepts free-form input
- Focusing input or clicking calendar icon should open calendar
- Supports initial date
- Supports disabled state

---

Screenshot 2025-08-10 at 12 07 00 PM

Screenshot 2025-08-10 at 12 07 04 PM


packages/ui/src/components/DatePicker.tsx
```diff
@@ -1,60 +1,221 @@
...
+
+  const [inputValue, setInputValue] = useState<string>(initialInputValue);
+
+  const [isCalendarOpen, setIsCalendarOpen] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const parseInputDate = useCallback((input: string): DateValue | null => {
+    try {
+      // Try parsing MM/DD/YYYY format
+      const mmddyyyy = /^(\d{1,2})\/(\d{1,2})\/(\d{4})$/.exec(input.trim());
```
scazan, line 67: I'm worried about this in terms of localization outside of the US as this could cause unintentional user-error. I'm gonna file a P1 issue for it in our project so we can start to use the component already but I think we'll have to think on how this works. In particular with 08/06/25 which is ambiguous (August 6th vs June 8th).

PR #197, opened 2025-10-01
Check for empty string in link tooltip

nourmalaeb:
Tooltips receiving empty strings as children were rendering as empty tooltips. This checks for the empty string and does not render the tooltip if the string is empty.


scazan (approved):
lgtm!

PR #203, opened 2025-10-06
Create profiles as soon as we create a user

scazan:
Creates a profile immediately upon user creation. Also adds a proper slug generation to our system.


services/db/migrations/0016_auto_create_profile_on_signup.sql
```diff
@@ -0,0 +1,57 @@
+CREATE OR REPLACE FUNCTION public.create_user_on_signup()
```
scazan, line 1: @nourmalaeb  FYI

PR #446, opened 2026-01-09
Updating transition scheduling logic

scazan:
Updates the transition scheduling system to properly handle the new DecisionSchemaDefinition template format. Transitions are now created based on phase rules rather than for every phase, and scheduling uses the correct date field of endDate.

If an update occurs, we remove any transitions that are no longer needed UNLESS they have already completed in which case we need to keep them for historical reasons as they were part of a process.

## Test Coverage Summary (19 tests)

  Core Transition Processing (3 tests)
  - ✅ should advance phase when transition is due
  - ✅ should update both currentStateId and currentPhaseId
  - ✅ should mark transition as completed

  Final Phase Transitions (1 test)
  - ✅ should correctly transition to final phase (results)

  Skipping Non-Due Transitions (2 tests)
  - ✅ should not process future transitions
  - ✅ should skip already completed transitions

  Multi-Instance Processing (2 tests)
  - ✅ should process transitions across multiple instances
  - ✅ should process multiple sequential transitions within same instance

  Result Structure (1 test)
  - ✅ should return result object with correct structure

  Transition Creation Rules (2 tests)
  - ✅ should not create transitions for manual advancement phases
  - ✅ should have no pending transitions for instance in final phase

  Deferred State Update Behavior (3 tests)
  - ✅ should update instance state only once after all transitions are processed
  - ✅ should handle orphaned transitions from earlier phases correctly
  - ✅ should process transition even when currentPhaseId does not match fromStateId

  Error Handling (2 tests)
  - ✅ should capture errors and not update state when transition processing fails
  - ✅ should handle successful and failing instances independently

  Instance Status Filtering (3 tests)
  - ✅ should not process transitions for draft instances
  - ✅ should not process transitions for completed instances
  - ✅ should not process transitions for cancelled instances


scazan:
Split out into multiple PRs. Closing this one.

PR #689, opened 2026-02-26
Cache Docker Image in Actions

scazan:


valentin0h (commented):
I've tried a few variations of this. Maybe you have more luck with it, but it usually takes longer than just pulling the images.

valentin0h:
> > I've tried a few variations of this. Maybe you have more luck with it, but it usually takes longer than just pulling the images.
> 
> > I've tried a few variations of this. Maybe you have more luck with it, but it usually takes longer than just pulling the images.
> 
> interesting! I just had a e2e test action that took so long to download that it timed out at about 10 minutes (so generated this to see how it does). Will play with it to see if it improves anything. Good to know you've been down this path already.

Ok 👍 I figured the real impact would be in making the tests run faster as we're adding a lot of them. But if we can have the cache solution be more reliable, even if slightly slower, that's fine. 

I experimented with Supabase branches. That would spin up very quickly, but unfortunately, we'd hit Supabase burst limits while tests are running in parallel.

PR #744, opened 2026-03-10
feat: Commas in the Invite Modal for typed lists

scazan:
Allows manual typing of lists using the comma separator (more can be added later as we think through it). When an email is typed out followed by a comma the email is validated and added into the list of users.


valentin0h (approved):

scazan (commented):
### Fallow findings

Fallow flagged significant duplication (189 lines) between `ProfileInviteModal.tsx` and `ShareProposalModal.tsx` — 9 clone groups. This is pre-existing and out of scope for this PR, but worth a follow-up to extract shared invite/share modal logic.

PR #785, opened 2026-03-16
chore: Add profileUsers Admin tests, Fix Decision Tests with Orgs

scazan:
Adds in test for the admin profileUser that is added via the DB trigger when a new user is created.
This also separates the org from the decision process in our tests since this was unnecessary and not testing the common situation of no org being connected. 

Comments are provided where the new tests are and other files are mostly switching from organization to a profile instead.

Stacked on top of: https://github.com/oneprojectorg/common/pull/784


valentin0h (commented):
Do we still need this?

PR #855, opened 2026-03-31
feat: Add Manual Transitions API

scazan:
Adds in the API endpoint for transitioning between one phase to another manually for an admin.

### DIFF
--- a/packages/common/src/services/decision/manualTransition.ts
+++ b/packages/common/src/services/decision/manualTransition.ts
@@ -13,7 +13,7 @@ import {
 import { getProfileAccessUser } from '../access';
 import { assertUserByAuthId } from '../assert';
 import { advancePhase } from './advancePhase';
-import { processResults } from './processResults';
+import { runResultsProcessing } from './runResultsProcessing';
 import type { DecisionInstanceData } from './schemas/instanceData';
 
 export interface ManualTransitionInput {
@@ -47,9 +47,6 @@ export async function manualTransition({
   user,
   fromPhaseId: expectedFromPhaseId,
 }: ManualTransitionInput): Promise<ManualTransitionResult> {
-  // assertUserByAuthId and the instance fetch are independent — run in parallel
-  // to save one round trip. getProfileAccessUser below still needs to wait on
-  // the instance fetch for profileId.
   const [dbUser, instance] = await Promise.all([
     assertUserByAuthId(user.id),
     db.query.processInstances.findFirst({
@@ -95,10 +92,6 @@ export async function manualTransition({
     throw new CommonError('Instance has no current phase set');
   }
 
-  // Optimistic concurrency check: if the caller declared which phase they
-  // observed, refuse to advance unless the instance is still on that phase.
-  // This prevents double-clicks and stale retries from advancing through
-  // multiple phases unintentionally.
   if (expectedFromPhaseId && expectedFromPhaseId !== fromPhaseId) {
     throw new ConflictError(
       `Instance is on phase '${fromPhaseId}', not '${expectedFromPhaseId}'`,
@@ -115,8 +108,7 @@ export async function manualTransition({
     throw new ValidationError('Already on final phase');
   }
 
-  const nextPhase = phases[currentPhaseIndex + 1]!;
-  const toPhaseId = nextPhase.phaseId;
+  const toPhaseId = phases[currentPhaseIndex + 1]!.phaseId;
   const isTransitioningToFinalPhase =
     currentPhaseIndex + 1 === phases.length - 1;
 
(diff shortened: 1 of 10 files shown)


packages/common/src/services/decision/triggerPhaseAdvancement.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (!dbUser.currentProfileId) {
+    throw new UnauthorizedError('User must have an active profile');
+  }
+
+  if (!instance) {
+    throw new NotFoundError('Process instance not found');
+  }
+
+  if (!instance.profileId) {
+    throw new CommonError(
```
scazan, line 65: WE really need to sort this soon where this is nullable.. 😅 very soon after all this is implemented

packages/common/src/services/decision/manualTransition.ts
```diff
@@ -0,0 +1,147 @@
...
+  const [dbUser, instance] = await Promise.all([
+    assertUserByAuthId(user.id),
+    db.query.processInstances.findFirst({
+      where: { id: instanceId },
+    }),
+  ]);
+
+  if (!dbUser.currentProfileId) {
+    throw new UnauthorizedError('User must have an active profile');
+  }
```
valentin0h:
```suggestion
  if (!dbUser.profileId) {
    throw new UnauthorizedError('User must have an active profile');
  }
```

packages/common/src/services/decision/manualTransition.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Manually advance a decision instance from its current phase to the next phase.
+ * Validates admin access and instance eligibility, then delegates the actual
+ * phase advance to the shared `advancePhase` core function.
+ *
+ * When transitioning to the final phase: additionally runs processResults to
+ * store final selections in the results tables and update proposal statuses.
+ */
+export async function manualTransition({
```
valentin0h: I would rename this to `nextPhaseTransition`.

packages/common/src/services/decision/runResultsProcessing.ts
```diff
@@ -0,0 +1,22 @@
...
+        `Results processing failed for process instance ${processInstanceId}:`,
+        processingResult.error,
+      );
+    }
+  } catch (error) {
+    console.error(
+      `Error processing results for process instance ${processInstanceId}:`,
+      error,
+    );
+  }
```
valentin0h, line 21: Both a `success` and try/catch?

packages/common/src/services/decision/triggerPhaseAdvancement.ts
```diff
@@ -0,0 +1,147 @@
...
+      },
+      fromPhaseId,
+      toPhaseId,
+      triggeredByProfileId: dbUser.currentProfileId,
+      transitionData: { manual: true },
+    });
+
+    if (result.conflict) {
+      throw new ConflictError(
+        'Phase was already advanced, or instance is no longer published',
```
valentin0h: I would leave out the cause unless we're sure those could apply.

packages/common/src/services/decision/manualTransition.ts
```diff
@@ -0,0 +1,147 @@
...
+   * The phase the caller observed as current. If provided, the transition only
+   * proceeds when the instance is actually on this phase. If the instance has
+   * already moved on (e.g. concurrent admin click, retry of a stale request),
+   * a ConflictError is thrown so the caller can refetch and re-decide.
+   */
+  fromPhaseId?: string;
+}
+
+export interface ManualTransitionResult {
+  instanceId: string;
```
valentin0h: Returning this in the output is redundant. `instance` would make sense, though (if we have it)

services/api/src/routers/decision/instances/manualTransition.ts
```diff
@@ -0,0 +1,34 @@
+import { Channels, manualTransition } from '@op/common';
+import { z } from 'zod';
+
+import { commonAuthedProcedure, router } from '../../../trpcFactory';
+
+export const manualTransitionRouter = router({
+  manualTransition: commonAuthedProcedure()
```
valentin0h: What do you think of `transitionToNextPhase` or `nextPhaseTransition` or `transitionFromCurrentPhase`

services/api/src/routers/decision/instances/manualTransition.test.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    const caller = await createAuthenticatedCaller(setup.userEmail);
+
+    await caller.decision.manualTransition({
+      instanceId: instance.instance.id,
+    });
+
+    const history = await db._query.stateTransitionHistory.findFirst({
+      where: eq(stateTransitionHistory.processInstanceId, instance.instance.id),
+    });
```
valentin0h: Use `db.query` if possible.

valentin0h (approved):

services/api/src/routers/decision/instances/manualTransition.ts
scazan: I think `transitionFromPhase` makes sense here given the params and keeping it as an action.

PR #871, opened 2026-04-01
chore: Use Biome

scazan:
Sets up Biome since it is quite a bit faster to run.
This would need to be setup with a follow-up PR that actually does a format so we can save it for when we feel there is a good time to not cause many rebase issues.


valentin0h (commented):
noice!

curious to try if https://oxc.rs/docs/guide/usage/formatter.html would be faster but biome is already an improvement

PR #901, opened 2026-04-07
feat(reviews): add getReviewAssignment API endpoint

valentin0h:
Adds a tRPC query that returns a single authorized review assignment with its proposal snapshot, rubric template, attachments, and saved review draft — needed for the review UI to render the full review pane.


packages/common/src/services/decision/getProposalAttachmentsWithSignedUrls.ts
```diff
@@ -0,0 +1,51 @@
...
+import {
+  type ProposalAttachment,
+  proposalAttachmentSchema,
+} from './schemas/proposal';
+
+/** Fetches proposal attachments with signed URLs for a single proposal. */
+export async function getProposalAttachmentsWithSignedUrls(
+  proposalId: string,
+): Promise<ProposalAttachment[]> {
+  const proposalAttachmentJoins = await db.query.proposalAttachments.findMany({
```
scazan, line 13: This table could potentially be profile attachments. It would make a lot of sense because then this would be much easier to apply to all kinds of other things. Otherwise, we're basically doing duplicated logic all over. 
But I'm not 100% convinced on that, I guess we should consider whether or not profiles/entities should have attachments. It could makes a lot of sense.

scazan (approved):
Lgtm

packages/common/src/services/decision/getProposalAttachmentsWithSignedUrls.ts
valentin0h: Yeah, I think that makes sense, especially considering we need that for processes soon.

PR #902, opened 2026-04-07
feat: hide Proposal Categories tab when organize toggle is off

scazan:
Dynamically hide the Proposal Categories sidebar section when the "Organize proposals into categories" toggle is off, matching the existing Review Rubric toggle pattern. Also hides categories from the launch modal and summary when disabled.


valentin0h (commented):
I'll let @nourmalaeb approve this, but it feels like we should fix this upstream in the prop\s/context we're using.

nourmalaeb (approved):

PR #914, opened 2026-04-08
chore: Factor out phase advance function

scazan:
Refactors the transition engine to use advancePhase instead. The function is a bit more pure and can be used by other services as well such as in manually triggered transitions.


packages/common/src/services/decision/transitionMonitor.ts
```diff
@@ -26,9 +23,28 @@ export interface ProcessDecisionsTransitionsResult {
...
- * This function is called by an external service either scheduled or on a cron job.
+ * Monitors and processes transitions that are due. Called by an external
+ * scheduler / cron worker.
+ *
+ * Each due transition is delegated to the shared `advancePhase` core, so the
+ * cron path writes stateTransitionHistory rows, runs the departing phase's
+ * selection pipeline, and persists surviving proposals into
+ * decisionTransitionProposals — matching the manual transition path.
  */
 export async function processDecisionsTransitions(): Promise<ProcessDecisionsTransitionsResult> {
```
scazan, line 41: entrypoint here

packages/common/src/services/decision/selectionPipeline/types.ts
```diff
@@ -232,35 +232,22 @@ export interface VariableExpression {
...
-
-  // Voting data aggregated per proposal
   voteData?: {
     [proposalId: string]: VoteAggregation;
   };
-
-  // Process instance data
   process: {
     instanceId: string;
     processId: string;
```
valentin0h: Is this used?

packages/common/src/services/decision/advancePhase.ts
```diff
@@ -0,0 +1,212 @@
...
+  /** Shared timestamp for all writes. Defaults to now. */
+  now?: string;
+}
+
+export interface AdvancePhaseResult {
+  /** True when the optimistic lock failed — another writer already advanced. Nothing was written. */
+  conflict: boolean;
+  transitionHistoryId?: string;
+  survivingProposalIds: string[];
+}
```
valentin0h: Do we need stricter typing here? e.g discriminated union on `conflict`

packages/common/src/services/decision/advancePhase.ts
```diff
@@ -0,0 +1,212 @@
...
+    transitionData = {},
+    now = new Date().toISOString(),
+  } = input;
+
+  const instanceId = instance.id;
+  const instanceData = instance.instanceData as DecisionInstanceData;
+
+  const departingPhase = instanceData.phases?.find(
+    (p: PhaseInstanceData) => p.phaseId === fromPhaseId,
+  );
```
valentin0h, line 84: Should we throw here if there's no `departingPhase`?

packages/common/src/services/decision/advancePhase.ts
```diff
@@ -0,0 +1,212 @@
...
+            '{currentPhaseId}',
+            to_jsonb(${toPhaseId}::text)
+          ),
+          '{stateData}',
+          coalesce(${processInstances.instanceData}->'stateData', '{}'::jsonb)
+        ),
+        array['stateData', ${toPhaseId}]::text[],
+        ${JSON.stringify({ enteredAt: now, metadata: transitionData })}::jsonb
+      )`,
+    })
```
valentin0h, line 124: @nourmalaeb fyi regarding instance data vs draft data (and the brittleness of keeping them in sync)

packages/common/src/services/decision/advancePhase.ts
```diff
@@ -0,0 +1,212 @@
...
+  transitionData?: Record<string, unknown>;
+  /** Shared timestamp for all writes. Defaults to now. */
+  now?: string;
+}
+
+export interface AdvancePhaseResult {
+  /** True when the optimistic lock failed — another writer already advanced. Nothing was written. */
+  conflict: boolean;
+  transitionHistoryId?: string;
+  survivingProposalIds: string[];
```
valentin0h: _nit_ not intuitive naming at first read; perhaps add a comment or rename.

valentin0h (approved):
Looks good to me. Mostly reviewed to get context.

packages/common/src/services/decision/advancePhase.ts
scazan: the `survivingProposalIds`? that would maybe make more sense as `selectedProposalIds` actually. Been working to rename all these to be better and more consistent since we will be working with them a lot. Lemme know if you see more.

packages/common/src/services/decision/advancePhase.ts
scazan: yeah, in light of the draft data.. we should be removing this from the json entirely.

packages/common/src/services/decision/selectionPipeline/types.ts
scazan: good catch

PR #916, opened 2026-04-09
fix: centralize process builder autosave to prevent data loss

nourmalaeb:
## Summary

- Centralizes process builder autosave into a shared `ProcessBuilderAutosaveProvider` — replaces duplicated per-section debounce/mutation/save-status logic across 7 components
- Store writes are immediate (survives navigation), API calls are debounced centrally
- Fixes rubric template fields lost when navigating between sections on published processes
- Fixes Review Rubric sidebar tab disappearing when clicked
- Flushes pending saves before launch/update; reverts store on phase delete failure
- Deep-merges `config` in dirty field accumulator to prevent cross-section data loss

## E2E tests

- `edit-published-process.spec.ts` (new): enables review on a phase → verifies sidebar tab persists → adds rubric criterion → navigates away and back → verifies data survived. Same pattern for proposal template fields.
- `create-process-instance.spec.ts` (fixed): scoped Next button selector, added template field to satisfy launch validation.

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx
@@ -1,6 +1,8 @@
+import { ProcessStatus } from '@op/api/encoders';
 import { createClient } from '@op/api/serverClient';
 import { forbidden, notFound } from 'next/navigation';
 
+import { ProcessBuilderAutosaveProvider } from '@/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext';
 import { ProcessBuilderContent } from '@/components/decisions/ProcessBuilder/ProcessBuilderContent';
 import { ProcessBuilderFooter } from '@/components/decisions/ProcessBuilder/ProcessBuilderFooter';
 import { ProcessBuilderHeader } from '@/components/decisions/ProcessBuilder/ProcessBuilderHeader';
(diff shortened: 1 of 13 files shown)


apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
```diff
@@ -108,69 +100,42 @@ function PhaseDetailForm({
...
-      if (isDraft) {
-        updateInstance.mutate({ instanceId, phases: phasesPayload });
-      } else {
-        markSaved(decisionProfileId);
-      }
-    },
-    AUTOSAVE_DEBOUNCE_MS,
-  );
-  debouncedSaveRef.current = () => debouncedSave.isPending();
+  const toPayload = (phases: PhaseDefinition[]) =>
```
scazan: Just generally, since this doesn't rely on any state internal to this function we can pull it out so we don't define it on every render.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx
```diff
@@ -42,37 +44,47 @@ const EditDecisionPage = async ({
     phases: instanceData.phases,
     proposalTemplate:
       instanceData.proposalTemplate as ProcessBuilderInstanceData['proposalTemplate'],
+    rubricTemplate:
+      instanceData.rubricTemplate as ProcessBuilderInstanceData['rubricTemplate'],
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx]: These casts seem really unnecessary because we could just cast incense data on line 37. This makes it a little easier later when we need to properly type this in the DB. You can even cast it earlier when we use the client to get the decision.

apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+export function useProcessBuilderAutosave(): AutosaveActions & {
+  saveState: AutosaveStatus;
+} {
+  const actions = useContext(ActionsContext);
+  if (!actions) {
+    throw new Error(
+      'useProcessBuilderAutosave must be used within ProcessBuilderAutosaveProvider',
+    );
+  }
+  const saveState = useContext(StatusContext);
```
scazan: Can we retitle this since we have `saveChanges` and `saveState`. Maybe better to be either just `state` or processBuilderState or something more descriptive that doesn't sound like a verb.

apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+      // Store the raw promise so flushPendingChanges can detect failure.
+      // Suppress unhandled rejection separately — errors are surfaced by
+      // the mutation's onError callback (toast + status).
+      const promise = updateInstance.mutateAsync({
+        instanceId,
+        ...payload,
+      });
+      inflightRef.current = promise;
+      promise.catch(() => {});
```
scazan: Looks like we're just swallowing the error here. We should probably handle it in some way, but this could also be handled with the mutation's onError.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
```diff
@@ -88,7 +85,7 @@ export function TemplateEditorContent({
   const hasCategories = categories.length > 0;
 
   const initialTemplate = useMemo(() => {
-    const saved = instanceData?.proposalTemplate;
+    const saved = storeData?.proposalTemplate ?? instanceData?.proposalTemplate;
```
scazan, line 88: Maybe not for this PR, but I'm kind of wondering if all this kind of logic here could also be up in the context? Maybe the context needs to fetch the instance rather. Could be a way to simplify a lot of things.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
```diff
@@ -216,44 +194,25 @@ export function TemplateEditorContent({
...
-  }, [template, decisionProfileId, setSaveStatus, debouncedSave]);
+    const normalized = ensureLockedFields(template, {
+      titleLabel: t('Proposal title'),
+      categoryLabel: t('Category'),
+      categories,
+      allowMultipleCategories,
+      requireCategorySelection,
+    });
+    saveChanges({ proposalTemplate: normalized });
+  }, [template]); // eslint-disable-line react-hooks/exhaustive-deps
```
scazan: Since we're not using ES-Lint anymore, I'm thinking this comment probably doesn't need to be there.

scazan (approved):
Approving with some comments! let's get this merged.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
nourmalaeb: Yeah agreed, would simplify a lot.

PR #921, opened 2026-04-09
feat(api): implement draft/live version model

nourmalaeb:
## Summary

Adds two new endpoints alongside the existing `updateDecisionInstance` (which is unchanged):

- **`updateDraftInstanceData`** — process builder autosave writes here. Merges incoming fields into the `draftInstanceData` JSONB column. Live columns are untouched.
- **`publishDecisionInstance`** — promotes `draftInstanceData` to live. Copies the draft blob wholesale to `instanceData`, extracts `name`/`description`/`stewardProfileId` to their own columns. On first publish (DRAFT → PUBLISHED): generates slug, creates transitions, dispatches invites. On subsequent publishes: updates transitions if phases changed.

Also:
- `draftInstanceDataSchema` extends `instanceDataSchema` (unified types)
- `draftInstanceDataEncoder` added to `processInstanceWithSchemaEncoder`
- `createInstanceFromTemplate` and `duplicateInstance` seed `draftInstanceData`
- Integration tests for both new endpoints

`updateDecisionInstance` is **not modified** — it continues to handle status changes, direct instance updates, and other non-builder uses.

## Commits

1. `feat: add draftInstanceData schema and encoder` — Zod schema, tRPC encoder, exports
2. `feat(api): add updateDraftInstanceData endpoint` — new service + route for draft writes
3. `feat(api): add publishDecisionInstance endpoint` — promote draft to live + tests
4. `feat: seed draftInstanceData on instance creation and duplication`
5. `fix: address review findings` — stale docstring, corruption guard, invite error handling, CANCELLED test
6. `test: add updateDraftInstanceData integration tests` — draft-only writes, config merge, auth, corruption guard

## Stack

1. #917 — migration (base)
2. **This PR** — backend
3. #922 — frontend wiring

## Test plan

- [ ] `updateDraftInstanceData` writes to `draftInstanceData` only, live columns unchanged
- [ ] Config merges into existing draft (doesn't replace)
- [ ] Malformed `draftInstanceData` throws instead of silently falling back
- [ ] Non-admin users rejected from draft updates
- [ ] `publishDecisionInstance` copies draft to live columns
- [ ] Slug generated on first publish, stable on subsequent promotes
- [ ] Transitions created on first publish, updated on phase changes
- [ ] COMPLETED and CANCELLED instances reject promotion
- [ ] Non-admin users cannot publish
- [ ] Invite dispatch failure after publish does not surface as error to user
- [ ] New instances and duplicates have `draftInstanceData` seeded
- [ ] `updateDecisionInstance` unchanged — existing tests pass as-is

### DIFF
--- a/packages/common/src/services/decision/schemas/instanceData.ts
+++ b/packages/common/src/services/decision/schemas/instanceData.ts
@@ -3,7 +3,6 @@
  */
 import type { UiSchema } from '@rjsf/utils';
 import type { JSONSchema7 } from 'json-schema';
-import { z } from 'zod';
 
 import { CommonError, ValidationError } from '../../../utils';
 import { schemaValidator } from '../schemaValidator';
@@ -47,6 +46,17 @@ export interface DecisionInstanceData {
   rubricTemplate?: RubricTemplateSchema;
 }
 
+/**
+ * Data stored in the `draftInstanceData` JSONB column.
+ * Same shape as DecisionInstanceData plus instance-column fields that are
+ * extracted to their own columns on publish.
+ */
+export type DraftInstanceData = DecisionInstanceData & {
+  name?: string;
+  description?: string;
+  stewardProfileId?: string;
+};
+
 export interface PhaseOverride {
   phaseId: string;
   name?: string;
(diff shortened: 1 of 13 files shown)


packages/common/src/services/decision/schemas/instanceData.ts
```diff
@@ -59,6 +60,55 @@ export interface PhaseOverride {
   settings?: Record<string, unknown>;
 }
 
+// ============ Zod Schemas ============
+
+/** Zod schema for the `instanceData` JSONB column (loose — tolerates extra fields). */
+export const instanceDataSchema = z.looseObject({
```
scazan: This should be a subset of the DecisionInstance Data.. we can just omit fields or pick fields so it doesn't stray from the other schema.

PR #922, opened 2026-04-09
feat: wire frontend to draft/live model

nourmalaeb:
## Summary

Connects the frontend to the new draft/live endpoints:

- **Edit page** seeds the store from `draftInstanceData` (falls back to `instanceData` for pre-migration instances)
- **Footer** "Update Process" calls `publishDecisionInstance` instead of `updateDecisionInstance` — no payload needed, the server reads from the draft column
- **Launch modal** calls `publishDecisionInstance` with `status: PUBLISHED`
- **DecisionListItem** reads draft name from `draftInstanceData` for drafts (autosave writes there, so the live `name` column may be stale)
- **DuplicateProcessModal** reads draft name for the duplicate naming default

## Stack

1. #917 — migration
2. #921 — backend (base)
3. **This PR** — frontend wiring

## Dependencies

- Assumes #916 (autosave centralization) merges first or concurrently. After #916 merges, the autosave context will need a one-line change: `updateDecisionInstance` → `updateDraftInstanceData` in the debounced save.

## Follow-up (after #916 merges)

- Flatten store to single instance (remove keyed maps)
- Remove hydration provider, seed from autosave provider
- Simplify hooks (drop query fallbacks from useNavigationConfig, useProcessPhases, usePhaseValidation)
- Remove getInstance query fallbacks from section components

## Test plan

- [ ] Edit page shows data from `draftInstanceData` when available
- [ ] "Update Process" promotes draft to live via `publishDecisionInstance`
- [ ] "Launch Process" publishes via `publishDecisionInstance` with status change
- [ ] Draft list items show the draft name, not the stale live name
- [ ] Duplicate modal uses draft name for the default
- [ ] Pre-migration instances (null `draftInstanceData`) fall back to `instanceData`

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/edit/page.tsx
@@ -32,17 +32,19 @@ const EditDecisionPage = async ({
 
   const { processInstance } = decisionProfile;
   const instanceId = processInstance.id;
-  const instanceData = processInstance.instanceData;
+  const { instanceData } = processInstance;
 
-  // Seed the store with server data so validation works immediately.
+  // Seed the store from draftInstanceData (the editable version).
+  // Falls back to instanceData for instances that predate the migration.
+  const draft = processInstance.draftInstanceData;
   const serverData: ProcessBuilderInstanceData = {
-    name: decisionProfile.name ?? undefined,
-    description: processInstance.description ?? undefined,
-    stewardProfileId: processInstance.steward?.id,
-    phases: instanceData.phases,
-    proposalTemplate:
-      instanceData.proposalTemplate as ProcessBuilderInstanceData['proposalTemplate'],
-    config: instanceData.config,
+    name: draft?.name ?? decisionProfile.name ?? undefined,
+    description: draft?.description ?? processInstance.description ?? undefined,
+    stewardProfileId: draft?.stewardProfileId ?? processInstance.steward?.id,
+    phases: draft?.phases ?? instanceData.phases,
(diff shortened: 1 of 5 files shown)


apps/app/src/components/decisions/ProcessBuilder/validation/processBuilderValidation.ts
```diff
@@ -76,7 +76,9 @@ function validateTemplateEditor(
   if (!data?.proposalTemplate) {
     return false;
   }
```
nourmalaeb, line 78: Fixes a bug where the dot for proposal template was rendering when only one category is defined.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/phaseUtils.ts
nourmalaeb, line 1: The entire file exists because the API/store uses `phaseId` as the key and the UI components (`PhaseDefinition` from `@op/api/encoders`) use `id`. Could make them the same but not for this PR

apps/app/src/components/decisions/ProcessBuilder/contentRegistry.tsx
```diff
@@ -14,6 +14,12 @@ import CriteriaSection from './stepContent/rubric/CriteriaSection';
 import ReviewSettingsSection from './stepContent/rubric/ReviewSettingsSection';
 import TemplateEditorSection from './stepContent/template/TemplateEditorSection';
 
+// TODO: Simplify this registry. Most section components no longer need
```
nourmalaeb: Adding this TODO so I don't forget to do this later. In hindsight, I realize this setup was way too much for what we needed.

PR #928, opened 2026-04-09
feat(reviews): wire up API to review screen

valentin0h:
Fetch real proposal and rubric template data via tRPC instead of using hardcoded placeholders, and fix empty dropdown option labels.

No review data yet.


apps/app/src/components/decisions/ReviewExplore/ReviewExploreLayoutClient.tsx
```diff
@@ -19,21 +22,50 @@ interface ReviewExploreLayoutClientProps {
...
+  return (
+    <APIErrorBoundary fallbacks={{ 404: () => notFound() }}>
+      <Suspense fallback={<ReviewExploreSkeleton />}>
+        <ReviewExploreContent slug={slug} reviewId={reviewId} />
+      </Suspense>
+    </APIErrorBoundary>
+  );
+}
+
+function ReviewExploreContent({
```
scazan: We might want to separate this out into a new file with 'use client' while leaving the suspense boundary on the server end (to make better use of PPR generally). Especially with layouts where we can render the layout statically.

scazan (approved):

PR #932, opened 2026-04-10
feat: create TipTap version snapshot on proposal submit

valentin0h:
Snapshots the current TipTap document content as a versioned record when a proposal is submitted, enabling historical document tracking.

### DIFF
--- a/packages/common/src/services/decision/submitProposal.ts
+++ b/packages/common/src/services/decision/submitProposal.ts
@@ -95,7 +95,6 @@ export const submitProposal = async ({
   // created by the DB trigger links back to a concrete document revision.
   const parsed = parseProposalData(existingProposal.proposalData);
 
-  // Step 1: Validate — must pass before we create a version snapshot.
   if (proposalTemplate) {
     await validateProposalAgainstTemplate(
       proposalTemplate,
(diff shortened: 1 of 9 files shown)


packages/common/src/services/decision/submitProposal.ts
```diff
@@ -95,28 +95,29 @@ export const submitProposal = async ({
...
+  }
+
+  // Create a named version snapshot. Best-effort — failures logged, never block.
+  if (!parsed.collaborationDocId) {
+    throw new ValidationError('Proposal is missing a collaboration document');
+  }
+
+  const collaborationDocVersionId = await getTipTapClient()
+    .createVersion(parsed.collaborationDocId, 'Submitted')
+    .then((v) => v.version)
```
scazan, line 113: Any reason we are using the .then() pattern here as opposed to destructing the result?

scazan (approved):

packages/common/src/services/decision/submitProposal.ts
valentin0h: In this case it's helpful to keep the whole operation succint

`const collaborationDocVersionId = await`

PR #938, opened 2026-04-10
chore: Redirect to process page after accepting decision invitation

scazan:
After accepting a decision invitation, redirect users to the decision process page instead of reloading or continuing to onboarding.

## demo
https://github.com/user-attachments/assets/05df6704-5a13-4b90-919d-fbfeb65569ce

### DIFF
--- a/apps/app/src/components/PendingDecisionInvites/index.tsx
+++ b/apps/app/src/components/PendingDecisionInvites/index.tsx
@@ -16,6 +16,7 @@ import {
 } from '@op/ui/NotificationPanel';
 import { ProfileItem } from '@op/ui/ProfileItem';
 import Image from 'next/image';
+import { useRouter } from 'next/navigation';
 import { Suspense } from 'react';
 
 import { useTranslations } from '@/lib/i18n';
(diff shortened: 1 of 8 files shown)


apps/app/src/components/PendingDecisionInvites/index.tsx
```diff
@@ -82,7 +84,15 @@ const PendingDecisionInvitesSuspense = () => {
...
                   className="w-full sm:w-auto"
-                  onPress={() => acceptInvite.mutate({ inviteId: invite.id })}
+                  onPress={() =>
+                    acceptInvite
+                      .mutateAsync({ inviteId: invite.id })
+                      .then(() => {
+                        if (profile.slug) {
+                          router.push(`/decisions/${profile.slug}`);
+                        }
+                      })
```
valentin0h, line 98: Maybe we add a toast for success and error?

valentin0h (approved):

apps/app/src/components/PendingDecisionInvites/index.tsx
scazan: Good idea.. added

PR #940, opened 2026-04-10
chore: Remove last state from json blob

scazan:
Removes the last remaining state from the JSON blob. This makes sure that all state exists on the row and it leaves the JSON blob to simply be configuration of the process.

A follow-up PR should migrate the column name from currentStateId to currentPhaseId.


nourmalaeb (approved):
Nice

nourmalaeb (approved):

PR #944, opened 2026-04-11
feat(admin): add decisions table to platform admin

valentin0h:


apps/app/src/components/screens/PlatformAdmin/DecisionsRow.tsx
```diff
@@ -0,0 +1,136 @@
...
+      </TableCell>
+      <TableCell className="text-sm font-normal text-neutral-charcoal">
+        {decision.status
+          ? (STATUS_DISPLAY[decision.status] ?? decision.status)
+          : '—'}
+      </TableCell>
+      <TableCell className="text-sm font-normal text-neutral-charcoal">
+        {createdAt ? (
+          <TooltipTrigger>
+            <Button className="cursor-default text-sm font-normal underline decoration-dotted underline-offset-2 outline-hidden">
```
scazan: I think these styles are probably built-in to the Button. Same with the tables. Hoping most of the custom styles here can probably just be removed and it looks good/correct.

apps/app/src/components/screens/PlatformAdmin/DecisionsRow.tsx
```diff
@@ -0,0 +1,136 @@
...
+  onOpenChange: (isOpen: boolean) => void;
+}) => {
+  const t = useTranslations();
+
+  return (
+    <Modal isOpen={isOpen} onOpenChange={onOpenChange} isDismissable>
+      <ModalHeader>{t('Instance data for {name}', { name })}</ModalHeader>
+      <ModalBody className="pb-6">
+        <pre className="bg-neutral-gray0 max-h-[60vh] overflow-auto rounded-lg p-4 text-xs">
+          {JSON.stringify(instanceData, null, 2)}
```
scazan, line 129: 👍

scazan, line 129: 👍

apps/app/src/components/screens/PlatformAdmin/DecisionsTable.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <SearchField
+            aria-label={t('Search decisions by name')}
+            placeholder={t('Search decisions by name')}
+            value={searchQuery}
+            onChange={setSearchQuery}
+          />
+        </div>
+      </div>
+      <Suspense fallback={<DecisionsTableSkeleton />}>
+        <DecisionsTableContent searchQuery={debouncedQuery} />
```
scazan, line 43: Full table will suspend when you update filters or paginate. For tables we use .useQuery more often for this reason although we could try to sort a better suspense query approach across the board.

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -0,0 +1,216 @@
...
+};
+
+const phaseLookupInstanceData = z
+  .object({
+    phases: z
+      .array(
+        z.object({
+          phaseId: z.string().optional(),
+          // Legacy name for phaseId; see instanceDataEncoder preprocessor in decision.ts
+          stateId: z.string().optional(),
```
scazan: Let's just remove this? I say we treat those 3 legacy instances entirely different and never add in legacy accomodations into any of our new code.

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (!currentStateId) {
+    return null;
+  }
+
+  const instance = phaseLookupInstanceData.safeParse(instanceData);
+  const schema = phaseLookupProcessSchema.safeParse(processSchema);
+
+  const instancePhase = instance.success
+    ? instance.data.phases?.find(
+        (p) => (p.phaseId ?? p.stateId) === currentStateId,
```
scazan: This can be fine for now just to avoid complexity of shortcircuiting for legacy instances.. but wondering if we just use `isLegacyInstance` and then don't display those 3 here. Keep it simple.

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -0,0 +1,216 @@
...
+          ? ilike(processInstances.name, `%${query}%`)
+          : undefined;
+
+      const whereCondition =
+        searchCondition && cursorCondition
+          ? and(cursorCondition, searchCondition)
+          : searchCondition || cursorCondition;
+
+      const [instances, [totalCountResult]] = await Promise.all([
+        db._query.processInstances.findMany({
```
scazan, line 126: Could we use relations v2?

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -0,0 +1,216 @@
...
+          ? encodeCursor({
+              date: new Date(lastItem.createdAt),
+              id: lastItem.id,
+            })
+          : null;
+
+      const instanceIds = items.map((i) => i.id);
+
+      // Fetch (instanceId, profileId) proposal pairs so we can compute
+      // per-instance proposal count and distinct-participant count in memory.
```
scazan: Thinking about our future upcoming pilot and whether it would be best to compute this on the DB level rather than in memory.

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -0,0 +1,216 @@
...
+
+import { withAuthenticatedPlatformAdmin } from '../../../middlewares/withAuthenticatedPlatformAdmin';
+import withRateLimited from '../../../middlewares/withRateLimited';
+import { commonProcedure, router } from '../../../trpcFactory';
+import { dbFilter } from '../../../utils';
+
+// Drizzle's _query typing widens single `one` relations to `T | T[]` when nested
+// in findMany. See services/api/src/routers/decision/proposals/get.ts:54 for the
+// same workaround.
+const unwrapOne = <T>(value: T | T[] | null | undefined): T | undefined => {
```
scazan, line 19: This is a super common thing to do. Maybe we move that to a util package in core?

services/api/src/routers/platform/admin/listAllUsers.ts
```diff
@@ -100,24 +97,10 @@ export const listAllUsersRouter = router({
...
-              .from(users)
-              .where(searchCondition);
-            return result ?? { value: 0 };
-          },
-          options: {
-            ttl: 1 * 60 * 1000, // 1 min
-            skipMemCache: true,
-          },
-        }),
+        db.select({ value: count() }).from(users).where(searchCondition),
```
scazan, line 100: I wonder if we should filter this by either joining on profiles or checking the onboarded_at column. Just noting that the intention here may or may not be to list the count of all user records but rather, users who have logged into the system successfully, maybe. .

scazan (approved):
No blockers because it's the admin panel but maybe some cleanup on the components and I would start avoiding any legacy data handling.

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
valentin0h: There's some question mark whether this is causing any issues with the timeouts. leaving like this for now before I migrate all of the platform admin ones.

services/api/src/routers/platform/admin/listAllUsers.ts
valentin0h: I'l revisit at a further point

services/api/src/routers/platform/admin/listAllDecisionInstances.ts
valentin0h: I prefer to migrate to `db.query` soon.

apps/app/src/components/screens/PlatformAdmin/DecisionsTable.tsx
valentin0h: yeah let's figure it out across the board because it wasn't straightforward

PR #945, opened 2026-04-11
fix: accept iframely embeds as valid content in required proposal fields

scazan:
## Problem

Required `long-text` proposal fields report "is required" when the only content is an iframely embed (e.g. a pasted YouTube/Vimeo URL that auto-converts). The embed is an atom node with no text children, so the existing text-extraction logic returns an empty string and the field is excluded from validation data.

## Root Cause

Two separate extraction paths both miss iframely atoms:

1. **Client** (`useProposalValidation.ts`): `getBlockText` iterates child `XmlText`/`XmlElement` nodes — iframely atoms have no children, so the function returns `""`.
2. **Server** (`validateProposalAgainstTemplate.ts`): TipTap Cloud called with `format: 'text'`, which also yields `""` for atom nodes.

## Changes

- **`apps/app/src/components/decisions/proposalEditor/useProposalValidation.ts`** — Early-return in `getBlockText` when `element.nodeName === 'iframely'`, returning `attrs.src` so the field is treated as non-empty.
- **`packages/common/src/services/decision/extractTextFromTipTapDoc.ts`** (new) — Pure helper that walks TipTap JSON, handles iframely atoms via `attrs.src`, mirrors the client fix.
- **`packages/common/src/services/decision/validateProposalAgainstTemplate.ts`** — Switch to `format: 'json'` and use the new extractor instead of relying on TipTap Cloud plain-text serialisation.
- **`packages/common/src/services/decision/extractTextFromTipTapDoc.test.ts`** (new) — Unit tests covering empty doc, plain text, iframely-only, and mixed content cases.

## Testing

- All 7 unit tests pass.
- Full typecheck (21 packages) clean.
- Manual: paste embeddable URL into a required field, embed renders, proposal submits without "is required" error. Truly empty required field still blocks submission.

Refs: https://app.asana.com/1/1108348036672214/project/1213315744811204/task/1213980804074933

### DIFF
--- a/apps/app/src/components/decisions/proposalEditor/useProposalValidation.ts
+++ b/apps/app/src/components/decisions/proposalEditor/useProposalValidation.ts
@@ -32,8 +32,11 @@ function getBlockText(element: Y.XmlElement): string {
 }
 
 /**
- * Extracts plain text from a Yjs XmlFragment, matching the output of
- * TipTap Cloud REST API with `format=text`.
+ * Extracts plain text from a Yjs XmlFragment.
+ *
+ * This is the client-side counterpart to `extractTextFromTipTapDoc`
+ * (which operates on the REST API JSON format). Both produce equivalent
+ * plain-text output, including handling of atom nodes like iframely.
  *
  * All fragment content (both TipTap editor fields and scalar values from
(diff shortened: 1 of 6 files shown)


scazan:
branch-length issues with this one. Resubmitting on a new branch

PR #946, opened 2026-04-12
feat(api): add realtime channels for proposal queries and mutations

valentin0h:
Adds realtime channel subscriptions for proposal endpoints so clients receive live updates when proposals are created, updated, deleted, or submitted.

### Demo
https://github.com/user-attachments/assets/80a96855-2799-4515-98b5-7c2f15e7be52

### DIFF
--- a/packages/common/src/services/decision/deleteProposal.ts
+++ b/packages/common/src/services/decision/deleteProposal.ts
@@ -95,7 +95,11 @@ export const deleteProposal = async ({
 
     console.log('DELETED PROPOSAL', deletedProposal.id, user.id);
 
-    return { success: true, deletedId: proposalId };
+    return {
+      success: true,
+      deletedId: proposalId,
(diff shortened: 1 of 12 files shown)


packages/common/src/services/decision/deleteProposal.ts
```diff
@@ -95,7 +95,11 @@ export const deleteProposal = async ({
 
     console.log('DELETED PROPOSAL', deletedProposal.id, user.id);
 
-    return { success: true, deletedId: proposalId };
+    return {
+      success: true,
```
scazan, line 99: Are we getting rid of these `success: true`?

scazan (approved):
🎉

packages/common/src/services/decision/deleteProposal.ts
valentin0h: https://github.com/oneprojectorg/common/pull/952

PR #948, opened 2026-04-12
fix: navigate to correct review rubric section from summary checklist

valentin0h:
Fix summary checklist navigation so review rubric items open the correct section when review_flow is enabled.

### DIFF
--- a/apps/app/src/components/decisions/ProcessBuilder/stepContent/participants/SummarySectionInner.tsx
+++ b/apps/app/src/components/decisions/ProcessBuilder/stepContent/participants/SummarySectionInner.tsx
@@ -6,22 +6,12 @@ import { Header2 } from '@op/ui/Header';
 import { useQueryState } from 'nuqs';
 import React from 'react';
 
+import { useFeatureFlag } from '@/hooks/useFeatureFlag';
 import { useTranslations } from '@/lib/i18n';
 
 import type { SectionProps } from '../../contentRegistry';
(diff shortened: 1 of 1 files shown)


apps/app/src/components/decisions/ProcessBuilder/stepContent/participants/SummarySectionInner.tsx
```diff
@@ -51,6 +41,19 @@ export function SummarySectionInner({
 
   const [, setSectionParam] = useQueryState('section', { history: 'push' });
 
+  const reviewFlowEnabled = useFeatureFlag('review_flow');
+  const rubricSection = reviewFlowEnabled ? 'reviewRubric' : 'criteria';
+  const checklistSectionMap: Record<string, string> = {
```
scazan, line 46: This type could be better.. either as a union or simpler `as const` but not sure why we need to have separate `reviewRubric` and `criteria` here though. Seems like maybe the type is a sign of something weird about that. Not necessarily related to this PR.

scazan (approved):

apps/app/src/components/decisions/ProcessBuilder/stepContent/participants/SummarySectionInner.tsx
valentin0h: I think it's the ff that messes with it. It will be cleaned up soon.

PR #950, opened 2026-04-13
Fix proposal title overflow in editor layout

valentin0h:
Fixes proposal editor header title overflow so long titles no longer overlap adjacent controls.


nourmalaeb:
Truncation works well, but the title is no longer centered in the viewport.

scazan (approved):

scazan (changes requested):
Sorry, hit approve but the centering is a real issue

scazan (approved):

PR #954, opened 2026-04-13
feat: Add Manual Transition into UI

scazan:
Connects the UI for manual transitions if you are an admin.

### demo
Screenshot 2026-04-13 at 13 12 38
Screenshot 2026-04-13 at 14 21 10
Screenshot 2026-04-13 at 13 12 52


apps/app/src/components/decisions/DecisionProcessStepper.tsx
```diff
@@ -26,22 +42,171 @@ export function DecisionProcessStepper({
...
-    sortOrder: phase.phase?.sortOrder,
-  }));
+  const [showConfirmModal, setShowConfirmModal] = useState(false);
+  const isMobile = useMediaQuery(`(max-width: ${screens.sm})`);
+
+  const transitionMutation = trpc.decision.transitionFromPhase.useMutation({
+    onSuccess: () => {
+      setShowConfirmModal(false);
+      toast.success({ message: t('Phase advanced successfully') });
+      router.refresh();
```
valentin0h, line 54: Is this needed?

apps/app/src/components/decisions/DecisionProcessStepper.tsx
```diff
@@ -26,22 +42,171 @@ export function DecisionProcessStepper({
...
+
+  const {
+    nextPhaseId,
+    currentPhaseName,
+    nextPhaseName,
+    currentPhaseAdvancement,
+  } = useMemo(() => {
+    const sorted = [...phases].sort(
+      (a, b) => (a.phase?.sortOrder ?? 0) - (b.phase?.sortOrder ?? 0),
+    );
```
valentin0h: Any reason we haven't made `sortOrder` required?

valentin0h (commented):
Should we feature flag it?

apps/app/src/components/decisions/DecisionProcessStepper.tsx
scazan: It is because the DecisionHeader needs to be invalidated/refreshed. This is likely why we are seeing the two browser situation not showing the invalidation properly. Let's address it separately to handle the server side invalidations (or convert DecisionHeader to 'use client' if necessary.

apps/app/src/components/decisions/DecisionProcessStepper.tsx
scazan: yes! this is a good catch because it's legacy. removing.

valentin0h (approved):

scazan:
> Should we feature flag it?

Nah, I don't think it needs to be feature flagged. It's safe and useful right away.

apps/app/src/components/decisions/DecisionProcessStepper.tsx
valentin0h: Because the decision is server-side fetched? If not, it shouldn't be needed.

PR #956, opened 2026-04-13
feat: Add email notifications for decision process phase transitions

scazan:
Adds a new `phaseTransitioned` Inngest event dispatched whenever a decision process moves between phases that send an email to all participants.

### DIFF
--- a/services/emails/emails/PhaseTransitionEmail.tsx
+++ b/services/emails/emails/PhaseTransitionEmail.tsx
@@ -1,0 +1,46 @@
+import { Button, Heading, Section, Text } from '@react-email/components';
+
+import EmailTemplate from '../components/EmailTemplate';
+
+export const PhaseTransitionEmail = ({
+  processTitle,
+  fromPhaseName,
+  toPhaseName,
+  processUrl = 'https://common.oneproject.org/',
+}: {
+  processTitle: string;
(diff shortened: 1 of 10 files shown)


services/emails/emails/PhaseTransitionEmail.tsx
```diff
@@ -0,0 +1,49 @@
...
+  processUrl = 'https://common.oneproject.org/',
+}: {
+  processTitle: string;
+  toPhaseName: string;
+  phaseNumber: number;
+  totalPhases: number;
+  processUrl: string;
+}) => {
+  return (
+    <EmailTemplate previewText={`${processTitle} — now in ${toPhaseName}`}>
```
nourmalaeb: Not a blocker, but the email's preview text and subject are currently exactly the same. It would be more useful for the preview text to be different. 

Ideas (in pseudocode):
- `The process has from to phase ${phaseNumber} of ${totalPhases}`
- `${phaseNumber} phases down, ${totalPhases - phaseNumber} to go`
- `Log back in to Common to see what's happening`
- `We've reached the ${"voting" | "review" | "final"} phase`

nourmalaeb (approved):
Just a comment about the email preview text being exactly the same as the subject line, which is not really providing any value to the user.

PR #957, opened 2026-04-13
feat: Enable Voting screen for new schema

scazan:
Use capability-based voting detection (phase.rules.voting.submit) instead of string-matching currentStateId against 'voting'. Removes legacy stateId fallback and hardcoded date/slug hack.

Screenshot 2026-04-14 at 11 18 35


apps/app/src/components/decisions/DecisionStateRouter.tsx
```diff
@@ -32,31 +32,41 @@ function DecisionStateRouterNew({
   const [instance] = trpc.decision.getInstance.useSuspenseQuery({ instanceId });
 
   const { currentStateId } = instance;
+  const phases = instance.instanceData?.phases ?? [];
+  const currentPhase = phases.find((p) => p.phaseId === currentStateId);
+  const isVotingEnabled = currentPhase?.rules?.voting?.submit === true;
```
scazan, line 41: Keeping this simple for a moment. If voting is enabled, then we use the voting view.

apps/app/src/components/decisions/pages/VotingPage.tsx
```diff
@@ -24,49 +25,79 @@ export function VotingPage({
...
+    (p) => p.phaseId === currentPhaseId,
   );
+  const currentPhase =
+    currentPhaseIndex >= 0
+      ? (phases[currentPhaseIndex] as InstancePhaseData)
+      : undefined;
+  const nextPhase =
+    currentPhaseIndex >= 0
+      ? (phases[currentPhaseIndex + 1] as InstancePhaseData | undefined)
+      : undefined;
```
valentin0h, line 56: Maybe not for this PR but I think we need to put this logic in the backend.

apps/app/src/components/decisions/pages/VotingPage.tsx
```diff
@@ -24,49 +25,79 @@ export function VotingPage({
...
-    : (instance.description ??
-      instance.instanceData?.templateDescription ??
-      undefined);
-  const aboutIsMarkup = !!instance?.description?.match('PPDESCRIPTION');
+  const resultsDate = nextPhase?.startDate
+    ? new Date(nextPhase.startDate).toLocaleDateString(locale, {
+        month: 'long',
+        day: 'numeric',
+      })
+    : undefined;
```
valentin0h, line 72: if we're assuming that `nextPhase` is a results phase, let's name it further up `resultsPhase`.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -225,12 +227,47 @@ const VotingProposalsList = ({
...
+                        />
+                      ) : undefined
+                    }
+                  />
+                  <ProposalCardMeta proposal={proposal} />
+                  <ProposalCardPreview proposal={proposal} />
+                </ProposalCardContent>
+              </VotingProposalCard>
+            );
+          }
```
valentin0h, line 267: I think a better approach would be to have a new list component for the voting list. We can reuse a lot, but stuffing all here will become unmanageable.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -492,6 +523,8 @@ export const ProposalsList = ({
   initialFilter?: ProposalFilter;
   /** When set to 'results', all proposals are returned as non-editable */
   phase?: 'results';
+  /** When true, the current phase has voting enabled — always show voting UI */
+  isVotingPhase?: boolean;
```
valentin0h, line 527: 🙃

packages/common/src/services/decision/votingEligibility.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Voting eligibility rules — client-safe (no server-only dependencies).
+ * Shared between frontend and backend to keep the definition consistent.
+ */
+
+/** Proposal statuses that are not eligible for voting. */
+export const VOTING_INELIGIBLE_STATUSES: readonly string[] = [
+  'draft',
+  'rejected',
+  'duplicate',
+];
```
valentin0h, line 11: There's two things we should consider here imo: 

1. introduces new types, we have proposal status types already, no?
2. make it more clear what the dependency is, e.g `isVotingEligibleForProposal`

apps/app/src/lib/i18n/dictionaries/en.json
valentin0h, line 1: Are the rest of translations already in place?

apps/app/src/components/decisions/DecisionProcessStepper.tsx
```diff
@@ -44,173 +26,24 @@ export function DecisionProcessStepper({
     [translation],
   );
 
-  const [showConfirmModal, setShowConfirmModal] = useState(false);
-  const isMobile = useMediaQuery(`(max-width: ${screens.sm})`);
-
-  const transitionMutation = trpc.decision.transitionFromPhase.useMutation({
```
valentin0h, line 50: It doesn't seem like we moved it somewhere else, wondering if it was by mistake?

valentin0h (commented):
Did we remove manual transition by mistake?

apps/app/src/components/decisions/ProposalsList.tsx
valentin0h: Commenting that we have both a `phase` and `isVotingPhase`.

apps/app/src/components/decisions/DecisionProcessStepper.tsx
scazan: whoa, sorry. I think the rebase I did right before went wrong.. fixing.

packages/common/src/services/decision/votingEligibility.ts
scazan: yeah, this should use the enum. I'll update

apps/app/src/components/decisions/ProposalsList.tsx
scazan: yeah, we should remove "phase" here as it's mostly used for legacy. I'll look at why we might still be using it.

apps/app/src/components/decisions/ProposalsList.tsx
scazan: I had the same thought when I was self-reviewing.. but then I realized this IS `VotingProposalsList`. Maybe it's clearer in a separate file? Are you thinking it's part of ProposalsList?

apps/app/src/components/decisions/ProposalsList.tsx
scazan: Actually, I need to keep it here for a moment. I need to update one of the old processes that will rely on this still but need to get their go-ahead for it.

valentin0h (approved):

PR #965, opened 2026-04-14
refactor(decisions): pass instance as prop instead of refetching

valentin0h:
Eliminates redundant getInstance queries in decision page components by passing the already-fetched instance from DecisionStateRouter as a prop.

### DIFF
--- a/apps/app/src/components/decisions/pages/ResultsPage.tsx
+++ b/apps/app/src/components/decisions/pages/ResultsPage.tsx
@@ -1,6 +1,8 @@
 'use client';
 
 import { APIErrorBoundary } from '@/utils/APIErrorBoundary';
+import type { RouterOutput } from '@op/api';
+import { trpc } from '@op/api/client';
 import { ProposalFilter } from '@op/api/encoders';
 import { match } from '@op/core';
 import { EmptyState } from '@op/ui/EmptyState';
@@ -22,7 +24,9 @@ import { ProposalListSkeleton, ProposalsList } from '../ProposalsList';
 import { ResultsList } from '../ResultsList';
 import { ResultsStats } from '../ResultsStats';
 
-// Common instance fields used by ResultsPage
+type Instance = RouterOutput['decision']['getInstance'];
+
+// Common instance fields used by ResultsPageContent
 interface ResultsPageInstance {
   id: string;
   name: string;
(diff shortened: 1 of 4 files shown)


nourmalaeb (approved):

apps/app/src/components/decisions/pages/ResultsPage.tsx
```diff
@@ -23,8 +24,11 @@ import { ProposalListSkeleton, ProposalsList } from '../ProposalsList';
 import { ResultsList } from '../ResultsList';
 import { ResultsStats } from '../ResultsStats';
 
-// Common instance fields used by ResultsPage
+type Instance = RouterOutput['decision']['getInstance'];
```
scazan, line 27: Let's use the type from the backend instead since this is already well typed.

apps/app/src/components/decisions/pages/ResultsPage.tsx
```diff
@@ -66,14 +66,10 @@ export function ResultsPage({
   useLegacy?: boolean;
 }) {
   if (useLegacy) {
-    return (
-      <ResultsPageLegacy instanceId={instanceId} profileSlug={profileSlug} />
-    );
+    return <ResultsPageLegacy instance={instance} profileSlug={profileSlug} />;
   }
-  const [instance] = trpc.decision.getInstance.useSuspenseQuery({ instanceId });
```
scazan, line 73: Let's hydrate instead of removing the query here. we can still pass the instance as the initialData instead. This way we get better client-side invalidation behavior and we can use stale times.

Otherwise this will break invalidations currently.

apps/app/src/components/decisions/pages/StandardDecisionPage.tsx
```diff
@@ -29,13 +32,10 @@ export function StandardDecisionPage({
   const t = useTranslations();
   const translation = useDecisionTranslation();
 
-  const [[{ proposals }, instance]] = trpc.useSuspenseQueries((t) => [
-    t.decision.listProposals({
-      processInstanceId: instanceId,
-      limit: 20,
-    }),
-    t.decision.getInstance({ instanceId }),
```
scazan, line 32: Same here.. we don't need to remove it, just give it initialData.

scazan (changes requested):

apps/app/src/components/decisions/pages/ResultsPage.tsx
valentin0h: Yes 👌

PR #968, opened 2026-04-14
feat(reviews): generate review assignments on phase transition

valentin0h:
Automatically generate review assignment rows when a decision instance transitions into a review-capable phase.

feat(reviews): add generateReviewAssignments with REVIEW capability filtering
refactor(decisions): extract onPhaseAdvanced hook to centralise post-advance side-effects

### DIFF
--- a/packages/common/src/services/decision/generateReviewAssignments.ts
+++ b/packages/common/src/services/decision/generateReviewAssignments.ts
@@ -1,11 +1,15 @@
-import { db, eq, inArray } from '@op/db/client';
+import { and, db, eq, inArray, sql } from '@op/db/client';
 import {
+  accessRolePermissionsOnAccessZones,
+  profileUserToAccessRoles,
   profileUsers,
   proposalReviewAssignments,
   proposals,
   users,
 } from '@op/db/schema';
 
+import { CommonError } from '../../utils';
+import { decisionPermission } from './permissions';
 import type { DecisionInstanceData } from './schemas/instanceData';
 import type { ReviewsPolicy } from './schemas/types';
 
@@ -18,12 +22,11 @@ export interface GenerateReviewAssignmentsInput {
 /**
  * Generate review assignment rows for proposals entering a review-capable phase.
  *
- * Assignment strategy depends on the instance's `reviewsPolicy`:
- * - `full_coverage` — every eligible reviewer is assigned every proposal.
- * - `self_selection` — no upfront assignments; reviewers claim proposals themselves.
- * - `random_assignment` — a random subset of reviewers is assigned per proposal (TODO).
+ * Only members with the REVIEW capability on the `decisions` access zone are
+ * eligible. Reviewers are never assigned their own proposals.
  *
- * Reviewers are never assigned their own proposals.
+ * Currently supports the `full_coverage` policy (every eligible reviewer is
+ * assigned every proposal). Throws for unsupported policies.
  */
 export async function generateReviewAssignments({
   instanceId,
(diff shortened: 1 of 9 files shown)


packages/common/src/services/decision/onPhaseAdvanced.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function onPhaseAdvanced(
+  input: OnPhaseAdvancedInput,
+): Promise<void> {
+  const targetPhase = input.phases.find((p) => p.phaseId === input.toPhaseId);
+
+  if (targetPhase?.rules?.proposals?.review) {
+    await runGenerateReviewAssignments(input);
+  }
+
+  const isNowOnFinalPhase =
```
scazan: We have a helper for this `isLastPhase` that we can use here. The more I think of it the more I think the `isLastPhase` should be called `isResultsPhase` which is how it is used. If we don't retitle that in this PR we should at least use it here so we can isolate that logic (eventually changing it to use the capability instead of the phase index)

packages/common/src/services/decision/transitionMonitor.ts
```diff
@@ -194,9 +193,15 @@ async function advanceInstanceTransitions({
         break;
       }
 
-      lastSuccessfulToStateId = transition.toStateId;
       result.processed++;
 
+      await onPhaseAdvanced({
```
scazan, line 198: Since we run this outside the transaction, should we move this into a separate step function to enable retries and better error reporting?

packages/common/src/services/decision/runGenerateReviewAssignments.ts
```diff
@@ -0,0 +1,21 @@
...
+): Promise<void> {
+  try {
+    await generateReviewAssignments({
+      instanceId: input.instanceId,
+      phaseId: input.toPhaseId,
+      selectedProposalIds: input.advanceResult.selectedProposalIds,
+      transitionHistoryId: input.advanceResult.transitionHistoryId,
+    });
+  } catch (error) {
+    console.error(
```
scazan, line 16: If we move this to be called in a step function we might want to throw here instead.

packages/common/src/services/decision/generateReviewAssignments.ts
```diff
@@ -0,0 +1,170 @@
...
+      `generateReviewAssignments: instance ${instanceId} not found`,
+    );
+  }
+
+  const instanceData = instance.instanceData as DecisionInstanceData;
+  const reviewsPolicy = instanceData.config?.reviewsPolicy;
+
+  if (reviewsPolicy && reviewsPolicy !== 'full_coverage') {
+    throw new CommonError(
+      `Review assignment policy '${reviewsPolicy}' is not implemented`,
```
scazan, line 58: If this makes it up to the end-user, maybe `is not supported` is better.

packages/common/src/services/decision/generateReviewAssignments.ts
```diff
@@ -0,0 +1,170 @@
...
+            eq(
+              accessRolePermissionsOnAccessZones.accessZoneId,
+              decisionsZone.id,
+            ),
+          ),
+        )
+        .where(
+          and(
+            eq(profileUsers.profileId, decisionProfileId),
+            sql`(${accessRolePermissionsOnAccessZones.permission} & ${decisionPermission.REVIEW}) != 0`,
```
scazan, line 115: Love this ability to mask at the DB level. Eventually we should pull this out so we can "hide" most of the bit stuff.

scazan (approved):
Nice to see the hooks in place. I would like that the isLastPhase comment is updated before merge though so we can easily replace that logic in the future. Also the step thing I think would help a lot in making it more robust and visible as we could also trigger off a notification to the user (for instance).

packages/common/src/services/decision/transitionMonitor.ts
valentin0h: Good idea, added in Asana

packages/common/src/services/decision/generateReviewAssignments.ts
valentin0h: It won't for now.

PR #988, opened 2026-04-16
feat(reviews): add email notification when revision is requested

valentin0h:
Notify proposal authors via email when a reviewer requests changes, so they can review feedback and submit a revision promptly.

### DIFF
--- a/services/workflows/src/functions/notifications/sendRevisionRequestedNotification.ts
+++ b/services/workflows/src/functions/notifications/sendRevisionRequestedNotification.ts
@@ -1,0 +1,146 @@
+import { OPURLConfig } from '@op/core';
+import { db } from '@op/db/client';
+import {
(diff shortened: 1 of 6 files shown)


services/workflows/src/functions/notifications/sendRevisionRequestedNotification.ts
```diff
@@ -0,0 +1,146 @@
...
+        'Assignment is no longer awaiting revision:',
+        assignmentId,
+        'status:',
+        assignment.status,
+      );
+      return;
+    }
+
+    const { proposal, processInstance } = assignment;
+    const authorEmails = proposal.profile.profileUsers;
```
scazan: nit: not quite authorEmails as much as profile users here

scazan (approved):

PR #994, opened 2026-04-16
Implement overall recommendation in rubric template editor

nourmalaeb:
Adds the toggle for Overall Recommendation in the rubric template editor. The property will have a well-known key in the schema so as to be able to ensure it is always in the correct order (last).

Rendering the field for the review will be added in a follow up PR. 

Preview of the rubric editor with the new Overall Recommendation toggle in place


valentin0h (approved):
lgtm 👍 

we need to fix the UI in the preview and in the rubric form to match designs.

PR #1007, opened 2026-04-17
Review phase experience for non-reviewer

nourmalaeb:
Non-reviewer view for review phase.

- Members route to proposals grid instead of reviewer list.
- Non-owners get Like/Follow (existing `ViewProposalsList` path).
- Decisions-zone admins without REVIEW bit fall through to grid. Profile admins still go to reviewer list via `ALL_TRUE_ACCESS`.
- `APIErrorBoundary` + `Suspense` wraps both branches.
- e2e spec covers routing for both member + admin.

Revise button + "Revision requested" badge deferred until multi-phase proposal card strategy lands.

### DIFF
--- a/apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
+++ b/apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
@@ -94,7 +94,7 @@ export function ProposalCardReviseAction({
 }) {
   const t = useTranslations();
   const href = revisionRequestId
-    ? `${proposalEditHref}?reviewRevision=${revisionRequestId}`
+    ? appendQueryParam(proposalEditHref, 'reviewRevision', revisionRequestId)
     : proposalEditHref;
 
   return (
@@ -104,9 +104,11 @@ export function ProposalCardReviseAction({
   );
 }
 
-/**
- * Badge shown on a proposal card when the author has a pending revision request.
- */
+function appendQueryParam(href: string, key: string, value: string) {
+  const separator = href.includes('?') ? '&' : '?';
+  return `${href}${separator}${key}=${encodeURIComponent(value)}`;
+}
+
 export function RevisionRequestedBadge() {
   const t = useTranslations();
   return (
--- a/apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
+++ b/apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
@@ -8,13 +8,7 @@ import { DialogTrigger } from '@op/ui/Dialog';
 import { Modal, ModalBody, ModalFooter, ModalHeader } from '@op/ui/Modal';
 import { toast } from '@op/ui/Toast';
 import { useState } from 'react';
-import {
-  LuBookmark,
-  LuCircleAlert,
-  LuHeart,
-  LuPencil,
-  LuTrash2,
-} from 'react-icons/lu';
+import { LuBookmark, LuHeart, LuPencil, LuTrash2 } from 'react-icons/lu';
 
 import { useTranslations } from '@/lib/i18n';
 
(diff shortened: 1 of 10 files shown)


apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -1,21 +1,36 @@
...
+  decisionProfileId,
 }: {
-  instance: Instance;
+  instanceId: string;
   decisionSlug: string;
+  slug: string;
+  decisionProfileId?: string | null;
 }) {
+  const t = useTranslations();
+  const [instance] = trpc.decision.getInstance.useSuspenseQuery({ instanceId });
```
valentin0h: Is this needed to fix the real-time invalidations? If so, we need to pass instance data, see @scazan comment

https://github.com/oneprojectorg/common/pull/965#discussion_r3109614586

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -26,6 +41,12 @@ export function ReviewPage({
     throw new Error(`Phase "${currentPhaseId}" not found in instance phases`);
   }
 
+  // Profile admins already get review=true via ALL_TRUE_ACCESS in getInstance,
+  // so this gate covers them too. Decisions-zone admins ("Manage Process")
+  // without explicit review permission fall through to the proposals grid —
+  // they can't actually submit reviews, so the assignments list would be empty.
+  const canReview = instance.access?.review ?? false;
```
valentin0h:
```suggestion
  const canReview = Boolean(instance.access?.review);
```

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -50,10 +71,38 @@ export function ReviewPage({
...
+                      {t("We couldn't load proposals")}
+                    </Header3>
+                    <p className="text-base text-neutral-charcoal">
+                      {t('Please refresh the page to try again.')}
+                    </p>
+                  </EmptyState>
+                ),
+              }}
+            >
+              <Suspense fallback={<ProposalListSkeleton />}>
```
valentin0h: we should lift this one level up

```jsx
<APIErrorBoundary><Suspense>
    review ? <ReviewList>  : <ProposalList>
<APIErrorBoundary><Suspense>
```

apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
```diff
@@ -74,6 +80,47 @@ export function ProposalCardActions({
...
+    <ButtonLink href={href} size="small" className="w-full">
+      {t('Revise proposal')}
+    </ButtonLink>
+  );
+}
+
+function appendQueryParam(href: string, key: string, value: string) {
+  const separator = href.includes('?') ? '&' : '?';
+  return `${href}${separator}${key}=${encodeURIComponent(value)}`;
+}
```
valentin0h: Can we use `nuqs`?

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -50,10 +71,38 @@ export function ReviewPage({
...
+                <ProposalsList
+                  slug={slug}
+                  instanceId={instanceId}
+                  decisionSlug={decisionSlug}
+                  decisionProfileId={decisionProfileId}
+                  permissions={instance.access}
+                  isReviewPhase
```
valentin0h: 🤔 It's best if we don't imply that this is the review phase because we're in the `ReviewPage`.

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -1,21 +1,36 @@
...
-  instance,
+  instanceId,
   decisionSlug,
+  slug,
+  decisionProfileId,
 }: {
-  instance: Instance;
+  instanceId: string;
   decisionSlug: string;
+  slug: string;
```
valentin0h, line 26: This is added because it's needed in ProposalList right? Can you check if that's still needed in the latter? Feels like `decisionSlug` should be all we need.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -153,6 +155,10 @@ interface ProposalsProps {
   hasFilter: boolean;
   /** When true, the current phase has voting enabled — always show voting UI */
   isVotingPhase?: boolean;
+  /** When true, the current phase is the review phase — show Revise action for owners */
+  isReviewPhase?: boolean;
+  /** Map of proposalId → pending revision request id for the current user's proposals */
+  revisionRequestIdByProposalId?: Record<string, string>;
```
valentin0h: Hmm, I see the need, but the coupling of phases and ProposalList will only grow if we go this route. I would suggest we ship without the `Revise` button for MVE until we have a good idea how we're going to show various phase capabilities on one screen. Wdyt?

In that case, most changes in ProposalsList wouldn't be needed in this PR.

valentin0h (commented):
Left a few comments to discuss. 

Nice to get this one in 👍

apps/app/src/components/decisions/pages/ReviewPage.tsx
nourmalaeb: Agree, good call

apps/app/src/components/decisions/ProposalCard/ProposalCardActions.tsx
nourmalaeb: 💯

apps/app/src/components/decisions/ProposalsList.tsx
nourmalaeb: Nice, let's do that. I'll drop the `Revise` button

apps/app/src/components/decisions/pages/ReviewPage.tsx
valentin0h: We can tackle this separately.

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -1,21 +1,37 @@
+'use client';
```
valentin0h: Do we need it? we can use `TranslatedText` if only for translations.

valentin0h (approved):
If we can remove `use client` it'd be great!

apps/app/src/components/decisions/pages/ReviewPage.tsx
nourmalaeb: I always forget about `TranslatedText`. I'll use that.

apps/app/src/components/decisions/pages/ReviewPage.tsx
nourmalaeb: Yeah it's needed in `ProposalsList`... out of scope to change that in this PR.

PR #1009, opened 2026-04-20
feat: Manual Selection of Proposal Advancements Backend

scazan:
Allows for manual selection of proposals between phases whether we have automatic advancement or manual.

This is the backend portion with the frontend portion stacked as #1031


services/api/src/routers/decision/instances/submitManualSelection.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const submitManualSelectionInputSchema = z.object({
+  processInstanceId: z.uuid(),
+  proposalIds: z.array(z.uuid()).min(1),
+});
+
+const submitManualSelectionOutputSchema = z.object({
+  transitionHistoryId: z.uuid(),
+  proposalIds: z.array(z.uuid()),
+});
```
valentin0h: I usually find this type of response not very useful. In cases where we need to use the response, we most certainly need the full object, and in most cases, no response is fine. Just noting it, no change required. My preference would be to remove it.

packages/common/src/services/decision/submitManualSelection.ts
```diff
@@ -0,0 +1,302 @@
...
+ * if the latest transition is no longer into `currentStateId`.
+ */
+export async function submitManualSelection({
+  processInstanceId,
+  proposalIds,
+  user,
+}: SubmitManualSelectionInput): Promise<SubmitManualSelectionResult> {
+  if (proposalIds.length === 0) {
+    throw new ValidationError('At least one proposal must be selected');
+  }
```
valentin0h: Interesting... Is there a benefit to enforcing this?

packages/common/src/services/decision/submitManualSelection.ts
```diff
@@ -0,0 +1,302 @@
...
+      : await tx
+          .select({ id: proposals.id })
+          .from(proposals)
+          .where(
+            and(
+              eq(proposals.processInstanceId, processInstanceId),
+              ne(proposals.status, ProposalStatus.DRAFT),
+              isNull(proposals.deletedAt),
+            ),
+          );
```
valentin0h: `candidateRows` is based on the previous transition attachments, so this feels like a model-level assumption that a phase’s proposal set is fixed at transition time. If that’s not guaranteed, this could drift from the actual current phase set by missing later-added proposals or including ones that were later removed/withdrawn.`

packages/common/src/services/decision/getInstance.ts
```diff
@@ -142,12 +143,21 @@ export const getInstance = async ({ instanceId, user }: GetInstanceInput) => {
...
+      },
+    });
+
     return {
       ...instance,
       instanceData: filteredInstanceData,
       proposalCount,
       participantCount,
       access,
+      selectionsConfirmed: manualSelectionStatus.selectionsConfirmed,
```
valentin0h: *nit*:  not immediately clear if this is a boolean

packages/common/src/services/decision/getManualSelectionState.ts
```diff
@@ -0,0 +1,248 @@
...
+    processInstanceId,
+    instance: {
+      instanceData: instance.instanceData,
+      currentStateId: instance.currentStateId,
+    },
+    dbClient,
+  });
+
+  if (resolved.selectionsConfirmed) {
+    return { selectionsConfirmed: true, proposals: [] };
```
valentin0h: Weird shape, we have confirmed the selection, but hard-coding the proposals?

packages/common/src/services/decision/getManualSelectionState.ts
```diff
@@ -0,0 +1,248 @@
...
+  | { selectionsConfirmed: false; previousPhaseId: string };
+
+/** Boolean-only resolver for `getInstance`; pass `instance` to skip the findFirst. */
+export async function resolveManualSelectionStatus({
+  processInstanceId,
+  instance: preloadedInstance,
+  dbClient = db,
+}: {
+  processInstanceId: string;
+  instance?: {
```
valentin0h: we're always passing it
```suggestion
  instance: {
```

packages/common/src/services/decision/getManualSelectionState.ts
```diff
@@ -0,0 +1,248 @@
...
+  if (isLegacyInstanceData(instance.instanceData)) {
+    return { selectionsConfirmed: true };
+  }
+
+  const currentStateId = instance.currentStateId;
+  if (!currentStateId) {
+    return { selectionsConfirmed: true };
+  }
+
+  const instanceData = instance.instanceData as DecisionInstanceData | null;
```
valentin0h: maybe type it at the function param level

packages/common/src/services/decision/getManualSelectionState.ts
```diff
@@ -0,0 +1,248 @@
...
+  const [latestRow] = await dbClient
+    .select({
+      id: stateTransitionHistory.id,
+      toStateId: stateTransitionHistory.toStateId,
+      transitionData: stateTransitionHistory.transitionData,
+    })
+    .from(stateTransitionHistory)
+    .where(eq(stateTransitionHistory.processInstanceId, processInstanceId))
+    .orderBy(desc(stateTransitionHistory.transitionedAt))
+    .limit(1);
```
valentin0h: Any reason not to use a relational query?

services/api/src/routers/decision/instances/getManualSelectionState.ts
```diff
@@ -0,0 +1,34 @@
...
+  sortOrder: z.enum(['newest', 'oldest']).default('newest'),
+});
+
+const getManualSelectionStateOutputSchema = z.object({
+  selectionsConfirmed: z.boolean(),
+  proposals: z.array(proposalSchema),
+});
+
+export const getManualSelectionStateRouter = router({
+  getManualSelectionState: commonAuthedProcedure()
```
valentin0h: I was thrown a bit by the naming but this feels like a list procedure to me, something like `listManualSelectionCandidates`. We already introduced `selectionsConfirmed` in `getInstance`.

services/api/src/routers/decision/instances/submitManualSelection.ts
```diff
@@ -0,0 +1,33 @@
+import { Channels, submitManualSelection } from '@op/common';
+import { z } from 'zod';
+
+import { commonAuthedProcedure, router } from '../../../trpcFactory';
+
+const submitManualSelectionInputSchema = z.object({
+  processInstanceId: z.uuid(),
+  proposalIds: z.array(z.uuid()).min(1),
```
valentin0h, line 8: Why limit it to min 1?

valentin0h (commented):
I left a comment about renaming one of the procedures to `list...Candidates` caring about any state if possible. `submitManualSelection` should be the one to gate state. I wanted to discuss before approving.

services/api/src/routers/decision/instances/submitManualSelection.ts
scazan: Agreed, nothing was using it. Dropped the output entirely in 872711c30 — both the router and the service return `void` now. The test was reading `result.transitionHistoryId`, but we already fetch the transition row in setup, so I just reused that.

packages/common/src/services/decision/submitManualSelection.ts
scazan: Good call, the schema `.min(1)` already covers it. Dropped the service check in 872711c30 so enforcement lives in one place.

packages/common/src/services/decision/getInstance.ts
scazan: Renamed to `selectionsAreConfirmed` in 872711c30 (and propagated to the frontend in 1a5c62107). Reads as a boolean at the call sites now.

packages/common/src/services/decision/getManualSelectionState.ts
scazan: This one got resolved by the rename below (#3131750660). The list endpoint returns `{ proposals }` only — no more empty-array-alongside-confirmed weirdness.

packages/common/src/services/decision/getManualSelectionState.ts
scazan: Done in 63c5b99b5 — extracted as `InstanceForStatusResolution` and exported from the new `resolveManualSelectionStatus.ts`.

services/api/src/routers/decision/instances/submitManualSelection.ts
scazan: Keeping the `.min(1)` on the schema so callers get a clean Zod error at the tRPC boundary. Paired with #3130654928 — I removed the service-level check, so enforcement lives in one place now.

packages/common/src/services/decision/submitManualSelection.ts
scazan: You were right, and it turned out the same drift existed in `advancePhase` too, so the auto-advance path was silently dropping mid-phase submissions as well. Fixed centrally in `getProposalsForPhase` (2cdce3e63): phase membership is now "attachments of the inbound transition, plus any proposals with a `createdAt` inside the phase window (non-draft, non-deleted)." I went with `createdAt` over `proposalHistory` deliberately — if a rejected proposal gets edited in the next phase, `createdAt` keeps it anchored to the phase it was actually submitted in, whereas a history-based lookup would pick up the new history row and resurface it.

packages/common/src/services/decision/getManualSelectionState.ts
scazan: Made it required in 63c5b99b5 and dropped the inline `findFirst` fallback. Both callers were already passing it, so the fallback was dead code.

packages/common/src/services/decision/getManualSelectionState.ts
scazan: Both the `stateTransitionHistory` fetch and the attachment check are on `dbClient.query.*.findFirst` now, as of 872711c30.

services/api/src/routers/decision/instances/getManualSelectionState.ts
scazan: Renamed to `listSelectionCandidates` in 63c5b99b5 — dropped "Manual" from the name since we could reuse the same endpoint to preview candidates in advance too. The list endpoint does not gate state anymore; `selectionsAreConfirmed` on `getInstance` is the single source of truth for that now. Frontend updated in 90d2b9b40.

PR #1012, opened 2026-04-20
refactor(ui): extract shared SplitPane for proposal/review layouts

valentin0h:
Unifies the three duplicated split-pane layouts behind one component so mobile tab behaviour, scrolling, and padding stay consistent across proposal view, proposal editor, and review. Each pane mounts once and toggles via CSS to keep stateful children (editors, forms) from remounting on tab switch.

https://apps-workshop-git-harmonise-split-pane-views-oneproject.vercel.app/?path=/docs/splitpane--docs

### DIFF
--- a/apps/app/src/components/decisions/proposalEditor/ProposalEditor.tsx
+++ b/apps/app/src/components/decisions/proposalEditor/ProposalEditor.tsx
@@ -11,6 +11,7 @@ import {
   type ProposalTemplateSchema,
   parseProposalData,
 } from '@op/common/client';
+import { SplitPane } from '@op/ui/SplitPane';
 import { toast } from '@op/ui/Toast';
 import type { Editor } from '@tiptap/react';
 import { useLocale } from 'next-intl';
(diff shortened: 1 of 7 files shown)


apps/app/src/components/decisions/proposalEditor/ProposalEditor.tsx
```diff
@@ -433,14 +434,23 @@ function ProposalEditorInner({
...
+            <SplitPane.Pane
+              id="proposal"
+              label={t('Proposal')}
+              className="gap-4"
+            >
               {editorBody}
-            </div>
-            <div className="min-w-0 basis-1/2 overflow-y-auto bg-white">
+            </SplitPane.Pane>
+            <SplitPane.Pane
```
scazan, line 445: I like this more and more and not sure why I haven't always been doing this namespacing. I'm gonna start doing that for other grouped components.

packages/ui/src/components/SplitPane.tsx
```diff
@@ -0,0 +1,135 @@
...
+      if (!pane.props.id || !pane.props.label) {
+        throw new Error(
+          'SplitPane children must have `id` and `label` props (use SplitPane.Pane).',
+        );
+      }
+    }
+  }
+
+  if (panes.length !== 2) {
+    return null;
```
scazan, line 83: Curious question here... what if it is 1? Can imagine dynamic usage where we go from 1 to 2 and that could be nice? Not a blocker though

packages/ui/src/components/SplitPane.tsx
```diff
@@ -0,0 +1,135 @@
...
+        onSelectionChange={handleSelectionChange}
+      >
+        <TabList className="mx-6" variant="default">
+          <Tab id={left.props.id}>{left.props.label}</Tab>
+          <Tab id={right.props.id}>{right.props.label}</Tab>
+        </TabList>
+      </Tabs>
+
+      <div className="flex min-h-0 flex-1 flex-col sm:flex-row">
+        <div
```
scazan: Should this be our built-in `<TabPanel>` from @op/ui? Just noting it goes away from our component library here.

scazan (approved):
great

packages/ui/src/components/SplitPane.tsx
valentin0h: There were accessibility issues with our component afair. I'll have a look

packages/ui/src/components/SplitPane.tsx
valentin0h: Don't have a strong opinion. Feels very edge case at present.

valentin0h: We should evolve this to take any number of panes, with a minimum I think.

valentin0h: updated, takes in 1+ now

PR #1015, opened 2026-04-20
feat(reviews): autosave review drafts as reviewers edit

valentin0h:
Persist in-progress reviewer input continuously so navigating away no longer loses work, and surface the assignment as in-progress as soon as the reviewer starts filling out the rubric.

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
@@ -148,8 +148,6 @@ export function ReviewFormProvider({
   );
 
   const handleSubmit = useCallback(() => {
-    // A trailing autosave landing after submit is filtered out at the DB
-    // level by saveReviewDraft's setWhere guard, so no client-side flush.
     submitReview.mutate({
       assignmentId,
       reviewData: { answers: values, rationales },
--- a/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
@@ -213,7 +213,6 @@ export function ReviewFormProvider({
     // Await pending/in-flight drafts so a trailing autosave can't overwrite
     // the submitted payload on the same client.
     await flushPendingChanges();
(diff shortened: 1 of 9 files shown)


packages/common/src/services/decision/saveReviewDraft.ts
```diff
@@ -0,0 +1,92 @@
...
+export async function saveReviewDraft({
+  assignmentId,
+  reviewData,
+  user,
+}: {
+  assignmentId: string;
+  reviewData: RubricReviewData;
+  user: User;
+}): Promise<{
+  review: typeof proposalReviews.$inferSelect;
```
scazan: Should we just set a type that we can re-use?

packages/common/src/services/decision/submitReview.ts
```diff
@@ -28,7 +24,10 @@ export async function submitReview({
   reviewData: RubricReviewData;
   overallComment?: string | null;
   user: User;
-}): Promise<ProposalReview & { processInstanceId: string }> {
+}): Promise<{
+  review: typeof proposalReviews.$inferSelect;
```
scazan: make a type here I'd say

apps/app/src/components/decisions/Review/ReviewFormContext.tsx
```diff
@@ -196,3 +215,61 @@ export function ReviewFormProvider({
     </ReviewFormContext.Provider>
   );
 }
+
+function useAutosaveDraft({
```
scazan, line 219: We have a few autoSaves now.. let's pull this out into a generic hook with a callback. Can avoid issues in the future where we re-invent it in different ways all over. Maybe either convert the one in the process builder or pull this one out (to eventually use in the process builder)?

scazan (approved):
Nice.. not blocking but we are duplicating a lot of autosave code that would be useful in many other spots.

apps/app/src/components/decisions/Review/ReviewFormContext.tsx
valentin0h: I'll have a look

PR #1016, opened 2026-04-21
feat: Add AccessBoundary component

scazan:
Adds in a AccessBoundary component for setting permission boundaries easier (and more centralized logic for it). Allows for fallbacks.

### DIFF
--- a/apps/app/src/components/AccessBoundary.tsx
+++ b/apps/app/src/components/AccessBoundary.tsx
@@ -1,0 +1,54 @@
+'use client';
+
+import { useUser } from '@/utils/UserProvider';
+import { ReactNode } from 'react';
+
+type Permissions = ReturnType<
+  ReturnType<typeof useUser>['getPermissionsForProfile']
+>;
+
+export type AccessBoundaryCondition = {
+  [Zone in keyof Permissions]?: Partial<Permissions[Zone]>;
+};
+
+export interface AccessBoundaryProps {
+  required: AccessBoundaryCondition | AccessBoundaryCondition[];
+  profileId?: string;
+  fallback?: ReactNode;
+  children: ReactNode;
+}
+
+export function AccessBoundary({
+  required,
+  profileId,
+  fallback = null,
+  children,
+}: AccessBoundaryProps) {
+  const { user, getPermissionsForProfile } = useUser();
+  const targetProfileId = profileId ?? user.currentProfileId;
+
+  if (!targetProfileId) {
+    return <>{fallback}</>;
+  }
+
+  const userPermissions = getPermissionsForProfile(targetProfileId);
+  const conditions = Array.isArray(required) ? required : [required];
+
+  const hasAccess = conditions.some((condition) =>
+    Object.entries(condition).every(([zone, needs]) => {
+      if (!needs) {
+        return true;
+      }
+      const zonePermission = userPermissions[zone as keyof Permissions];
+      if (!zonePermission) {
+        return false;
+      }
+      return Object.entries(needs).every(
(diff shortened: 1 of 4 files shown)


apps/app/src/components/AccessBoundary.tsx
```diff
@@ -0,0 +1,50 @@
+'use client';
+
+import { isPermissionAction, useUser } from '@/utils/UserProvider';
+import type { ZonePermissions } from 'access-zones';
+import { ReactNode } from 'react';
+
+export type AccessBoundaryCondition = {
+  [Zone in keyof ZonePermissions]?: Partial<ZonePermissions[Zone]>;
```
valentin0h, line 8: Is this valid `<AccessBoundary required={{}} ...`

apps/app/src/components/AccessBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export interface AccessBoundaryProps {
+  required: AccessBoundaryCondition | AccessBoundaryCondition[];
```
valentin0h, line 12: I'd name it `condition`

apps/app/src/components/AccessBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+  profileId: string;
+  fallback?: ReactNode;
+  children: ReactNode;
```
valentin0h, line 15: I wonder if we should couple these two with the condition(s). 

Otherwise, how do we do multiple conditions with different fallback/children?

valentin0h (approved):
Nice. 

I would have preferred something like this to start with, feels like a more flexible primitive.
`const hasAccess = useHasAccess({ profileId, required });`

apps/app/src/components/AccessBoundary.tsx
scazan: I think required is more communicative here.. condition is a bit vague and technical.

apps/app/src/components/AccessBoundary.tsx
scazan: ah good idea.. like we do with ApiErrorBoundary? Although I think children should remain as the passing path (feels more intuitive)

scazan:
> Nice.
> 
> I would have preferred something like this to start with, feels like a more flexible primitive. `const hasAccess = useHasAccess({ profileId, required });`

We can update the primitive.. actually that's kinda the point. We have a more idiomatic Access boundary with fallback here but we also centralize the logic so we can update it easily (the primitive is likely to live here mostly). That being said, this is partially in preparation for redoing the way the backend sends FE permissions a little bit. It should be passing back flatter capabilities computed rather than needing the FE to know the structure of what the zones.

This is also partially inspired by Clerk's approach: https://clerk.com/docs/react/reference/components/control/show

apps/app/src/components/AccessBoundary.tsx
scazan: yes and would have expected results (all access cause you didn't add any requirements). But there is a not-ideal side-effect that I will fix here. Mainly that typos might end up with an unexpected issue. I'll update this and won't allow empty permissions.

scazan: Ok, still valid input but will result in denial now.

apps/app/src/components/AccessBoundary.tsx
scazan: I think I'm gonna merge this in as-is for now.. Let's update when we have a use-case as I'm not sure we'll need it yet (and can change it later when we do)

PR #1017, opened 2026-04-21
chore: Remove TRPCError from routers

scazan:


packages/common/src/utils/error/index.ts
```diff
@@ -61,3 +61,14 @@ export class ConflictError extends CommonError {
...
 }
+
+export class RateLimitError extends CommonError {
+  public readonly statusCode: number = 429;
+  public readonly retryAfterSeconds?: number;
+
+  constructor(message?: string, retryAfterSeconds?: number) {
+    const defaultMessage = 'Too many requests. Please try again later.';
+    super(message ?? defaultMessage);
+    this.retryAfterSeconds = retryAfterSeconds;
```
valentin0h: What's this used for?

services/api/src/routers/organization/inviteUser.ts
```diff
@@ -112,26 +111,14 @@ export const inviteUserRouter = router({
           });
         }
       } catch (error) {
-        if (error instanceof TRPCError) {
-          throw error;
-        }
-
-        // Handle specific errors
```
valentin0h: Do we need the catch?

packages/common/src/services/posts/createPostInOrganization.ts
```diff
@@ -90,10 +86,12 @@ export const createPostInOrganization = async (
       allStorageObjects,
     };
   } catch (error) {
+    if (error instanceof CommonError) {
+      throw error;
+    }
```
valentin0h: Do we need the catch?

valentin0h (approved):
nice

packages/common/src/services/posts/createPostInOrganization.ts
scazan: I guess not, it looks more like a conversion of any other errors here. Lemme look and see if we are already handling this in the TRPC client conversion

packages/common/src/utils/error/index.ts
scazan: Just info. I'll remove it.

scazan: Removed.

services/api/src/routers/organization/inviteUser.ts
scazan: Good catch. The service was throwing a raw `Error` so the router was string-matching to coerce it. Fixed at the source — `inviteUsersToOrganization` now throws `UnauthorizedError` directly, and the try/catch is gone.

PR #1018, opened 2026-04-21
feat(reviews): share pending revision request view across reviewers

valentin0h:
Every reviewer on a proposal sees the same paused-for-revision banner and feedback modal when any reviewer has an open revision request, so they have context to stop instead of racing into duplicates. Cancel stays owner-only.

Gating is currently enforced on the client rather than the API because we expect the API may soon need to support multiple concurrent revision requests.

---
  - https://app.asana.com/0/0/1214152232549652

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
@@ -38,7 +38,7 @@ interface ReviewFormState {
   isSubmitted: boolean;
   isPausedForRevision: boolean;
   revisionRequest: ProposalReviewRequest | null;
-  canCancelRevisionRequest: boolean;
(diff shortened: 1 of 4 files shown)


apps/app/src/components/decisions/Review/ReviewFormContext.tsx
```diff
@@ -187,14 +219,14 @@ function ReviewFormProviderInner({
   );
 
   const handleCancelRevision = useCallback(() => {
-    if (!revisionRequest) {
+    if (!ownRequestedRevisionRequest) {
```
nourmalaeb: Nit: could use `if (!canCancelRevisionRequest)` here for clarity

nourmalaeb (approved):

PR #1028, opened 2026-04-22
refactor(reviews): prefetch review page data on the server

valentin0h:
Render the review page with data already available so we can use Next's loading.tsx instead of a client Suspense boundary.

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormContext.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { APIErrorBoundary } from '@/utils/APIErrorBoundary';
 import { trpc } from '@op/api/client';
 import {
   type ProposalReview,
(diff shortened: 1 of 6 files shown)


apps/app/src/components/decisions/Review/ReviewFormContext.tsx
```diff
@@ -54,23 +59,44 @@ export function useReviewForm(): ReviewFormState {
...
   children: ReactNode;
 }) {
   const t = useTranslations();
   const router = useRouter();
+
+  const [reviewAssignment] = trpc.decision.getReviewAssignment.useSuspenseQuery(
+    { assignmentId },
+    // 'always' forces one fetch per mount, which is what registers the
+    // realtime channel via the tRPC client link.
+    { refetchOnMount: 'always' },
```
valentin0h, line 90: I have an idea how to avoid having this (outside of this PR)

nourmalaeb (approved):

PR #1033, opened 2026-04-22
feat(decisions): enforce per-phase voting cap

nourmalaeb:
## Summary
- Add `maxVotes` to phase rules; voting service reads it from the current phase
- Drop the dead instance-level `fieldValues.maxVotesPerMember` fallback (previously hardcoded to 5/3)
- Undefined `maxVotes` = no cap, threaded through validator, tRPC response, and ballot UI

UI picker stacked on this branch: #1034

### DIFF
--- a/apps/app/src/components/decisions/ProposalsList.tsx
+++ b/apps/app/src/components/decisions/ProposalsList.tsx
@@ -181,24 +181,19 @@ const VotingProposalsList = ({
   // Determine voting state
   const hasVoted = voteStatus?.hasVoted || false;
   const isReadOnly = hasVoted || !canVote;
-  const maxVotesPerMember =
-    voteStatus?.votingConfiguration?.maxVotesPerMember || 0;
+  const maxVotesPerMember = voteStatus?.votingConfiguration?.maxVotesPerMember;
 
-  // Handle proposal selection
   const toggleProposal = (proposalId: string) => {
     setSelectedProposalIds((prev) => {
       const isSelected = prev.includes(proposalId);
 
       if (isSelected) {
-        // Remove from selection
         return prev.filter((id) => id !== proposalId);
-      } else {
-        // Add to selection if under limit
-        if (prev.length < maxVotesPerMember) {
-          return [...prev, proposalId];
-        }
-        return prev;
       }
+      if (maxVotesPerMember === undefined || prev.length < maxVotesPerMember) {
+        return [...prev, proposalId];
+      }
+      return prev;
     });
   };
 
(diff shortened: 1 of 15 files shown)


apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -181,24 +181,19 @@ const VotingProposalsList = ({
   // Determine voting state
   const hasVoted = voteStatus?.hasVoted || false;
   const isReadOnly = hasVoted || !canVote;
-  const maxVotesPerMember =
-    voteStatus?.votingConfiguration?.maxVotesPerMember || 0;
+  const maxVotesPerMember = voteStatus?.votingConfiguration?.maxVotesPerMember;
```
scazan, line 182: I think it's a good idea to have a fallback here (like 1 or 0.. maybe 1?)

packages/common/src/services/decision/schemas/types.ts
```diff
@@ -8,6 +8,9 @@ import type { JSONSchema7 } from 'json-schema';
 import type { SelectionPipeline } from '../selectionPipeline/types';
 import type { ProposalTemplateSchema, RubricTemplateSchema } from '../types';
 
+/** Per-participant vote cap. Undefined means unlimited. */
+export type VoteCap = number | undefined;
```
scazan: Seems redundant given that where we use it we already make it optional (resulting in this type anyhow).
`maxVotes?: number` would amount to `number | undefined`

services/api/src/encoders/decision.ts
```diff
@@ -105,6 +106,7 @@ const phaseRulesEncoder = z.object({
     .object({
       submit: z.boolean().optional(),
       edit: z.boolean().optional(),
+      maxVotes: z.number().int().positive().optional(),
```
scazan: I wonder why switch this to maxVotes instead of maxVotesPerMember.. the `VotesPerMember` aspect made sure that we could allow a separate `maxVotes` in the future that would indicate the maximum number of votes.. so best to scope it to "member". We might also have maxVotesPerOrganization or maxVotesPerX.

scazan (commented):
I'm wondering if we should keep the variable naming consistent and scoped `maxVotesPerMember` even in the phase rules. I think the root level `maxVotesPerMember` could potentially go away in the future but serves us fine for determining a default for now.

apps/app/src/components/decisions/ProposalsList.tsx
nourmalaeb: Should there be a fallback? It would make it be an opaque default. According to the designs the default is "No limit" so the fallback is `undefined` which is "No limit" in our system.

services/api/src/encoders/decision.ts
nourmalaeb: 👍 makes sense

packages/common/src/services/decision/schemas/types.ts
nourmalaeb: The idea was to make it explicit since `undefined` translates to "No limit". But maybe too defensive, I'll remove this type if you prefer keeping it simple.

scazan: Maybe we remove the optional then

nourmalaeb: Went the other way — dropped `VoteCap` entirely. Optional fields are `maxVotesPerMember?: number` (which resolves to `number | undefined` as you noted), required sites spell out `number | undefined` inline. Pushed in ef57ff550.

scazan (approved):

PR #1034, opened 2026-04-22
feat(decisions): add per-phase vote limit picker

nourmalaeb:
## Summary
- Adds a Select under the Voting toggle in `PhaseDetailPage` for per-phase vote cap (No limit / 1 / 2 / 3 / 4 / 5 / 10)
- Renders only when voting is enabled for the phase
- Single ICU plural key across en/bn/es/fr/pt for the numeric labels
- Out-of-preset stored `maxVotes` (e.g. seeded via DB) render as an extra option so they remain visible and editable

Stacked on #1033 (`voting-limits-backend`). Merge base will flip to `dev` once that PR lands.


apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
```diff
@@ -442,3 +459,45 @@ function PhaseDetailForm({
     </div>
   );
 }
+
+// 'none' is a reserved option id representing "no limit"; must not collide with String(n).
+const VOTE_LIMIT_OPTIONS = ['none', '1', '2', '3', '4', '5', '10'] as const;
```
scazan, line 487: Any reason we are hardcoding the options? I think it should be a number here and we should be able to allow for 20 votes per member (for instance)

scazan (changes requested):
I don't think we should hardcode the vote limits. It'll simplify some things here too potentially

scazan (approved):

PR #1039, opened 2026-04-23
feat(reviews): render submitted reviews as read-only result cards

valentin0h:
Replaces the editable rubric controls with answer cards once a review is submitted so reviewers can read back what they turned in, including the overall feedback to the author.

## Demo

CleanShot 2026-04-23 at 18 13 27@2x

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
@@ -11,7 +11,7 @@ export function FormShell({ children }: { children: ReactNode }) {
   return (
     <div className="flex flex-col gap-6">
       <div className="border-b border-neutral-gray1 pb-4">
-        <Header3 className="font-serif">
+        <Header3 className="font-serif font-light">
           <TranslatedText text="Review Proposal" />
         </Header3>
       </div>
--- a/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
@@ -11,7 +11,7 @@ export function FormShell({ children }: { children: ReactNode }) {
   return (
     <div className="flex flex-col gap-6">
       <div className="border-b border-neutral-gray1 pb-4">
-        <Header3 className="font-serif !text-title-base font-light">
+        <Header3 className="font-serif">
           <TranslatedText text="Review Proposal" />
         </Header3>
       </div>
--- a/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
@@ -3,17 +3,16 @@ import { Header3 } from '@op/ui/Header';
 import { Surface } from '@op/ui/Surface';
 import type { ReactNode } from 'react';
 
-import { useTranslations } from '@/lib/i18n';
+import { TranslatedText } from '@/components/TranslatedText';
 
 import { getCriteria } from '../rubricTemplate';
 
 export function FormShell({ children }: { children: ReactNode }) {
-  const t = useTranslations();
   return (
     <div className="flex flex-col gap-6">
       <div className="border-b border-neutral-gray1 pb-4">
         <Header3 className="font-serif !text-title-base font-light">
-          {t('Review Proposal')}
+          <TranslatedText text="Review Proposal" />
         </Header3>
       </div>
       {children}
@@ -28,7 +27,6 @@ export function TotalScoreCard({
   rubricTemplate: RubricTemplateSchema;
   values: Record<string, unknown>;
 }) {
-  const t = useTranslations();
   const criteria = getCriteria(rubricTemplate);
 
   const totalScore = criteria.reduce<number | null>((total, criterion) => {
@@ -47,7 +45,7 @@ export function TotalScoreCard({
       className="flex items-start justify-between rounded-lg border-neutral-gray1 p-4"
     >
       <span className="text-base text-neutral-charcoal">
-        {t('Total Score')}
+        <TranslatedText text="Total Score" />
       </span>
       <span className="text-base text-neutral-black">
         {totalScore === null ? '–' : totalScore}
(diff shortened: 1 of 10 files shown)


apps/app/src/components/decisions/Review/ReviewFormShell.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { TranslatedText } from '@/components/TranslatedText';
+
+import { getCriteria } from '../rubricTemplate';
+
+export function FormShell({ children }: { children: ReactNode }) {
+  return (
+    <div className="flex flex-col gap-6">
+      <div className="border-b border-neutral-gray1 pb-4">
+        <Header3 className="font-serif !text-title-base font-light">
+          <TranslatedText text="Review Proposal" />
```
scazan, line 15: I wonder if this is still "Review Proposal" if we are looking at a read-only version of it. If it's read-only we aren't reviewing the proposal as much as looking at a "Proposal Review". I made a note in the figma as well. But I'm also having a hard time finding the Figma attached to this since it doesn't seem to exist under "Review Proposal" (only the actual editor).

Might be more of a design question.

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { TranslatedText } from '@/components/TranslatedText';
+
+import { getCriteria } from '../rubricTemplate';
+
+export function FormShell({ children }: { children: ReactNode }) {
+  return (
+    <div className="flex flex-col gap-6">
+      <div className="border-b border-neutral-gray1 pb-4">
+        <Header3 className="font-serif !text-title-base font-light">
```
scazan: Header3 already includes text-title-base and the font-weight: 300 as well. So only font-serif is needed here.

apps/app/src/components/decisions/Review/ReviewRubricForm.tsx
```diff
@@ -55,24 +51,17 @@ export function ReviewRubricForm() {
...
-  const totalScore = criteria.reduce<number | null>((total, criterion) => {
-    const value = values[criterion.id];
-
-    if (typeof value !== 'number') {
-      return total;
-    }
-
-    return (total ?? 0) + value;
-  }, null);
+  if (review?.state === 'submitted') {
```
scazan: is there an exported enum we can use maybe?

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
```diff
@@ -0,0 +1,55 @@
...
+export function TotalScoreCard({
+  rubricTemplate,
+  values,
+}: {
+  rubricTemplate: RubricTemplateSchema;
+  values: Record<string, unknown>;
+}) {
+  const criteria = getCriteria(rubricTemplate);
+
+  const totalScore = criteria.reduce<number | null>((total, criterion) => {
```
scazan, line 32: This is a much larger question, but is total score simply just a sum of all scores? This could be pretty intense if we have lots. Perhaps it's better to divide by the total criteria so that instead we get a percentage out of one or out of ten or hundred.
Otherwise, the score itself might be pretty useless for end users since we have no particular scale defined.

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+  );
+}
+
+export function TotalScoreCard({
+  rubricTemplate,
+  values,
+}: {
+  rubricTemplate: RubricTemplateSchema;
+  values: Record<string, unknown>;
```
scazan: I feel like this `Record<string, unknown>` is really seeping into our code a lot and I don't think it's necessary that it's unknown. The JSON type in the database wasn't meant to be untyped as much as it is meant to simply not be typed at the database level (although as we mature, we can also do that).

Could we have strict typing over what is possible with the rubrics as well much like the instance data? I know that the proposal data as well as the rubric data are a bit more flexible than instanceData in the sense that you can add a variety of fields but I think we can have a type on it so that we don't end up with `Record<string, unknown>` everywhere. It might be too large of a change for this PR but perhaps we need to solve this before it gets too far with `Record<string, unknown>`.

apps/app/src/components/decisions/Review/SubmittedReviewView.tsx
```diff
@@ -0,0 +1,138 @@
...
+  );
+}
+
+function RubricFieldResult({
+  field,
+  value,
+  rationale,
+}: {
+  field: FieldDescriptor;
+  value: unknown;
```
scazan, line 115: This would be another place where we CAN have a type since we are inferringCriterionType. We already know what is possible.

apps/app/src/components/decisions/Review/SubmittedReviewView.tsx
```diff
@@ -0,0 +1,138 @@
...
+  value,
+  description,
+}: {
+  value?: ReactNode;
+  description?: ReactNode;
+}) {
+  return (
+    <div className="flex items-center gap-4 rounded-xl border border-neutral-gray1 p-6">
+      {value !== undefined && value !== null && value !== '' && (
+        <div className="font-serif text-title-base font-light text-neutral-black">
```
scazan: This looks like a `<Header>`?

scazan (approved):
Approving but I think there is a bit of component cleanup here before merge. Let's discuss the `Record<string, unknown>` concern a bit more if we need to as a team.
My other thought here is that I don't quite understand what this screen is or when you would see it. Maybe you could note how we get here?

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
valentin0h: Yes, that's how the total score was defined for MVP.

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
valentin0h: Made a few adjustments, I can have a more in-depth look at a later stage.

PR #1042, opened 2026-04-23
fix: Use Proper Permission Guard

scazan:
Reverts an AccessBoundary check as we need to treat decisions a little different at the moment.

### DIFF
--- a/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
+++ b/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useUser } from '@/utils/UserProvider';
+import { trpc } from '@op/api/client';
 
 import { useTranslations } from '@/lib/i18n';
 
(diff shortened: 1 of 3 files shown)


valentin0h (changes requested):
we need to use useSuspenseQuery otherwise we're prone to race conditions.

apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
```diff
@@ -1,50 +1,51 @@
...
   const t = useTranslations();
   const navigationConfig = useNavigationConfig(instanceId, decisionProfileId);
 
   const phases = useProcessPhases(instanceId, decisionProfileId);
 
   const { currentSection } = useProcessNavigation(navigationConfig, phases);
 
+  const { data: decisionProfile } = trpc.decision.getDecisionBySlug.useQuery({
+    slug,
+  });
```
valentin0h: we can use `getInstance` no? since we have the `instanceId`

valentin0h (approved):

PR #1048, opened 2026-04-24
refactor(admin): convert platform admin routes to drizzle v2 db.query

valentin0h:
Collapses each admin listing into a single LATERAL-join SQL statement (rather than V1's per-relation-level fan-out), so admin pages issue one round-trip per procedure instead of 6-7.

### DIFF
--- a/services/api/src/routers/platform/admin/listAllDecisionInstances.ts
+++ b/services/api/src/routers/platform/admin/listAllDecisionInstances.ts
@@ -6,22 +6,13 @@ import {
 import { adminDecisionInstanceSchema } from '@op/common/client';
 import { and, count, countDistinct, db, ilike, inArray } from '@op/db/client';
 import { processInstances, proposals } from '@op/db/schema';
+import type { SQL } from 'drizzle-orm';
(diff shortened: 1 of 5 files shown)


services/api/src/routers/platform/admin/listAllDecisionInstances.ts
```diff
@@ -101,30 +92,37 @@ export const listAllDecisionInstancesRouter = router({
...
+                  const conds: SQL[] = [];
+                  if (decodedCursor) {
+                    const cursorCond = getGenericCursorCondition({
+                      columns: { id: table.id, date: table.createdAt },
+                      cursor: decodedCursor,
+                    });
+                    if (cursorCond) conds.push(cursorCond);
+                  }
+                  if (hasSearch) {
+                    conds.push(ilike(table.name, `%${query}%`));
```
scazan, line 120: We should eventually switch this over to `WHERE name ILIKE %${query}% OR ${query} <% name` which is a lot more useful for handling typo tolerance and fuzzier match while maintaining performance

scazan (approved):

PR #1056, opened 2026-04-25
refactor(common): migrate organization + assert services to drizzle v2 query api

valentin0h:
Continues retiring v1 db._query so we can drop the legacy relations and the dual API.

### DIFF
--- a/packages/common/src/services/organization/getOrganization.ts
+++ b/packages/common/src/services/organization/getOrganization.ts
@@ -1,5 +1,5 @@
-import { db, eq, inArray } from '@op/db/client';
-import { profiles, taxonomyTerms } from '@op/db/schema';
+import { db, eq } from '@op/db/client';
+import { profiles } from '@op/db/schema';
 
 import { NotFoundError } from '../../utils';
 
@@ -51,7 +51,11 @@ export const getOrganization = async ({ slug }: { slug: string }) => {
           },
         },
       },
-      strategies: true,
+      strategies: {
+        with: {
+          term: true,
+        },
+      },
     },
   });
 
@@ -59,17 +63,9 @@ export const getOrganization = async ({ slug }: { slug: string }) => {
     throw new NotFoundError('Could not find organization');
   }
 
-  const termIds = org.strategies.map((record) => record.taxonomyTermId);
-  const terms = termIds.length
-    ? await db
-        .select()
-        .from(taxonomyTerms)
-        .where(inArray(taxonomyTerms.id, termIds))
-    : [];
-
   return {
     ...org,
     whereWeWork: org.whereWeWork.map((record) => record.location),
-    strategies: terms,
+    strategies: org.strategies.map((record) => record.term),
   };
 };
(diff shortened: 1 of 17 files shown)


packages/common/src/services/organization/getOrganization.ts
```diff
@@ -45,26 +47,29 @@ export const getOrganization = async ({ slug }: { slug: string }) => {
...
           },
         },
       },
-      // TODO: CONVERT TO TERMS
-      strategies: {
-        with: {
-          term: true,
-        },
-      },
+      strategies: true,
```
scazan: Are we sure we are ok to remove these fields like the term?

packages/common/src/services/organization/getOrganization.ts
```diff
@@ -45,26 +47,29 @@ export const getOrganization = async ({ slug }: { slug: string }) => {
...
   });
 
   if (!org) {
     throw new NotFoundError('Could not find organization');
   }
 
-  org.whereWeWork = org.whereWeWork.map((record) => record.location);
-  org.strategies = org.strategies.map((record) => record.term);
+  const termIds = org.strategies.map((record) => record.taxonomyTermId);
+  const terms = termIds.length
```
scazan: We are making an extra DB call here after removing the join earlier and I assume because of the parentId relation? Do we prefer that over casting the type? I worry that we are shaping our DB calls because of limitations in drizzle that will come up a lot (especially around self-referential items in graphs). Not a blocker but noting it to consider. It smells a bit to me to do that but acknowledging that the @ts-expect-error also smells but that is a drizzle bug.

packages/common/src/services/organization/listOrganizations.ts
```diff
@@ -59,27 +51,29 @@ export const listOrganizations = async ({
...
-                latLng: false,
               },
             },
           },
         },
       },
-      orderBy: (_, { asc, desc }) =>
-        dir === 'asc' ? asc(orderByColumn) : desc(orderByColumn),
+      orderBy: (table, { asc, desc }) => {
+        const col = orderBy === 'createdAt' ? table.createdAt : table.updatedAt;
```
scazan: This was cleaner when we were using getCursorCondition instead? Any reason this had to change?

packages/common/src/services/organization/listOrganizations.ts
```diff
@@ -59,27 +51,29 @@ export const listOrganizations = async ({
...
     if (!result) {
       throw new NotFoundError('Organizations not found');
     }
 
-    result.forEach((org) => {
-      org.whereWeWork = org.whereWeWork.map((item) => item.location);
-    });
+    const flattened = result.map((org) => ({
+      ...org,
+      whereWeWork: org.whereWeWork.map((item) => item.location),
```
scazan, line 78: Just noting that this a less efficient approach. This list won't get so big so it probably doesn't matter.. but it seems like this makes no real change here other than that we copy the objects and increase memory usage.

scazan (approved):

packages/common/src/services/organization/listOrganizations.ts
valentin0h: would need casting if we mutate in place.

packages/common/src/services/organization/getOrganization.ts
valentin0h: Agreed 👍 reverted. Added an task to fix @ts-expect-error

PR #1060, opened 2026-04-25
feat: Decision Scoped Posts

scazan:
This adds in the permission setup for posts which have potentially infinite children that need to be permission-scoped to their root.
This PR is stacked onto #1086. The DB changes were merged in #1069 .

Frontend changes will land separately but my sketched out UI for testing e2e is shown here to provide some context (not representing the final design):
03-overlay-panel-post-and-reaction


packages/common/src/services/reactions/reactionAuth.ts
```diff
@@ -0,0 +1,71 @@
...
+}: {
+  user: { id: string };
+  postId: string;
+}): Promise<{ context: PostContext; profileId: string }> => {
+  const context = await loadPostContext(postId);
+  await assertProfileTypeAccess({
+    user,
+    profileIds: context.associatedProfileIds,
+    policies: {
+      [EntityType.DECISION]: {
```
scazan: not supeeer in love with this pattern necessarily but it is easy to change as we go.

packages/common/src/services/posts/getPost.ts
```diff
@@ -18,76 +19,79 @@ export const getPost = async ({
...
+            }
+          : {}),
+      },
+    }),
+    getCurrentProfileId(authUserId),
+    db
+      .select({ profileId: postsToProfiles.profileId })
+      .from(postsToProfiles)
+      .where(eq(postsToProfiles.postId, postId)),
+  ]);
```
valentin0h: suggestion: if we're going towards the a post can only be associated with a single profile (via root profile id) than this could be `db.query.findFirst`

packages/common/src/services/posts/getPosts.ts
```diff
@@ -1,20 +1,27 @@
...
   maxDepth?: number;
-  authUserId: string; // User ID for authentication
+  authUserId: string;
 }
 
+const postRelations = {
+  profile: { with: { avatarImage: true } },
+  attachments: { with: { storageObject: true } },
+  reactions: { with: { profile: true } },
+} as const;
```
valentin0h: nit: this can live inside `getPosts` if only used there.

packages/common/src/services/posts/getPosts.ts
```diff
@@ -26,153 +33,88 @@ export const getPosts = async (input: GetPostsInput) => {
...
-                        profile: true,
-                      },
-                    },
-                  },
-                },
-              }
-            : {}),
         },
         limit,
         offset,
```
valentin0h: Can we double-check that the pagination is correct here?

packages/common/src/services/posts/listProfilePosts.ts
```diff
@@ -0,0 +1,127 @@
...
+  const decodedCursor = (() => {
+    if (!cursor) {
+      return null;
+    }
+    try {
+      return decodeCursor(cursor);
+    } catch {
+      return null;
+    }
+  })();
```
valentin0h: Feels like we move into `decodeCursor`?

services/api/src/routers/posts/getPosts.ts
```diff
@@ -16,6 +17,18 @@ export const getPosts = router({
         ...input,
         authUserId: ctx.user.id,
       });
+
+      const channels: ChannelName[] = [];
+      if (input.profileId) {
+        channels.push(Channels.profilePosts(input.profileId));
+      }
+      if (input.parentPostId) {
+        channels.push(Channels.postComments(input.parentPostId));
```
valentin0h, line 26: question: I wonder if this is excessive, as it would mean that we'd refetch the list when a reaction has been made on a comment.

packages/common/src/services/posts/postContext.ts
valentin0h, line 1: suggestion: it seems like we're exposing various things here some related to reactions and some not. Worth checking for a similar pattern in the codebase and align how we do context gathering plus asserting access. I don't think we use the term `auth` much.

valentin0h (approved):

PR #1061, opened 2026-04-25
fix: Don't auto-select orgs that you are already a member of when searching for matching orgs

scazan:
This disables organizations that you are already a member of when searching for matching organizations. This avoids the possibility of a ValidationError when we auto-select matching orgs. It also filters organizations you already belong to from the matchingOrganizations endpoint.

Screenshot 2026-04-25 at 23 39 20

### DIFF
--- a/apps/app/src/components/Onboarding/OrganizationSearchScreen.tsx
+++ b/apps/app/src/components/Onboarding/OrganizationSearchScreen.tsx
@@ -213,12 +213,11 @@ function SearchDropdown({
       {searchResults && searchResults.length > 0 ? (
         searchResults.map((org) => {
           const location = getOrgLocation(org);
-          const isMember = org.isCurrentMember;
           return (
             <button
               key={org.id}
               type="button"
-              disabled={isMember}
+              disabled={org.isCurrentMember}
               className="flex w-full items-center px-4 py-3 text-left hover:bg-neutral-offWhite disabled:cursor-not-allowed disabled:opacity-50 disabled:hover:bg-transparent"
               onClick={() => onSelect(org)}
             >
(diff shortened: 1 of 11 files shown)


services/api/src/routers/organization/searchOrganizations.ts
```diff
@@ -1,39 +1,59 @@
...
+import { eq } from 'drizzle-orm';
 import { z } from 'zod';
 
 import { organizationsEncoder } from '../../encoders/organizations';
 import { commonAuthedProcedure, router } from '../../trpcFactory';
 import { dbFilter } from '../../utils';
 
+const searchResultEncoder = organizationsEncoder.extend({
+  isCurrentMember: z.boolean(),
+});
```
valentin0h: Maybe a type `{ org, isMember }` or `{ org, membership }` would be cleaner. But also a good use case for when we search orgs and want to display a user's role in org.

apps/app/src/components/Onboarding/OrganizationSearchScreen.tsx
```diff
@@ -215,7 +217,8 @@ function SearchDropdown({
             <button
               key={org.id}
               type="button"
-              className="flex w-full items-center px-4 py-3 text-left hover:bg-neutral-offWhite"
+              disabled={org.isCurrentMember}
+              className="flex w-full items-center px-4 py-3 text-left hover:bg-neutral-offWhite disabled:cursor-not-allowed disabled:opacity-50 disabled:hover:bg-transparent"
```
valentin0h, line 222: Do we need these extra classes because we are not using `Button`?

apps/app/src/components/Onboarding/OrganizationSearchScreen.tsx
```diff
@@ -90,7 +90,9 @@ export const OrganizationSearchScreen = ({
     onContinue(
       selectedOrgs.flatMap((org) => {
         const profileId = org.profile?.id;
-        if (!profileId) {
+        // Skip orgs the user is already a member of — the join request
+        // backend rejects those, so don't bother dispatching one.
+        if (!profileId || org.isCurrentMember) {
```
valentin0h: I wonder why this isn't a backend check.

valentin0h (approved):

apps/app/src/components/Onboarding/OrganizationSearchScreen.tsx
scazan: It's also checked in the backend.. but trying to join an org you are already a member of throws a validation error rather than a no-op. We avoid handling the validation error by trying not to do it but it seems more correct to keep that validation error in the backend.

PR #1063, opened 2026-04-25
chore(docker): expose api supabase URL + seed Common org for dev stack

scazan:
The docker-compose api service was missing the same NEXT_PUBLIC_SUPABASE_URL
override the app service has, so OTP login from the api container failed
whenever PORT_PREFIX was non-default (the env_file default points at port
3121, but the proxy listens on <PORT_PREFIX>21).

Also adds Common alongside One Project in seedOrgs and a small
seed-orgs-for-test.ts helper that inserts both directly so the join flow
can be exercised end-to-end in the docker dev stack without depending on
the full seed (which has unrelated breakage around supabase admin user
creation and missing seedData/AccessRoles.sql).

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -86,6 +86,7 @@ services:
     env_file: .env.docker
     environment:
       TIPTAP_PRO_TOKEN: ${TIPTAP_PRO_TOKEN}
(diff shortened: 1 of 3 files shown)


nourmalaeb (approved):

scazan:
Unnecessary in the end

PR #1064, opened 2026-04-26
feat: Increase filesize limit to support small videos

scazan:
Increases the base max file size to allow for small videos to be uploaded as well.


valentin0h (approved):
lgtm but we need more changes to support videos

scazan:
> lgtm but we need more changes to support videos

ah, didn't realize that we were restricting videos. This was just meant to up the file size limits. ok, updated so people can add an mp4 to a proposal at least for now (embedding and streaming a video is obviously going to be something different). Will file a separate task to work out how we might want to deal, generally, with filetype support.

PR #1065, opened 2026-04-26
feat(reviews): add admin endpoint with per-proposal review aggregates

valentin0h:
Admin endpoint for the overview/selections screens. Aggregation is server-side because rubric scoring is — clients shouldn't re-derive `averageScore` / `overallRecommendationCount` from raw submissions.

Two dispatch modes:
- **Hydration** (`proposalIds`): overview cards reuse aggregates without a second query path.
- **Paginated** (`createdAt DESC`, phase-scoped): feeds the selections table.


packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,363 @@
...
+    limit: input.limit,
+    cursor: input.cursor,
+    scoredCriterionKeys,
+  });
+}
+
+// ── Filtered mode (caller-given proposalIds) ───────────────────────────
+
+async function listProposalsFiltered({
+  proposalIds,
```
valentin0h, line 116: I think I'll validate this against `getProposalIdsForPhase`

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+  const scoredCriterionKeys = rubricTemplate
+    ? getRubricScoringInfo(rubricTemplate)
+        .criteria.filter((c) => c.scored)
+        .map((c) => c.key)
+    : [];
+
+  const phaseId = input.phaseId ?? instance.currentStateId ?? undefined;
+  const phaseProposalIds = await getProposalIdsForPhase({
+    instanceId: processInstanceId,
+    phaseId,
```
valentin0h, line 90: Is this ok, or should we leave it to the function to define? I'm thinking because I'm still not sure what phase we're in, `review` or `voting`? cc @scazan

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+  z.object({
+    processInstanceId: z.uuid(),
+    phaseId: z.string().optional(),
+    proposalIds: z.array(z.uuid()).min(1),
+  }),
+  z.object({
+    processInstanceId: z.uuid(),
+    phaseId: z.string().optional(),
+    limit: z.number().int().min(1).max(100).default(50),
+    cursor: z.string().optional(),
```
valentin0h, line 47: There's more filtering that will be added separately

packages/common/src/services/decision/schemas/reviews.ts
```diff
@@ -99,6 +99,57 @@ export const proposalRevisionRequestListSchema = z.object({
...
+  reviewers: z.array(
+    z.object({
+      profile: proposalProfileSchema,
+      status: z.enum(ProposalReviewAssignmentStatus),
+    }),
+  ),
+});
+
+/** Taxonomy-backed category attached to a proposal. */
+export const proposalCategorySchema = z.object({
```
scazan, line 130: Don't we already have a type for this and an encoder?

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+  const rubricTemplate = (instance.instanceData.rubricTemplate ??
+    null) as RubricTemplateSchema | null;
+  const scoredCriterionKeys = rubricTemplate
+    ? getRubricScoringInfo(rubricTemplate)
+        .criteria.filter((c) => c.scored)
+        .map((c) => c.key)
+    : [];
+
+  const phaseId = input.phaseId ?? instance.currentStateId ?? undefined;
+  const phaseProposalIds = await getProposalIdsForPhase({
```
valentin0h, line 88: `getProposalIdsForPhase` queries the database one more time for the instance. We can address that separately.

packages/common/src/services/decision/schemas/reviews.ts
```diff
@@ -99,6 +99,57 @@ export const proposalRevisionRequestListSchema = z.object({
...
+ *
+ * `overallRecommendationCount` is a tally of submitted answers to the
+ * well-known overall-recommendation criterion (e.g. `{ yes: 2, no: 1 }`).
+ * Empty when the rubric doesn't include the field or no reviews are in.
+ */
+export const proposalReviewAggregatesSchema = z.object({
+  assignmentsTotal: z.number().int(),
+  reviewsSubmitted: z.number().int(),
+  averageScore: z.number(),
+  overallRecommendationCount: z.record(z.string(), z.number().int()),
```
scazan, line 120: is this always going to be 'yes' | 'no' rather than just 'string'?

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+  const instance = await getInstance({ instanceId: processInstanceId, user });
+
+  if (!instance.access.admin) {
+    throw new UnauthorizedError(
+      "You don't have admin access to this process instance",
+    );
+  }
+
+  const rubricTemplate = (instance.instanceData.rubricTemplate ??
+    null) as RubricTemplateSchema | null;
```
valentin0h:
```suggestion
  const rubricTemplate = instance.instanceData.rubricTemplate
```

packages/common/src/services/decision/schemas/reviews.ts
```diff
@@ -99,6 +99,57 @@ export const proposalRevisionRequestListSchema = z.object({
...
+ * (sum of integer rubric criteria, divided by `reviewsSubmitted`). Returns
+ * 0 when no submissions exist.
+ *
+ * `overallRecommendationCount` is a tally of submitted answers to the
+ * well-known overall-recommendation criterion (e.g. `{ yes: 2, no: 1 }`).
+ * Empty when the rubric doesn't include the field or no reviews are in.
+ */
+export const proposalReviewAggregatesSchema = z.object({
+  assignmentsTotal: z.number().int(),
+  reviewsSubmitted: z.number().int(),
```
scazan: I'm having a little trouble just looking at this somewhat blindly and knowing what these are. I think it would help to be able to do that. So below we have overallRecommendationCount and here we have reviewsSubmitted. It would be nice if there was some consistentcy. Is this a count of reviews that were submitted? In which case we should say reviewsSubmittedCount or something similar.

packages/common/src/services/decision/schemas/reviews.ts
```diff
@@ -99,6 +99,57 @@ export const proposalRevisionRequestListSchema = z.object({
...
+ * `averageScore` is the mean of per-review scores across submitted reviews
+ * (sum of integer rubric criteria, divided by `reviewsSubmitted`). Returns
+ * 0 when no submissions exist.
+ *
+ * `overallRecommendationCount` is a tally of submitted answers to the
+ * well-known overall-recommendation criterion (e.g. `{ yes: 2, no: 1 }`).
+ * Empty when the rubric doesn't include the field or no reviews are in.
+ */
+export const proposalReviewAggregatesSchema = z.object({
+  assignmentsTotal: z.number().int(),
```
scazan: This naming I'm also questioning. is this a count of the assignments or is this the total score for the assignments? I'll probably find out later when I look at the rest of the code in this PR but flagging that it's not immediately obvious and maybe could be.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
scazan: We want to cast as close to the DB query as possible for a single consistent cast that is easily updated later. getInstance already returns the correct type so we shouldn't have to cast it here. If there is an issue around needing to cast this it needs to be solved further up. We can't treat instanceData as untyped JSON in the codebase itself.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+
+// ── Input schema ───────────────────────────────────────────────────────
+
+/**
+ * Single union schema for both dispatch modes:
+ *   - filtered: caller passes `proposalIds`, no pagination.
+ *   - paginated: phase-scoped, cursor-paginated.
+ */
+export const listProposalsWithReviewAggregatesInputSchema = z.union([
+  z.object({
```
scazan, line 38: I'm curious why we're not just supporting pagination on both?

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
scazan: Is that what you are doing  in filteredProposalIds?

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+    loadCategoriesByProposalIds(filteredProposalIds),
+  ]);
+
+  const proposalsById = new Map(proposalsFull.map((p) => [p.id, p]));
+
+  // Drop unmatched (cross-instance, deleted, etc). Preserve caller order.
+  const items = filteredProposalIds
+    .map((id) => proposalsById.get(id))
+    .filter((p): p is NonNullable<typeof p> => p !== undefined)
+    .filter((p) => p.processInstanceId === processInstanceId)
```
scazan: Why not do this in the query itself so we don't overfetch? (plus would allow for pagination)

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+            getCursorCondition({
+              column: table.createdAt,
+              tieBreakerColumn: table.id,
+              cursor: decodedCursor,
+              direction: 'desc',
+            }),
+        }),
+      },
+      with: proposalRelations({ processInstanceId, phaseId }),
+      orderBy: { createdAt: 'desc', id: 'desc' },
```
scazan, line 201: We should just use standard pagination utils here so we can support sorting

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+      where: assignmentWhere,
+      with: {
+        reviewer: { with: { avatarImage: true } },
+        reviews: true,
+      },
+    },
+  } as const;
+}
+
+async function loadCategoriesByProposalIds(
```
scazan: Feels more like a `getCategoriesByProposalIds`

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * Per-proposal review aggregates computed from the loaded review assignments.
+ * Duck-typed input — only the fields the function actually reads — so callers
+ * can pass the relational query result directly without a named type.
+ *
+ * `proposal_reviews_assignment_unique` makes `reviews` 0-or-1; we read just
+ * the first row even though the relation is declared as many.
+ */
+function computeReviewAggregates(
```
scazan: `getComputedReviewAggregates` maybe? Trying to differentiate from non-pure functions.. without the get it feels like it could also be mutating something. The get makes it clearer.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -0,0 +1,377 @@
...
+      const recommendationKey = String(recommendation);
+      overallRecommendationCount[recommendationKey] =
+        (overallRecommendationCount[recommendationKey] ?? 0) + 1;
+    }
+  }
+
+  const averageScore =
+    reviewsSubmitted === 0 ? 0 : totalScore / reviewsSubmitted;
+
+  return {
```
scazan, line 361: This is nice and straightforward. What I'm wondering is how we support more complex and potentially user-defined shapes here. Maybe we need to pass in a shaping function or something along those lines? Maybe a weighted scoring would be more common than a shaped scoring (such as using a bell curve). Not for this PR but important to leave room for.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
scazan: (discussed in our team chat)

scazan (approved):
Only one real blocker which is that we are consistently doing these casts and we really need to avoid them: https://github.com/oneprojectorg/common/pull/1065/changes#r3159821031

We should treat the JSON fields as strictly typed fields coming out of the DB.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: yes indeed.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: Yes it's coming separately, there's nothing input for that yet.

packages/common/src/services/decision/schemas/reviews.ts
valentin0h: For the time being, we control its type. We can restrict it here. I'll see if I do it.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: The use case here is an existing list of proposals that requires extra review of aggregate data. You wouldn't want to paginate in that case since the first list is what's paginated. We could add pagination if there's a good case for it.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: This is odd; the proposal ID is a primary key. Removing.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: yep!

PR #1066, opened 2026-04-26
feat(reviews): add reviewer tooltip to admin review progress cards

valentin0h:
Lets admins see which reviewers have submitted per proposal directly in the grid, without drilling into each card.


apps/app/src/components/decisions/ReviewAssignmentCard.tsx
```diff
@@ -60,11 +69,37 @@ export function ReviewAssignmentCard({
...
+
+  if (completedReviewers.length === 0) {
+    return null;
+  }
+
+  const names = completedReviewers.map((r) => r.profile.name).join(', ');
+
+  return (
+    <TooltipTrigger>
+      <Button className="text-neutral-gray4 underline decoration-dotted underline-offset-2">
```
scazan: This shouldn't be a button for accessibility reasons since it is not an action. We need to change it to a span or a focusable in react-aria.

apps/app/src/components/decisions/ReviewAssignmentsList.tsx
```diff
@@ -49,6 +51,18 @@ export function ReviewAssignmentsList({
   });
 
   const assignments = data?.assignments ?? [];
+  const proposalIds = assignments.map((a) => a.assignment.proposal.id);
+
+  const { data: aggregatesData } =
+    trpc.decision.listWithReviewAggregates.useQuery(
```
scazan, line 57: Should we handle errors here or silently ignore?

scazan (approved):

apps/app/src/components/decisions/ReviewAssignmentCard.tsx
```diff
@@ -60,11 +69,37 @@ export function ReviewAssignmentCard({
...
-      <ReviewStatusBadge status={status} />
+      <div className="flex items-center justify-between gap-2">
+        <ReviewStatusBadge status={status} />
+        {reviewers ? <ReviewersTooltip reviewers={reviewers} /> : null}
+      </div>
     </ProposalCard>
   );
 }
 
+function ReviewersTooltip({ reviewers }: { reviewers: Reviewers }) {
```
scazan, line 81: The naming of this function is confusing too. It's not a tooltip so much as a display of the counts that is wrapped in a tooltip.

apps/app/src/components/decisions/ReviewAssignmentsList.tsx
valentin0h: It's ok to ignore in this case. The ux won't suffer and we should get a log

PR #1068, opened 2026-04-27
feat(decisions): add admin review-selection screen

valentin0h:
Gives admins a per-proposal aggregate view during the review-selection phase so they can see reviewer scores before finalizing winners. Gated on admin_review_selection; confirm-decisions mutation is intentionally deferred.

Not in this PR:
- Pagination
- Filters


apps/app/src/components/decisions/DecisionStateRouter.tsx
```diff
@@ -40,8 +41,26 @@ function DecisionStateRouterNew({
...
+      previousPhase?.rules?.proposals?.review === true && reviewFlowEnabled;
+
+    if (previousWasReview && previousPhase && isAdmin) {
+      return (
+        <ReviewSelectionPage
+          instance={instance}
+          previousPhaseId={previousPhase.phaseId}
+        />
+      );
+    }
```
valentin0h, line 63: @scazan 👀

apps/app/src/components/decisions/useManualSelection.ts
valentin0h, line 1: Added this because I needed a sync between mounts. In the next PR, we make selections in a child screen.

packages/common/src/services/decision/getInstance.ts
```diff
@@ -76,6 +76,11 @@ export const getInstance = async ({ instanceId, user }: GetInstanceInput) => {
         process: true,
         owner: true,
         steward: true,
+        profile: {
+          columns: {
+            slug: true,
+          },
+        },
```
valentin0h, line 83: This is good to have because we avoid keeping a slug reference in the client when we have the `instance`

packages/common/src/services/decision/reviewHelpers.ts
```diff
@@ -163,12 +163,18 @@ export function resolveAssignmentProposal(assignment: {
 
   const proposalData = parseProposalData(snapshot.proposalData);
 
-  if (!proposalData.collaborationDocId) {
-    throw new ValidationError(`Proposal ${id} is missing collaborationDocId`);
+  // Temporary: accept HTML-only proposals until local TipTap lands.
+  if (!proposalData.collaborationDocId && !proposalData.description) {
+    throw new ValidationError(`Proposal ${id} has no document content`);
```
valentin0h, line 168: To avoid creating tiptap docs when testing locally. I'll see how we go about this in the future. As soon as we have legacy decisions this is fine.

apps/app/src/components/decisions/pages/ReviewSelectionPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import { TranslatedText } from '@/components/TranslatedText';
+
+import { DecisionHero } from '../DecisionHero';
+import {
+  ReviewSelectionList,
+  ReviewSelectionListSkeleton,
+} from '../ReviewSelection/ReviewSelectionList';
+
+type Instance = RouterOutput['decision']['getInstance'];
```
scazan: Do we want to start using this pattern? I'm seeing it more and more popping up but it doesn't fit with the rest of the code. We can just use the backend encoder type to keep the type consistent across the board.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
```diff
@@ -0,0 +1,172 @@
+'use client';
```
scazan, line 1: There is quite a bit of duplication in this versus manual selection list and I'm wondering if we can just use manual selection list for the most part and feed it data or restructure it so that it can be more composable. There are more scenarios beyond reviews and a bare manual selection that we will need in the future. Plus we'll want to keep the styles in sync and the duplication will cause issues as we update it.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { Bullet } from '@/components/Bullet';
+
+import { SelectionConfirmDialog } from '../SelectionConfirmDialog';
+import { useManualSelection } from '../useManualSelection';
+import {
+  ReviewSelectionTable,
+  ReviewSelectionTableSkeleton,
+} from './ReviewSelectionTable';
+
+type Instance = RouterOutput['decision']['getInstance'];
```
scazan: We can just use the back end type.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const advancing = useMemo(() => new Set(advancingIds), [advancingIds]);
+  const [isConfirmOpen, setIsConfirmOpen] = useState(false);
+
+  const [data] = trpc.decision.listWithReviewAggregates.useSuspenseQuery({
+    processInstanceId,
+    phaseId: previousPhaseId,
+  });
+  const utils = trpc.useUtils();
+
+  const items = data.items;
```
scazan: nit: just destruct these?

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
```diff
@@ -0,0 +1,172 @@
...
+  ReviewSelectionTableSkeleton,
+} from './ReviewSelectionTable';
+
+type Instance = RouterOutput['decision']['getInstance'];
+
+export function ReviewSelectionList({
+  instance,
+  previousPhaseId,
+}: {
+  instance: Instance;
```
scazan: Should we just lean on cache here instead of drilling this through? We should reduce external dependencies and keep it composable.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
```diff
@@ -0,0 +1,172 @@
...
+          <Bullet />
+          <span className="font-serif text-title-base text-neutral-black">
+            {total}
+          </span>
+        </div>
+      </div>
+
+      {items.length === 0 ? (
+        <EmptyState icon={<LuLeaf className="size-6" />}>
+          <Header3 className="font-serif !text-title-base font-light text-neutral-black">
```
scazan: I think there are extra unnecessary classes here that override our system. If we use ManualSelectionList that'll all get solved though.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+  TableRow,
+} from '@op/ui/ui/table';
+import { LuCheck } from 'react-icons/lu';
+
+import { useTranslations } from '@/lib/i18n';
+
+type Item = ProposalWithAggregates;
+
+const RECOMMENDATION_INTENT: Record<RecommendationValue, StatusDotIntent> = {
+  yes: 'success',
```
scazan, line 30: Not sure if we have constants for this... just a smell when I see them hardcoded like this.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+          <Skeleton className="hidden h-4 w-32 md:block" />
+          <Skeleton className="hidden h-4 w-12 md:block" />
+          <Skeleton className="h-8 w-28" />
+        </div>
+      ))}
+    </div>
+  );
+}
+
+function ProposalCard({
```
scazan, line 189: I know this is internal but probably still want to say, ProposalReviewCard or ReviewProposalCard to not confuse it.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+  advancing: boolean;
+  onPress: () => void;
+  /** Proposal title — surfaced as the button's accessible name so screen
+   * readers and tests can disambiguate between rows. */
+  title: string;
+}) {
+  const t = useTranslations();
+
+  return (
+    <Button
```
scazan: Should we re-use the same component we use for advancing in Manual Selection List?
There is an issue here visually in that the checkmark grows the button when you set it active (whereas in the manual selection table it does not as per design's request).

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+          </>
+        ) : (
+          t('Advance')
+        )}
+      </span>
+    </Button>
+  );
+}
+
+function CategoryChips({ categories }: { categories: Item['categories'] }) {
```
scazan: Another one we use are duplicating from the other table. It starts to really feel like this should be composed more.

apps/app/src/components/decisions/DecisionStateRouter.tsx
scazan: Good enough for the moment :) It brings up that maybe we can use something really composable here for the table so that we can do any combo of capabilities in the future.

apps/app/src/components/decisions/ManualSelectionList.tsx
```diff
@@ -186,11 +189,14 @@ export const ManualSelectionList = ({
         />
       )}
 
-      <VotingSubmitFooter isVisible>
-        <div className="flex w-full items-center justify-between px-4 sm:max-w-6xl sm:px-8">
-          <span className="text-neutral-black">
+      <FooterBar position="fixed" className="bg-neutral-offWhite/95">
+        <FooterBar.Start>
+          <span className="text-base text-neutral-black">
```
scazan, line 194: This looks like our defaults. Is this necessary?

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+  onPress: () => void;
+  /** Proposal title — surfaced as the button's accessible name so screen
+   * readers and tests can disambiguate between rows. */
+  title: string;
+}) {
+  const t = useTranslations();
+
+  return (
+    <Button
+      size="small"
```
scazan: Don't think this matches the design. Should be the same border-radius as the Confirm button. You can see in the video that it flips between the sizes.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
```diff
@@ -0,0 +1,372 @@
...
+  TableCell,
+  TableColumn,
+  TableHeader,
+  TableRow,
+} from '@op/ui/ui/table';
+import { LuCheck } from 'react-icons/lu';
+
+import { useTranslations } from '@/lib/i18n';
+
+type Item = ProposalWithAggregates;
```
scazan: Why mask this? we can just use ProposalWithAggregates down below

scazan (changes requested):
Feels like we are duplicating a lot of the login around manual selection here and we are going to have to keep the styles in sync.
There are a few design things that are not matching up as well.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
valentin0h: They are typed, it's fine.

scazan (approved):
Adding a "conditional approval" here so we can get it in after updates.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
valentin0h: I think `instance` and `previousPhaseId` are the primary inputs here, not external deps. The router already computed the previous phase to gate this page, so it's just handing down what it already needed. Going through the cache would add a Suspense boundary for data we already have.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
valentin0h: updated

apps/app/src/components/decisions/pages/ReviewSelectionPage.tsx
valentin0h: We needed to have that type in op/common/schemas hence I didn't reach for the "backend" type in this PR. But yes using encoder types is slightly better in this case. Updated.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
valentin0h: Moved some components that can be reused.

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionList.tsx
valentin0h: https://github.com/oneprojectorg/common/pull/1068/changes#r3180809971

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
valentin0h: Should be fixed now due to https://github.com/oneprojectorg/common/pull/1068#discussion_r3180814304

apps/app/src/components/decisions/ReviewSelection/ReviewSelectionTable.tsx
valentin0h: Should be fixed now as of https://github.com/oneprojectorg/common/pull/1068#discussion_r3180814304

PR #1073, opened 2026-04-27
test(a11y): add axe-core baseline scan for WCAG 2.1 AA

nourmalaeb:
Adds `axe-core` and a playwright test that generates a report. Will follow up with CI that diffs report against baseline and adds a comment to the PR.


tests/e2e/tests/a11y-baseline.spec.ts
```diff
@@ -0,0 +1,807 @@
...
+    occurrences: number;
+    routes: number;
+    help: string;
+    helpUrl: string;
+    wcagCriteria: WcagCriterion[];
+  }>;
+  routes: RouteResult[];
+  screenshotsAttempted: number;
+  screenshotsCaptured: number;
+}
```
valentin0h, line 82: Aren't these types exposed by `axe-core`?

tests/e2e/tests/a11y-baseline.spec.ts
```diff
@@ -0,0 +1,807 @@
...
+      `[a11y-baseline] critical=${report.totals.critical} serious=${report.totals.serious} moderate=${report.totals.moderate} minor=${report.totals.minor}`,
+    );
+    console.log(
+      `[a11y-baseline] screenshots ${report.screenshotsCaptured}/${report.screenshotsAttempted}`,
+    );
+  });
+});
+
+async function seedDynamicRoutes(
+  org: import('@op/test').CreateOrganizationResult,
```
valentin0h: Do we need dynamic import?

tests/e2e/tests/a11y-baseline.spec.ts
```diff
@@ -0,0 +1,807 @@
...
+  try {
+    await page.goto(route.url, {
+      waitUntil: 'domcontentloaded',
+      timeout: 30_000,
+    });
+    // Best-effort settle: wait up to 15s for the page to quiesce so axe sees a stable DOM.
+    // Pages with persistent connections (websockets, SSE, hot reload) never reach idle —
+    // we proceed regardless and warn so CI logs flag routes that always fall through.
+    await page
+      .waitForLoadState('networkidle', { timeout: 15_000 })
```
valentin0h: `networkidle` is discouraged in docs https://playwright.dev/docs/api/class-page#page-wait-for-load-state-option-state

valentin0h (approved):
nice 💯 

A few things to consider, and some questions I have

- Dynamic imports
- Do we need three files generated? Could 1 be enough, mainly talking about CI? 
- Are screenshots necessary?

tests/e2e/tests/a11y-baseline.spec.ts
nourmalaeb: they don't export all of them but i did manage to simplify these a bit.

tests/e2e/tests/a11y-baseline.spec.ts
nourmalaeb: good catch, fixed 👍

tests/e2e/tests/a11y-baseline.spec.ts
nourmalaeb: oh didn't know that, updated to wait for `load` instead as the docs recommend

PR #1075, opened 2026-04-28
feat(decisions): default selection pipeline to pass-none

scazan:
## Summary

- Flip the implicit selection-pipeline fallback in `advancePhase` from pass-all to **pass-none**. Phases without an explicit `selectionPipeline` no longer auto-advance every proposal — they require a manual selection (or an explicit pipeline) instead.
- Phases that should still let everything through opt in via an empty pipeline: `selectionPipeline: { version: '1.0.0', blocks: [] }`.

### DIFF
--- a/packages/common/src/services/decision/schemas/definitions.ts
+++ b/packages/common/src/services/decision/schemas/definitions.ts
@@ -51,7 +51,7 @@ export const simpleVoting: DecisionSchemaDefinition = {
       rules: {
         proposals: { submit: true },
         voting: { submit: false },
-        advancement: { method: 'date', endDate: '2026-01-01' },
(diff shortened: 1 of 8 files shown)


packages/common/src/services/decision/schemas/definitions.ts
```diff
@@ -90,7 +90,7 @@ export const simpleVoting: DecisionSchemaDefinition = {
       rules: {
         proposals: { submit: false },
         voting: { submit: false },
-        advancement: { method: 'date', endDate: '2026-01-02' },
+        advancement: { method: 'manual', endDate: '2026-01-02' },
```
valentin0h, line 100: Can we remove the end dates or are they needed?

valentin0h (approved):

PR #1076, opened 2026-04-28
feat(reviews): add per-proposal submitted-reviews endpoint

valentin0h:
Powers the Review Summary panel and reviewer drill-down with one call: the existing aggregate triple plus a submitted-only reviews list carrying per-row score and overall recommendation.

### DIFF
--- a/packages/common/src/services/decision/getProposalWithReviewAggregates.ts
+++ b/packages/common/src/services/decision/getProposalWithReviewAggregates.ts
@@ -1,0 +1,95 @@
+import { db } from '@op/db/client';
+import type { User } from '@op/supabase/lib';
+import { z } from 'zod';
+
+import { NotFoundError, UnauthorizedError } from '../../utils';
+import { getInstance } from './getInstance';
+import { getRubricScoringInfo } from './getRubricScoringInfo';
+import {
+  getCategoriesByProposalIds,
+  getComputedReviewAggregates,
+  getSubmittedReviewScore,
+  proposalRelations,
+} from './listProposalsWithReviewAggregates';
+import {
+  type ProposalWithSubmittedReviews,
+  proposalWithSubmittedReviewsSchema,
+} from './schemas/reviews';
+
+export const getProposalWithReviewAggregatesInputSchema = z.object({
+  processInstanceId: z.uuid(),
+  proposalId: z.uuid(),
+  phaseId: z.string().optional(),
+});
+
+export type GetProposalWithReviewAggregatesInput = z.infer<
+  typeof getProposalWithReviewAggregatesInputSchema
+>;
+
+/**
+ * Submitted-only by design: drafts and unstarted assignments contribute to
+ * `aggregates.assignmentsCount` but are not surfaced in `reviews[]`.
+ */
+export async function getProposalWithReviewAggregates(
+  input: GetProposalWithReviewAggregatesInput & { user: User },
+): Promise<ProposalWithSubmittedReviews> {
+  const { user, processInstanceId, proposalId } = input;
+
+  const instance = await getInstance({ instanceId: processInstanceId, user });
+
+  if (!instance.access.admin) {
+    throw new UnauthorizedError(
+      "You don't have admin access to this process instance",
+    );
+  }
+
+  const rubricTemplate = instance.instanceData.rubricTemplate;
+  const scoredCriterionKeys = rubricTemplate
+    ? getRubricScoringInfo(rubricTemplate)
+        .criteria.filter((c) => c.scored)
+        .map((c) => c.key)
+    : [];
+
+  const phaseId = input.phaseId ?? instance.currentStateId ?? undefined;
+
+  const [proposal, categoriesByProposalId] = await Promise.all([
+    db.query.proposals.findFirst({
+      where: { id: proposalId },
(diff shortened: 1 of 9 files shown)


packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -366,3 +355,32 @@ function getComputedReviewAggregates(
...
+  review: { state: string; reviewData: unknown } | null | undefined,
+  scoredCriterionKeys: string[],
+): { score: number; overallRecommendation: string | null } | null {
+  if (!review || review.state !== ProposalReviewState.SUBMITTED) {
+    return null;
+  }
+
+  const data = review.reviewData as {
+    answers?: Record<string, unknown>;
+  } | null;
```
valentin0h, line 370: I'll fix this separately upstream.

packages/common/src/services/decision/schemas/reviews.ts
```diff
@@ -150,6 +150,22 @@ export const proposalsWithReviewAggregatesListSchema = z.object({
   next: z.string().nullable(),
 });
 
+// ── Single proposal with submitted reviews ─────────────────────────────
+
+/** `score` and `overallRecommendation` are precomputed so clients don't redo the rubric math. */
+export const submittedReviewItemSchema = z.object({
```
valentin0h, line 156:
```suggestion
export const reviewItemSchema = z.object({
```

packages/common/src/services/decision/getProposalWithReviewAggregates.ts
```diff
@@ -0,0 +1,95 @@
...
+});
+
+export type GetProposalWithReviewAggregatesInput = z.infer<
+  typeof getProposalWithReviewAggregatesInputSchema
+>;
+
+/**
+ * Submitted-only by design: drafts and unstarted assignments contribute to
+ * `aggregates.assignmentsCount` but are not surfaced in `reviews[]`.
+ */
```
valentin0h, line 32: Might change this but work for now.

services/api/src/routers/decision/proposals/getWithReviewAggregates.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  getProposalWithReviewAggregates,
+  getProposalWithReviewAggregatesInputSchema,
+  proposalWithSubmittedReviewsSchema,
+} from '@op/common';
+
+import { commonAuthedProcedure, router } from '../../../trpcFactory';
+
+export const getWithReviewAggregatesRouter = router({
+  getWithReviewAggregates: commonAuthedProcedure()
```
scazan: This kind of falls out of our naming convention. The get here is unclear. Should it not be getProposalWithReviewAggregates? We should keep it consistent and actually I still like this way of doing it with getProposalWithReviewAggregates()

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
```diff
@@ -330,28 +330,17 @@ function getComputedReviewAggregates(
...
-      }
-    }
-
-    const recommendation = answers[OVERALL_RECOMMENDATION_KEY];
-    if (recommendation != null) {
-      const recommendationKey = String(recommendation);
-      overallRecommendationCount[recommendationKey] =
-        (overallRecommendationCount[recommendationKey] ?? 0) + 1;
+    if (scored.overallRecommendation != null) {
+      const key = scored.overallRecommendation;
```
scazan: It seems like this would be a value rather than a key given what is in `getSubmittedReviewScore`. Is this correct?

packages/common/src/services/decision/getProposalWithReviewAggregates.ts
```diff
@@ -0,0 +1,95 @@
...
+export async function getProposalWithReviewAggregates(
+  input: GetProposalWithReviewAggregatesInput & { user: User },
+): Promise<ProposalWithSubmittedReviews> {
+  const { user, processInstanceId, proposalId } = input;
+
+  const instance = await getInstance({ instanceId: processInstanceId, user });
+
+  if (!instance.access.admin) {
```
scazan, line 40: This totally works.. but there's something fishy about it since it is meant to resolve permissions for the frontend really (wondering if we should use an assertAccess here instead). Can't think of a good reason not to use this necessarily since it wouldn't cause an issue if it changes. just flagging it. I will think more on it.

scazan (approved):

packages/common/src/services/decision/getProposalWithReviewAggregates.ts
valentin0h: I agree, there's some indirection here. I'll update separately.

packages/common/src/services/decision/listProposalsWithReviewAggregates.ts
valentin0h: Updated name

services/api/src/routers/decision/proposals/getWithReviewAggregates.ts
valentin0h: No strong preference. I'll rename to `getProposalWithReviewAggregates` but it'll read like this in usage `proposal.getProposalWithReviewAggregates`.

PR #1077, opened 2026-04-28
feat(reviews): add review summary page for proposals

valentin0h:
Surfaces per-proposal review aggregates and individual reviewer detail at `/decisions/[slug]/proposal/[slug]/reviews`

### DIFF
--- a/apps/app/src/components/decisions/ReviewSummary/ReviewSummaryLayout.tsx
+++ b/apps/app/src/components/decisions/ReviewSummary/ReviewSummaryLayout.tsx
@@ -1,0 +1,118 @@
+import {
+  HydrationBoundary,
+  createServerUtils,
+  dehydrate,
+} from '@op/api/server';
+import { createClient } from '@op/api/serverClient';
+import { CommonError } from '@op/common';
+import { forbidden, notFound } from 'next/navigation';
+
+import { ReviewSummaryView } from './ReviewSummaryView';
+
+interface ReviewSummaryLayoutProps {
+  decisionSlug: string;
+  proposalProfileId: string;
+}
+
+export async function ReviewSummaryLayout({
+  decisionSlug,
+  proposalProfileId,
+}: ReviewSummaryLayoutProps) {
+  const client = await createClient();
+
+  let decisionProfile;
+  try {
+    decisionProfile = await client.decision.getDecisionBySlug({
+      slug: decisionSlug,
+    });
+  } catch (error) {
+    const cause = error instanceof Error ? error.cause : null;
+    if (cause instanceof CommonError && cause.statusCode === 403) {
+      forbidden();
+    }
+    if (cause instanceof CommonError && cause.statusCode === 404) {
+      notFound();
+    }
+    throw error;
+  }
+
+  if (!decisionProfile?.processInstance) {
+    notFound();
+  }
+
+  if (!decisionProfile.processInstance.access?.admin) {
+    forbidden();
+  }
+
+  const instanceId = decisionProfile.processInstance.id;
+
+  const { utils, queryClient } = await createServerUtils();
+
+  const [proposal, instance] = await Promise.all([
+    (async () => {
+      try {
+        return await utils.decision.getProposal.fetch({
+          profileId: proposalProfileId,
+        });
+      } catch (error) {
+        const cause = error instanceof Error ? error.cause : null;
+        if (cause instanceof CommonError && cause.statusCode === 404) {
+          notFound();
+        }
+        throw error;
+      }
+    })(),
+    utils.decision.getInstance.fetch({ instanceId }),
+  ]);
+
+  if (proposal.processInstanceId !== instanceId) {
+    notFound();
+  }
+
+  const proposalId = proposal.id;
+  const phaseId = resolveReviewPhaseId(instance);
+
+  await utils.decision.getProposalWithReviewAggregates.prefetch({
+    processInstanceId: instanceId,
+    proposalId,
+    phaseId,
+  });
+
+  return (
+    <HydrationBoundary state={dehydrate(queryClient)}>
+      <ReviewSummaryView
+        decisionSlug={decisionSlug}
+        instanceId={instanceId}
+        proposalId={proposalId}
+        proposalProfileId={proposalProfileId}
+        phaseId={phaseId}
+      />
+    </HydrationBoundary>
+  );
+}
+
+// Walk back from the current phase to find the most recent review phase —
+// after a review phase ends, currentStateId no longer points at it but the
+// review assignments are still pinned to the original review phase id.
+function resolveReviewPhaseId(instance: {
+  currentStateId: string | null;
+  instanceData?: {
+    phases?: Array<{
+      phaseId: string;
+      rules?: { proposals?: { review?: boolean } };
+    }>;
+  } | null;
+}): string | undefined {
+  const phases = instance.instanceData?.phases ?? [];
+  const currentIdx = phases.findIndex(
+    (p) => p.phaseId === instance.currentStateId,
+  );
+  const startIdx = currentIdx === -1 ? phases.length - 1 : currentIdx;
+  for (let i = startIdx; i >= 0; i--) {
+    const phase = phases[i];
+    if (phase?.rules?.proposals?.review === true) {
+      return phase.phaseId;
+    }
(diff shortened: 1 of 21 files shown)


apps/app/src/components/decisions/ReviewSummary/ReviewerDetail.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </Button>
+
+      <div className="flex items-center justify-between border-b border-neutral-gray1 pb-4">
+        <div className="flex items-center gap-2">
+          <ProfileAvatar
+            profile={item.reviewer}
+            withLink={false}
+            className="size-6"
+          />
+          <Header3 className="font-serif">
```
scazan, line 67: Just flaggin that this looked a bit too small in the demo video you had, but here it seems to be to spec. Worth checking maybe? Check that we don't need to override here.

apps/app/src/components/decisions/ReviewSummary/ReviewerDetail.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const recommendationLabel = (() => {
+    if (!hasOverallRecommendation || !item.overallRecommendation) {
+      return null;
+    }
+    const options = parseSchemaOptions(
+      rubricTemplate.properties?.[OVERALL_RECOMMENDATION_KEY],
+    );
+    const match = options.find(
+      (option) => option.value === item.overallRecommendation,
```
scazan: I always find this confusing and I have stop for a minute when I read it (saw it in a previous PR as well). I think specifically because it is the option.value we are matching here. Can you add a comment here or, better, pull this out into a named util?

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+export function ReviewerList({
+  aggregates,
+  rubricTemplate,
+  rubricSummary,
+  onSelectAssignment,
+}: ReviewerListProps) {
+  const t = useTranslations();
+  const { reviewsSubmittedCount, assignmentsCount, averageScore } =
+    aggregates.aggregates;
```
scazan: Maybe a good moment to rethink this naming if we end up with aggregates.aggregates.

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
```diff
@@ -0,0 +1,231 @@
...
+              />
+            ))}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+interface ReviewerGroup {
```
scazan, line 131: Not for this PR necessarily but this is such a standard type here where we could just use a generic. but maybe unnecessary.

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
```diff
@@ -0,0 +1,231 @@
...
+  value: string;
+  label: string;
+  items: SubmittedReviewItem[];
+}
+
+function groupReviewsByRecommendation(
```
scazan: I tend to strongly prefer these phrased more as getReviewsGroupedByRecommendation(). I think it's a bit clearer and signals that there definitely is a return value here.

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
```diff
@@ -0,0 +1,231 @@
...
+  showScore: boolean;
+  totalPoints: number;
+  onSelect: (assignmentId: string) => void;
+}) {
+  const t = useTranslations();
+  return (
+    <Button
+      unstyled
+      onPress={() => onSelect(item.review.assignmentId)}
+      className="flex h-14 w-full cursor-pointer items-center justify-between rounded-lg border border-neutral-gray1 bg-white px-3 py-2 text-left outline-0 outline-transparent transition-colors duration-200 hover:bg-neutral-offWhite focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-data-blue"
```
scazan, line 209: It's odd that this is both unstyled but also we attach a border, various colors and so forth. Is there no variant that can support this and should we have a variant?

apps/app/src/components/decisions/ReviewSummary/ReviewSummaryLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  } | null;
+}): string | undefined {
+  const phases = instance.instanceData?.phases ?? [];
+  const currentIdx = phases.findIndex(
+    (p) => p.phaseId === instance.currentStateId,
+  );
+  const startIdx = currentIdx === -1 ? phases.length - 1 : currentIdx;
+  for (let i = startIdx; i >= 0; i--) {
+    if (phases[i]?.rules?.proposals?.review === true) {
+      return phases[i]!.phaseId;
```
scazan: fishy with a `!` here.

apps/app/src/components/decisions/ReviewSummary/ReviewSummaryLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+        instanceId={instanceId}
+        proposalId={proposalId}
+        proposalProfileId={proposalProfileId}
+        phaseId={phaseId}
+      />
+    </HydrationBoundary>
+  );
+}
+
+// Walk back from the current phase to find the most recent review phase —
```
scazan, line 94: Won't this always want to be the previous phase? Because if it's not the previous phase, we may have come out of a different type of phase. Not sure if it's a problem, but wondering if that's expected behavior.

apps/app/src/components/decisions/ReviewSummary/ReviewSummaryNavbar.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export function ReviewSummaryNavbar({
+  decisionSlug,
+}: {
+  decisionSlug: string;
+}) {
+  const t = useTranslations();
+
+  return (
+    <header className="sticky top-0 z-20 flex h-14 shrink-0 items-center justify-between border-b bg-white px-6 md:px-8">
```
scazan: We keep replicating this. We should pull it out into a component. There might already be one actually since we use this all over the place.

scazan (approved):

apps/app/src/components/decisions/ReviewSummary/ReviewerDetail.tsx
valentin0h: Matches spec :thumbsup:

apps/app/src/components/decisions/ReviewSummary/ReviewerDetail.tsx
valentin0h: Updated :thumbsup:

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
valentin0h: Renamed to `proposalWithReviews` :thumbsup:

apps/app/src/components/decisions/ReviewSummary/ReviewerList.tsx
valentin0h: Closest variant (`color="secondary"`) fights the row layout (`w-fit`, `justify-center`, `shadow-md`). Only use site for now — happy to extract a `row`/`card` variant once the pattern shows up a second time.

apps/app/src/components/decisions/ReviewSummary/ReviewSummaryLayout.tsx
valentin0h: Walk-back is intentional — we want the most recent review phase even if a non-review phase sits between the current state and that review (e.g. `review → discussion → decision`, currently in `decision`).

apps/app/src/components/decisions/ReviewSummary/ReviewSummaryNavbar.tsx
valentin0h: Extracted to `DecisionSubpageHeader`; `ReviewSummaryNavbar` and `ReviewNavbar` now share it :thumbsup:

PR #1082, opened 2026-04-28
chore(deps): upgrade react-aria to 3.48 and react-aria-components to 1.17

nourmalaeb:
## Summary

- Bump `react-aria` (3.37 → 3.48), `react-aria-components` (1.6 → 1.17), `react-stately` (3.35 → 3.46), `@internationalized/date` (3.8 → 3.12, deduped with RAC's bundled version).
- Thread the new selection-mode generic through `@op/ui` `Select` wrapper.
- Widen `onSelectionChange` handlers to `Key | null` (RAC now allows null on clear).

---
  - https://app.asana.com/0/0/1214523507160361


package.json
```diff
@@ -92,8 +92,8 @@
     "postcss": "^8.5.3",
     "postgres": "^3.4.5",
     "react": "^19.0.1",
-    "react-aria": "^3.37.0",
-    "react-aria-components": "^1.6.0",
+    "react-aria": "^3.48.0",
+    "react-aria-components": "^1.17.0",
```
valentin0h: suggestion: If we're using them in a few places, consider adding them to `pnpm-workspace.yml`

valentin0h (approved):
Nice 👌  

Let's just double-check that all places that use Select are updated if needed.

package.json
nourmalaeb: yes good suggestion 👍

PR #1084, opened 2026-04-29
fix: Scope drafts per phase

scazan:
This PR fixes a bug where draft proposals weren't part of the phase scoping in listProposals: a draft created in Phase 1 would still appear when viewing Phase 2 after the instance advanced.

Fix: Drafts are now scoped to the phase they were created in via a createdAt temporal window ([inboundAt, outboundAt) — half-open so boundary timestamps land in exactly one phase). Ownership filtering (creator + collaborators via profileUsers) is pushed into the SQL subquery so we don't load drafts the caller can't access.

  Implementation: New getPhaseProposalAndDraftIds resolves the instance context and phase window once, then runs the attachment, non-draft, and draft queries in parallel. listProposals skips the draft fetch entirely for explicit-scope callers (ballots, internal proposalIds) and trusted contexts. This is intentional to keep the path of processing proposals free of any draft concerns.

Secondarily,
It also improves the getProposalIdsForPhase to deduplicate instance queries.


services/api/src/routers/decision/proposals/listProposalsPhaseScoped.test.ts
valentin0h, line 1: Any reason why we have these `listProposals` tests not in `listProposals.test.ts`? I was confused for a sec and went searching for a `listProposalsPhaseScoped` router. 

We can split to something like `listProposals.scoped.test.ts` if the file is the concern. It should be clearer which file we're testing.

services/api/src/routers/decision/proposals/listProposalsPhaseScoped.test.ts
```diff
@@ -185,4 +191,313 @@ describe.concurrent('listProposals: phase-scoped proposal visibility', () => {
...
+      processInstanceId: instanceId,
+      proposalData: { title: `Phase-2 draft ${task.id}` },
+    });
+
+    const result = await caller.decision.listProposals({
+      processInstanceId: instanceId,
+    });
+
+    expect(result.proposals.map((p) => p.id)).toContain(draftInReview.id);
+  });
```
valentin0h: Can we merge this into `'shows the creator their draft when viewing the phase it was created in'` for performance?

services/api/src/routers/decision/proposals/listProposalsBallot.test.ts
```diff
@@ -174,6 +174,63 @@ describe.concurrent('listProposals: votedByProfileId (ballot filter)', () => {
...
+    // Voter creates one submitted proposal (which will be on the ballot) and
+    // one draft proposal (which must NOT appear in ballot results).
+    const submittedProposal = await testData.createProposal({
+      userEmail: voter.email,
+      processInstanceId: instance.instance.id,
+      proposalData: { title: `Submitted ${task.id}` },
+    });
+    await voterCaller.decision.submitProposal({
+      proposalId: submittedProposal.id,
+    });
```
valentin0h: Use the `status` in testData.createProposal above instead.

packages/common/src/services/decision/listProposalSubmitters.ts
```diff
@@ -55,9 +60,17 @@ export const listProposalSubmitters = async ({
...
+  const phaseProposalIds =
+    instanceContext.isLegacy || !instanceContext.currentPhaseId
+      ? await getActiveNonDraftIdsForInstance({
+          instanceId: processInstanceId,
+        })
+      : await getProposalIdsForPhase({
+          instanceId: processInstanceId,
+          phaseId: instanceContext.currentPhaseId,
+          instanceContext,
+        });
```
valentin0h: I'm not sure about this. We're branching on what function to use to get phaseProposalIds. But since we pass `instanceContext` to `getProposalIdsForPhase`, couldn't we just have this logic in `getProposalIdsForPhase`?

packages/common/src/services/decision/listProposals.ts
```diff
@@ -168,62 +172,94 @@ export const listProposals = async ({
...
+      return { phaseProposalIds: ids, phaseDraftIds: [] };
+    }
+    const ids = await getPhaseProposalAndDraftIds({
+      instanceId: processInstanceId,
+      phaseId: input.phaseId,
+      authUserId: input.authUserId,
+      instanceContext,
+    });
+    return { phaseProposalIds: ids.nonDraftIds, phaseDraftIds: ids.draftIds };
+  })();
```
valentin0h, line 232: I'm wondering whether we'll end up duplicating this conditional logic every time we need to fetch phased proposal IDs. Would it be possible to have a single function `getProposalIdsForPhase` controlled by flags/params?

packages/common/src/services/decision/getProposalsForPhase.ts
```diff
@@ -170,63 +207,93 @@ async function submittedDuringIds(
...
-    return allActiveIds(instanceId, db);
-  }
-
-  const resolvedPhaseId = phaseId ?? ctx.currentPhaseId;
-  if (!resolvedPhaseId) {
-    return allActiveIds(instanceId, db);
-  }
+  const nonDraftPredicate = ne(proposals.status, ProposalStatus.DRAFT);
 
   const window = await resolvePhaseWindow(
```
valentin0h: Can we name this something else? I get confused for a second thinking it's the browser `window`.

packages/common/src/services/decision/getProposalsForPhase.ts
```diff
@@ -235,23 +302,156 @@ export async function getProposalIdsForPhase({
...
+  db = defaultDb,
+}: {
+  instanceId: string;
+  phaseId?: string;
+  authUserId: string;
+  /**
+   * Pre-derived instance context. Pass this when the caller has already
+   * fetched the `processInstances` row to avoid a duplicate read.
+   */
+  instanceContext?: InstanceContext;
```
valentin0h: I don't see much value in having this as optional. It'll force the caller to make better decisions about data fetching. Imo ideally we pass `instance: {...}`, instead of `instanceId` and `instanceContext?`.

valentin0h (approved):

packages/common/src/services/decision/getProposalsForPhase.ts
scazan: ah interesting. It's a standard name for this sort of thing. Maybe we could call it resolvePhaseRange?

packages/common/src/services/decision/listProposals.ts
scazan: hmm... I don't know.. I like the composable approach more here because the choice is pretty specific to the use-case. It leaves all the logic on whether the caller ALSO wants drafts to the caller rather than in the function. Not a big fan of the flags approach generally.

packages/common/src/services/decision/getProposalsForPhase.ts
scazan: ah you are referring to the variable here not the resolvePhaseWindow on the same line. What do you think about `phaseWindow`?

PR #1090, opened 2026-04-30
feat(ui): add StatusDot component

valentin0h:
Reusable status indicator for review summaries and similar surfaces.

https://apps-workshop-git-op-ui-status-dot-oneproject.vercel.app/

### DIFF
--- a/packages/ui/src/components/StatusDot.tsx
+++ b/packages/ui/src/components/StatusDot.tsx
@@ -1,0 +1,40 @@
+import type { ReactNode } from 'react';
+
+import { cn } from '../lib/utils';
+
+export type StatusDotIntent = 'success' | 'danger' | 'warning' | 'neutral';
+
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/StatusDot.tsx
```diff
@@ -0,0 +1,45 @@
...
+  className,
+  children,
+}: StatusDotProps) => {
+  const dotColor = DOT_COLOR_BY_INTENT[intent];
+
+  if (!children) {
+    return (
+      <span
+        className={cn('inline-block size-2 rounded-full', dotColor, className)}
+        aria-hidden
```
scazan, line 32: We shouldn't use this.. instead let's set an aria-label fallback that uses the intent so screenreaders can take advantage of the warning.

scazan (approved):
Approved but needing one fix around the aria-label here.

scazan (approved):
Approved but needing one fix around the aria-label here.

PR #1091, opened 2026-04-30
feat: Decision Scoped Posts Frontend

scazan:
Adds in a sidepanel which will be used for resources as well as this Decision updates feature.

Screenshot 2026-05-19 at 12 58 28


apps/app/src/components/PostFeed/index.tsx
```diff
@@ -526,6 +526,7 @@ export const usePostFeedActions = () => {
       void utils.organization.listPosts.invalidate();
       void utils.organization.listAllPosts.invalidate();
       void utils.posts.getPosts.invalidate();
+      void utils.posts.listProfilePosts.invalidate();
```
scazan, line 529: deferring this to a separate PR which will activate the realtime channel invalidation for all posts

apps/app/src/components/decisions/DecisionSidePanel.tsx
```diff
@@ -0,0 +1,239 @@
...
+  return (
+    <>
+      {isOpen ? (
+        <div
+          aria-hidden="true"
+          onClick={close}
+          className="fixed inset-0 z-30 bg-neutral-black/30 sm:hidden"
+        />
+      ) : null}
+      <aside
```
scazan: We aren't using a sheet here because this component acts a bit different than usual. I do think actually we will probably want to extract this out but to keep this PR small, we're having a custom one-off component here.

apps/app/src/components/decisions/DecisionSidePanel.tsx
```diff
@@ -0,0 +1,239 @@
...
+}: {
+  decisionProfileId: string;
+  canPostUpdate: boolean;
+  canReadUpdates: boolean;
+}) => {
+  const t = useTranslations();
+  const utils = trpc.useUtils();
+
+  const handlePostSuccess = useCallback(() => {
+    void utils.posts.listProfilePosts.invalidate({
```
scazan, line 153: Deferring the channel invalidation here to a different PR which handles post invalidations across the board

apps/app/src/components/decisions/DecisionInstanceHeader.tsx
```diff
@@ -47,6 +51,7 @@ export const DecisionInstanceHeader = ({
       </div>
 
       <div className="flex items-center justify-end gap-2 md:gap-4">
+        <DecisionUpdatesToggle ariaLabel={t('Open updates panel')} />
```
nourmalaeb: Nit: should this be "Toggle updates panel" (instead of "Open updates panel")?

apps/app/src/components/decisions/DecisionSidePanel.tsx
nourmalaeb: The Sidebar component should work here, was there an issue using it?

nourmalaeb (approved):

apps/app/src/components/decisions/DecisionSidePanel.tsx
scazan: It's a little bit different, but mostly just in the floating part of it. That one pushes everything out of the way whereas this one floats on top. We could try to create a variant, but since this was the only instance I just put together this, not sure if we were going to replicate it as a component for other things. However, I think it's worth asking the question about whether we want it to overlay or whether we want it to push everything aside. Maybe to push everything aside is a better approach?

scazan: Ok, I'm updating Sidebar

PR #1092, opened 2026-05-01
feat(reviews): add instance-level review-progress endpoint

valentin0h:
Powers the admin overview header above the proposal grid. Counts can't be derived from listProposalsWithReviewAggregates without iterating every page, so this is a dedicated query.


packages/common/src/services/decision/getPhaseReviewProgress.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  const instance = await getInstance({ instanceId: processInstanceId, user });
+
+  if (!instance.access.admin) {
+    throw new UnauthorizedError(
+      "You don't have admin access to this process instance",
+    );
+  }
+
+  const phaseId = input.phaseId ?? instance.currentStateId ?? undefined;
```
scazan: Can we remove this currentStateId? do we need this for legacy processes?

packages/common/src/services/decision/getPhaseReviewProgress.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  const [row] = await db
+    .select({
+      count: countDistinct(proposalReviewAssignments.proposalId),
+    })
+    .from(proposalReviewAssignments)
+    .where(
+      and(
+        eq(proposalReviewAssignments.processInstanceId, processInstanceId),
+        ...(phaseId ? [eq(proposalReviewAssignments.phaseId, phaseId)] : []),
```
scazan: without a phaseID are we just returning all of the phases here? Just checking that that is expected. Not an issue since we query against the passed proposalIds but then wondering if it's just defensive here.

packages/common/src/services/decision/getPhaseReviewProgress.ts
```diff
@@ -0,0 +1,155 @@
...
+          proposalReviewAssignments.status,
+          ProposalReviewAssignmentStatus.COMPLETED,
+        ),
+      ),
+    );
+
+  return row?.count ?? 0;
+}
+
+export function computeDaysLeft({
```
scazan, line 132: I wonder if we should just be soon adding a date library for this. Can you add a comment to the magic number or pull it out into a descriptive variable (86_400_000)

scazan (approved):

packages/common/src/services/decision/getPhaseReviewProgress.ts
valentin0h: Is `currentStateId` legacy?

packages/common/src/services/decision/getPhaseReviewProgress.ts
valentin0h: Actually, this is a bug. Good catch, updating.

packages/common/src/services/decision/getPhaseReviewProgress.ts
valentin0h: Good to have an answer on this, but it's not relevant anymore.

packages/common/src/services/decision/getPhaseReviewProgress.ts
valentin0h: Updated

PR #1093, opened 2026-05-01
fix(auth): surface SSO sign-in errors in /login UI

nourmalaeb:
- Redirect OAuth callback errors to `/login?error=<code>` instead of bare origin (which renders landing page for unauthed users).
- Add a client-side hash handler on `/` to catch provider-side errors (e.g. user cancels Google consent) that Supabase puts in the URL fragment.
- Map error codes to copy in `LoginPanel`; surface provider's `error_description` as a sub-message.
- Migrate the callback off deprecated `trpcVanilla` to `createClient`.

PR #1096, opened 2026-05-01
fix(auth): redirect SSO callback errors to /login

nourmalaeb:
- Bare-origin error redirects landed unauthed users on a route with no error UI
- LoginPanel already reads `?error=` from the query string — point at `/login` so the message renders
- 3-line change; minimal diff for the P0 hotfix

Splits PR #1093 — see `chore/auth-callback-trpc-client` and `feat/login-error-ux` for the rest of that work, both stacked on this branch.


scazan (approved):

scazan:
Let's add an e2e test for this perhaps

PR #1101, opened 2026-05-01
fix: Newline not working in comments when displaying

scazan:


scazan:
Duplicate PR to #1110

PR #1103, opened 2026-05-03
feat(decisions): add defaultProposalsHidden config

scazan:
## Summary
- Add `defaultProposalsHidden` config option to decision schema so admins can hide proposals on submission by default
- Apply hidden status during proposal submission when the config is enabled
- Add admin toggle in the Overview section of decision setup with i18n support

Closes Asana task 1213994771108172

## Test plan
- [x] Unit tests for submit proposal with `defaultProposalsHidden` enabled/disabled


scazan:
Closed as the approach was misguided. We want a different approach to this altogether.

PR #1111, opened 2026-05-05
fix: Fix proposal viewer 'Content could not be loaded' flash

scazan:
## Summary
- Removed `DocumentNotAvailable` synchronous fallback from `ProposalPreview` and `ProposalCardPreview` — empty/draft proposals now render nothing instead of flashing an error state
- Added `key={profileId}` to edit-page `ErrorBoundary` so it resets on navigation instead of staying tripped

## Test plan
- [ ] Create a draft proposal, go back to the proposal listing page — no flash of "Content could not be loaded"
- [ ] Submit a draft proposal, click "Back to proposals" — card renders without error flash
- [ ] View an individual proposal — no content flash on page load
- [ ] Verify proposals with actual content still render correctly

Fixes Asana task 1214539304463546


scazan:
Revision pushed. New commits:
- ca27bb130 RALPH: Fix "Content could not be loaded" flash across proposal flows

nourmalaeb (approved):

PR #1113, opened 2026-05-05
fix: Update process creator to use profileId of individual

scazan:
Updates the process creation to use the profile of the individual instead of the currentProfileId which could be an  organization. The organization relation is now based on the steward instead of the owner in this case.


scazan:
Revision pushed. New commits:
- 18c2fb1ed Fix getLegacyInstance test: create DECISION profile for access control

PR #1120, opened 2026-05-05
fix(a11y): stabilize baseline scan against render timing flake

nourmalaeb:
## Summary

- axe scan was racing render: fonts not yet loaded → `color-contrast` drift; Suspense not yet settled → `button-name`/`link-name` drift. Recent CI runs swung between 53–61 violations against a committed baseline of 70 with no related code changes.
- Fixes in `tests/e2e/tests/a11y-baseline.spec.ts`: wait for `document.fonts.ready`, wait for DOM mutations to quiet for 500ms, stop swallowing the `load`-event timeout
- Plus a hydration-warning console listener: if React emits a hydration mismatch during a route render, the route is marked errored rather than baselined
- Only posts a comment when there is deviation from the baseline.

Two consecutive local runs both produce 70/13/41/16, matching the committed baseline.

## Next: moving to a "known violations" model

This stabilizes the current count baseline, but the count approach is fragile by design — a fixed violation plus a new violation cancel out to zero net change, so swaps go undetected.

Follow-up on branch `a11y-known-violations` (already created off `dev`) will replace `summary.json` with `known-violations.json` — an identity-keyed list of `(rule, route, fingerprint)` tuples. CI fails on any violation not in the list and on any list entry that no longer fires. Each entry is debt to drive to zero as part of the WCAG 2.1 AA sprint, and the existing `summary.json` becomes the seed for the initial list.

---
  - https://app.asana.com/0/0/1214453886143072

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -461,6 +461,7 @@ jobs:
             || echo '{}' > /tmp/base-summary.json
 
       - name: Render a11y diff
+        id: render-diff
         if: ${{ github.event_name == 'pull_request' && !cancelled() }}
         run: |
           pnpm tsx scripts/a11y-baseline-diff.ts \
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yml
valentin0h, line 1: nice 💯

valentin0h (approved):

PR #1123, opened 2026-05-06
feat(profiles): Add locations to all entities

scazan:
Adds in the ability to attach a Location to any profile of any type. This enables GIS functionality on all entities. This only adds the database changes, the backend changes will come separately.
We also support multiple locations attached to an entity.


scazan:
Revision pushed. New commits:
- 6c38e4299 Fix migration: remove CONCURRENTLY from transactional migration
- 162e304b0 RALPH: Add location column to proposals and proposal history

services/db/schema/tables/profiles.sql.ts
```diff
@@ -93,6 +101,44 @@ export const profilesRelations = relations(profiles, ({ many, one }) => ({
...
+      fields: [profilesLocations.profileId],
+      references: [profiles.id],
+    }),
+    location: one(locations, {
+      fields: [profilesLocations.locationId],
+      references: [locations.id],
+    }),
+  }),
+);
+
```
valentin0h:
```suggestion

```

services/db/schema/tables/profiles.sql.ts
```diff
@@ -93,6 +101,44 @@ export const profilesRelations = relations(profiles, ({ many, one }) => ({
...
+      .references(() => profiles.id, {
+        onUpdate: 'cascade',
+        onDelete: 'cascade',
+      }),
+    locationId: uuid('location_id')
+      .notNull()
+      .references(() => locations.id, {
+        onUpdate: 'cascade',
+        onDelete: 'cascade',
+      }),
```
valentin0h, line 121: thought: we might need to somehow distinguish between the locations for ordering, filtering, etc. We are missing timestamps and some flagging what's a primary

valentin0h (approved):

services/db/schema/tables/profiles.sql.ts
scazan: I'll add in the timestamps.. not sure what primary means yet though so I think we can add it when we need it.

PR #1125, opened 2026-05-07
fix(a11y): require accessible name on IconButton

nourmalaeb:
- IconButton, SidebarTrigger, and OptionMenu now require `aria-label` or `aria-labelledby` via a shared `RequireAccessibleName` type helper
- Added context-specific `t()`'d labels + translations (en/es/fr/pt/bn) to all existing icon-only button callers
- Also fixes label visibility to screen readers on loading buttons


packages/ui/src/components/IconButton.tsx
```diff
@@ -31,13 +31,13 @@ const iconButtonStyle = tv({
...
 
-export interface IconButtonProps
-  extends
-    Omit<React.ComponentProps<typeof RACButton>, 'children'>,
-    IconButtonVariants {
-  children: React.ReactNode;
-  className?: string;
-}
+export type RequireAccessibleName<T> = T &
+  ({ 'aria-label': string } | { 'aria-labelledby': string });
```
valentin0h: suggestion: lift it up into some other file because it's not specific to `IconButton` and all other components end up importing from `IconButton`

valentin0h (approved):

packages/ui/src/components/IconButton.tsx
nourmalaeb: good suggestion, done

PR #1132, opened 2026-05-07
fix(a11y): bump core color tokens for WCAG contrast

nourmalaeb:
## Summary

- Darkens teal, red, and green semantic tokens (and their *Black variants) in `packages/styles/shared-styles.css` to meet AA contrast on white
- Leaves `--color-data-purple` / `--color-data-blue` at `-500` since deeper shades aren't defined in `tokens.css` and -500 is already AAA

### DIFF
--- a/apps/app/src/components/LoginPanel.tsx
+++ b/apps/app/src/components/LoginPanel.tsx
@@ -252,9 +252,9 @@ export const LoginPanel = () => {
                       </Button>
 
                       <div className="flex w-full items-center justify-center gap-4 text-darkGray">
-                        <div className="h-px grow bg-current" />
+                        <div className="h-px grow bg-current opacity-50" />
                         <span>{t('or')}</span>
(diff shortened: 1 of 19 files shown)


apps/app/src/components/LoginPanel.tsx
```diff
@@ -251,10 +251,10 @@ export const LoginPanel = () => {
                         {t('Continue with Google')}
                       </Button>
 
-                      <div className="flex w-full items-center justify-center gap-4 text-midGray">
-                        <div className="h-px grow bg-current" />
+                      <div className="flex w-full items-center justify-center gap-4 text-darkGray">
+                        <div className="h-px grow bg-current opacity-50" />
```
nourmalaeb, line 255: the borders felt too dark now that the `or` string was darkened for legibility... so i knocked back the opacity.

scazan (approved):
nice!

PR #1134, opened 2026-05-08
feat(reviews): honor reviewsAllowRevisions toggle on review page

valentin0h:
Hide the Request Revision button when the process config disables revisions, so toggling the setting in the process builder takes effect for reviewers.

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewLayout.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewLayout.tsx
@@ -23,16 +23,20 @@ export async function ReviewLayout({
 }: ReviewLayoutProps) {
   const { utils, queryClient } = await createServerUtils();
 
-  await Promise.all([
+  const [decision] = await Promise.all([
+    utils.decision.getDecisionBySlug.fetch({ slug: decisionSlug }),
     utils.decision.getReviewAssignment.prefetch({ assignmentId }),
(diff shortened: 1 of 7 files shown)


valentin0h (commented):
We're not doing a backend check for now; it's not worth it at this point.

apps/app/src/components/decisions/Review/ReviewLayout.tsx
```diff
@@ -21,15 +24,38 @@ export async function ReviewLayout({
...
 
-  await utils.decision.getReviewAssignment.prefetch({ assignmentId });
+  let decisionProfile;
+  try {
+    [decisionProfile] = await Promise.all([
+      client.decision.getDecisionBySlug({ slug: decisionSlug }),
+      utils.decision.getReviewAssignment.prefetch({ assignmentId }),
+    ]);
+  } catch (error) {
+    const cause = error instanceof Error ? error.cause : null;
```
scazan, line 39: Does this end up throwing a CommonError or a TRPC error since we're using API here? I have a memory that it is a TRPC error.

scazan (approved):

apps/app/src/components/decisions/Review/ReviewLayout.tsx
valentin0h: Added a test to be sure.

PR #1142, opened 2026-05-10
feat: Allow manual selection into decision results

scazan:
## Summary
- Run results processing inline in the same transaction as `submitManualSelection` so the selection set and the persisted `decision_process_results` row land atomically.
- Skip the final phase's selection pipeline in `processResults` — selection happens upstream (prior phase's pipeline or admin manual selection); the writer just persists what advanced ( we write a new row to ensure that results are tracked as a process could theoretically still be edited at some point).
- Make `processResults` append-only: every run inserts a new row with its own `executedAt`, and read sites pick the latest. Prior outcomes stay in the table as an audit trail across re-runs and across schema edits that shift which phase is "final."
- Stop mutating `proposals.status` from `processResults`. The selections table is the source of truth for "which proposals advanced"; `proposals.status` stays in its existing lifecycle and is left untouched here.

https://github.com/oneprojectorg/common/pull/1145 adds the results-specific UI on top of this function


packages/common/src/services/decision/processResults.ts
```diff
@@ -1,175 +1,173 @@
...
-      );
-    }
-
-    const instanceData = processInstance.instanceData as DecisionInstanceData;
-    const currentPhaseId = processInstance.currentStateId;
-
-    const processProposals = await getProposalsForPhase({
-      instanceId: processInstanceId,
+  tx?: DbClient;
+  instance?: PhaseScopedInstance;
```
valentin0h, line 30: suggestion: we can improve the signature here to be something like `processInstanceId | processInstance`

packages/common/src/services/decision/processResults.ts
```diff
@@ -1,175 +1,173 @@
...
+  processInstanceId,
+  preloadedInstance,
+}: {
+  tx: DbClient;
+  processInstanceId: string;
+  preloadedInstance?: PhaseScopedInstance;
+}): Promise<void> {
+  const resolvedInstance =
+    preloadedInstance ??
+    (await loadPhaseScopedInstance({ db: tx, processInstanceId }));
```
valentin0h, line 84: suggestion: lift this up so that we limit functions with the type of kind 

```typescript
  processInstanceId: string;
  preloadedInstance?: PhaseScopedInstance;
```

packages/common/src/services/decision/processResults.ts
```diff
@@ -1,175 +1,173 @@
...
+
+  const [selectedProposalIds, voterCount] = await Promise.all([
+    resolvedInstance
```
valentin0h: slight regression: we need to throw if there's no `resolvedInstance`

valentin0h (approved):
lgtm 👍

packages/common/src/services/decision/processResults.ts
scazan: I can switch the naming to processInstance

PR #1143, opened 2026-05-10
chore(db): unique index on decision_process_results.process_instance_id

scazan:
## Summary
Swap the composite `(process_instance_id, executed_at)` index on `decision_process_results` for a UNIQUE index on `process_instance_id` so the row can be safely upserted on conflict. This ensures that we only have one result per process at any time.

This is a structural prerequisite for #1142 (stacked on top of this PR), which moves results processing to `INSERT ... ON CONFLICT (process_instance_id) DO UPDATE`. Shipping the index on its own keeps the schema change isolated and reversible.

* No existing rows in `decision_process_results` violate the new uniqueness constraint


scazan:
Pivoting to append-only writes for decision_process_results so we keep an audit trail when phases are added/removed after results are processed. The original (process_instance_id, executed_at) non-unique index already supports the 'latest by timestamp' lookup, so no migration is needed. PR #1142 will rebase off dev and switch to plain INSERT.

PR #1145, opened 2026-05-11
feat: Result-specific Phase Selection Screens

scazan:
Adds in the results-specific advance screen for final phases which differentiate slightly from the other advance screens.

#### demo
Screenshot 2026-05-11 at 13 11 56

##### Results modal after selection (confetti not in screenshot but is active)
Screenshot 2026-05-13 at 17 50 57


apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -25,70 +25,61 @@ interface DecisionHeaderProps {
...
-        }}
+        backTo={{ href: '/decisions' }}
         title={
           profileName ||
           instance.name ||
-          instanceData.templateName ||
-          (instance as any).process?.name
+          instance.instanceData?.templateName ||
+          instance.process?.name ||
+          ''
```
scazan: Maybe we should set a default here. "Untitled"

apps/app/src/components/decisions/FinalPhaseSelectionConfirmDialog.tsx
```diff
@@ -0,0 +1,104 @@
+'use client';
```
scazan: Does this need to be 'use client' since we only use useTranslations()? This could potentially be used in a server tree as well.

apps/app/src/components/decisions/FinalPhaseSelectionFooter.tsx
```diff
@@ -0,0 +1,50 @@
+'use client';
```
scazan, line 1: Same here. We possibley don't need use client as we can use TranslatedText

apps/app/src/components/decisions/pages/FinalPhaseManualSelectionPage.tsx
```diff
@@ -0,0 +1,73 @@
+'use client';
+
+import { APIErrorBoundary } from '@/utils/APIErrorBoundary';
+import { EmptyState } from '@op/ui/EmptyState';
+import { Header3 } from '@op/ui/Header';
+import { Suspense } from 'react';
+import { LuTriangleAlert } from 'react-icons/lu';
+
+import { useTranslations } from '@/lib/i18n/routing';
```
scazan, line 9: Do we need use client here?

apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -25,70 +25,61 @@ interface DecisionHeaderProps {
...
-      name: p.name || templateState?.name,
+      name: p.name || templateState?.name || '',
       description: p.description || templateState?.description,
-      type: templateState?.type,
-      config: templateState?.config,
-      phase: templateState?.phase || {
-        startDate: p.startDate,
-        endDate: p.endDate,
+      phase: {
+        startDate: p.startDate ?? templateState?.startDate,
```
scazan: We don't want to fall back to the template startDate here. startDate is not required so we should be able to leave it undefined.

apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -25,70 +25,61 @@ interface DecisionHeaderProps {
...
 
-export async function DecisionHeader({
+export function DecisionHeader(props: DecisionHeaderProps) {
+  if (props.useLegacy) {
+    return <LegacyDecisionHeaderContent {...props} />;
+  }
+  return <NewDecisionHeaderContent {...props} />;
+}
+
+function NewDecisionHeaderContent({
```
scazan: We should never call these "New". This is the normal case whereas Legacy is an old case.. so this can just be DecisionHeaderContent

apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -25,70 +25,61 @@ interface DecisionHeaderProps {
...
+  const [instance] = trpc.decision.getInstance.useSuspenseQuery({ instanceId });
 
-  const instanceData = instance.instanceData as DecisionInstanceData;
-  const instancePhases = instanceData.phases ?? [];
-
-  // For legacy instances, fall back to process.processSchema states/phases for names.
-  const processSchema = (instance as any).process?.processSchema;
-  const templateStates = processSchema?.states || processSchema?.phases || [];
+  const instancePhases = instance.instanceData?.phases ?? [];
+  const templateStates = instance.process?.processSchema?.phases ?? [];
```
scazan: New processes do not have the processSchema so we shouldn't use it here. only legacy processes have that and now we are already branching legacy processes off earlier in the tree.

packages/common/src/services/decision/listSelectionCandidates.ts
```diff
@@ -88,19 +104,58 @@ export async function listSelectionCandidates({
...
+  const withVotes: Proposal[] = enriched.map((p) => ({
+    ...p,
+    voteCount: voteCountByProposalId.get(p.id) ?? 0,
+  }));
+
+  // The DB returned createdAt order (newest/oldest); for `votes`, re-sort here
+  // — vote counts aren't a DB column on `proposals`, so adding a sort to the
+  // candidate query would require a join. Sorting a bounded result set in JS
+  // keeps the query simple.
+  if (sortOrder === 'votes') {
```
nourmalaeb: Should we add a test for this?

apps/app/src/components/decisions/pages/FinalPhaseManualSelectionPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+            "We've selected the suggested winning proposals based on voting rules and total budget.",
+          )}
+          variant="standard"
+          gradient={HERO_GRADIENT}
+        />
+      </div>
+
+      <div className="mt-8 flex w-full justify-center border-t bg-white">
+        <div className="w-full gap-8 p-4 sm:max-w-6xl sm:p-8">
+          <div className="flex flex-col gap-6 lg:col-span-3">
```
nourmalaeb: Something funky with these tailwind classNames ... `col-span-3` won't do anything if the ancestor is not a `grid`. Am I missing something?

apps/app/src/components/decisions/ManualSelectionList.tsx
```diff
@@ -6,18 +6,22 @@ import { ProposalFilter } from '@op/api/encoders';
 import type { Proposal } from '@op/common/client';
 import { Button } from '@op/ui/Button';
 import { EmptyState } from '@op/ui/EmptyState';
-import { FooterBar } from '@op/ui/FooterBar';
 import { Header3 } from '@op/ui/Header';
 import { toast } from '@op/ui/Toast';
+import { useRouter } from 'next/navigation';
```
nourmalaeb: Should we import our i18n version of `useRouter`?

apps/app/src/components/decisions/FinalPhaseSelectionFooter.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+const FinalPhaseProposalCard = ({ proposal }: { proposal: Proposal }) => {
+  const t = useTranslations();
+  const { title, budget, categories, submitterName } =
+    resolvePresentationFields({
+      proposal,
+      defaultTitle: t('Untitled Proposal'),
+    });
+  const categoryLabel = categories[0];
```
nourmalaeb: do we need to show all categories if the proposal has more than 1 category?

nourmalaeb (approved):

PR #1148, opened 2026-05-11
Improve decisions/proposals skeleton UX

scazan:
## Summary

Tightens the loading experience across the decisions area. Skeletons now match the real layout (no flash on handoff), and queries no longer waterfall unnecessarily.

- **`/decisions`** — collapsed the dual skeleton flash into one. Drafts-existence check is now non-suspending so the tab bar renders immediately; the drafts tab appears when the check resolves.
- **`/decisions/[slug]`** — removed duplicate `getInstance` fetch (it's resolved once in `DecisionStateRouter` and passed down). `DecisionPageSkeleton`'s proposal block now uses the same `ProposalListSkeleton` shape as the real grid, so there's no re-layout when content lands.
- **`/decisions/[slug]/proposal/[profileId]`** — replaced the inline gray skeleton with a layout-matching `ProposalViewSkeleton` (header, title, metadata, author row, engagement bar, body). Streamed comments behind their own Suspense boundary with a dedicated `ProposalCommentsSkeleton`, and parallelized the comments fetch via `usePrefetchQuery` so the cache is warm by the time `ProposalComments` reads it.
- Moved shared skeletons (`ProposalCardSkeleton`, `ProposalListSkeleton`, `ProposalViewSkeleton`, `ProposalCommentsSkeleton`) into `components/skeletons/ProposalSkeleton.tsx` so route-level `loading.tsx` and in-page Suspense fallbacks render the same component.
- Reordered touched files to put primary exports first per `component-file-structure` guidance.

## Test plan

- [ ] `pnpm w:app typecheck` passes
- [ ] Walk `/decisions` with DevTools Network throttled to Slow 4G — confirm exactly one skeleton state shows, tab pills appear in the skeleton, no second flash
- [ ] On `/decisions/[slug]`, confirm only one `getInstance` request in Network, and the page-level skeleton's grid matches the real grid (no layout shift)
- [ ] On `/decisions/[slug]/proposal/[profileId]`, confirm skeleton shape lines up with real content and comments render after the body with their own skeleton


scazan:
Superceded by a PR we built together in a coworking session

PR #1151, opened 2026-05-11
feat(decisions): show revision-requested CTA on author's cards in review phase

valentin0h:
Authors lose track of which of their proposals need attention once reviewers request revisions. This adds a "Revision requested" chip and primary "Revise proposal" CTA on the author's own cards during the review phase, gated behind a non-blocking query that only fires in that phase.

## Demo
CleanShot 2026-05-13 at 11 13 12

### DIFF
--- a/apps/app/src/components/decisions/ProposalsList.tsx
+++ b/apps/app/src/components/decisions/ProposalsList.tsx
@@ -165,18 +165,10 @@ interface ProposalsProps {
   permissions?: DecisionAccess | null;
   votedProposalIds?: string[];
   hasFilter: boolean;
-  /**
-   * When true, the current phase has voting enabled — always show voting UI.
-   * Derived from `currentPhase.rules.voting.submit` upstream.
-   */
+  /** When true, the current phase has voting enabled — always show voting UI */
   isVotingPhase?: boolean;
   /** When true, new proposals are hidden by default in the current phase. */
   proposalsHidden?: boolean;
-  /**
-   * Proposal IDs for which the current user has a pending revision request.
-   * Empty in non-review phases.
-   */
-  revisionRequestedProposalIds?: Set<string>;
 }
 
 const VotingProposalsList = ({
@@ -452,7 +444,13 @@ const ViewProposalsList = ({
   hasFilter,
   proposalsHidden,
   revisionRequestedProposalIds,
-}: ProposalsProps) => {
+}: ProposalsProps & {
+  /**
+   * Proposal IDs for which the current user has a pending revision request.
+   * Only computed in review phases; empty otherwise.
+   */
+  revisionRequestedProposalIds?: Set<string>;
+}) => {
   const canManageProposals = permissions?.admin ?? false;
   if (!proposals || proposals.length === 0) {
     if (proposalsHidden && !hasFilter) {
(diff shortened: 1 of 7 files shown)


apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -556,11 +579,13 @@ export const ProposalsList = ({
   initialFilter?: ProposalFilter;
   /** When set to 'results', all proposals are returned as non-editable */
   phase?: 'results';
-  /** When true, the current phase has voting enabled — always show voting UI */
-  isVotingPhase?: boolean;
+  /** Current phase; capability flags are derived from `rules`. */
+  currentPhase?: InstancePhaseData;
```
valentin0h, line 591: Let's not make this optional; it's a bigger refactor. To be done separately.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -530,7 +556,12 @@ const Proposals = (props: ProposalsProps) => {
     return <VotingProposalsList {...props} />;
   }
 
-  return <ViewProposalsList {...props} />;
+  return (
+    <ViewProposalsList
+      {...props}
+      revisionRequestIdByProposalId={revisionRequestIdByProposalId}
```
scazan, line 562: Any reason we are separating and then re-adding this after spreading?

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -653,6 +686,18 @@ export const ProposalsList = ({
   const { proposals: allProposals } = proposalsData ?? {};
   const canManageProposals = permissions?.admin ?? false;
 
+  const { data: revisionRequestsData } =
```
scazan, line 689: This component is getting thick with all functionality. I wonder why we don't try to compose this better as we are mixing all the capability checks (and queries) into it. We should branch on components. But I understand it's a bigger refactor to do that so ok without that in this PR.

apps/app/src/components/decisions/RevisionRequestChip.tsx
```diff
@@ -0,0 +1,15 @@
+import { Chip } from '@op/ui/Chip';
```
scazan, line 1: Do we need use client here since this uses useTranslations

scazan (approved):
no blocking comments.

apps/app/src/components/decisions/ProposalsList.tsx
valentin0h: Yeah, it definitely needs a refactor.

apps/app/src/components/decisions/RevisionRequestChip.tsx
valentin0h: We don't, but it's such a small component under a client component that it's not worth overthinking it, I think. We can tell the agent to prefer server components when it can.

apps/app/src/components/decisions/ProposalsList.tsx
valentin0h: So that we keep `return <VotingProposalsList {...props} />;` as is. I'll have a look if it's worth doing it the other way with ...restProps.

PR #1152, opened 2026-05-12
fix(decisions): re-anchor review assignment after revision response

valentin0h:
Reviewers kept seeing the pre-revision proposal after the author resubmitted because the assignment's history snapshot was never updated past phase-transition time.


nourmalaeb (approved):
Nice!

PR #1154, opened 2026-05-12
fix(decisions): scroll to top on process builder section change

nourmalaeb:
## Summary
- Next/Back/sidebar navigation only flips a `?section=` query param, so the scrollable container kept its prior scroll position
- Reset all scrollable ancestors to top whenever `currentSection` changes

---
  - https://app.asana.com/0/0/1213823490678546

### DIFF
--- a/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
+++ b/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
@@ -33,7 +33,10 @@ export function ProcessBuilderContent({
     while (el) {
       const { overflowY } = getComputedStyle(el);
       if (overflowY === 'auto' || overflowY === 'scroll') {
-        el.scrollTo({ top: 0 });
+        if (el.scrollTop !== 0) {
+          el.scrollTo({ top: 0 });
+        }
+        break;
       }
       el = el.parentElement;
     }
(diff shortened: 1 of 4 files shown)


apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderContent.tsx
```diff
@@ -25,6 +26,22 @@ export function ProcessBuilderContent({
   const access = useUser();
   const isAdmin = access.getPermissionsForProfile(decisionProfileId).admin;
 
+  const wrapperRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    let el: HTMLElement | null = wrapperRef.current;
+    while (el) {
+      const { overflowY } = getComputedStyle(el);
+      if (overflowY === 'auto' || overflowY === 'scroll') {
```
scazan: Little worried about an infinite loop here. Maybe good to comment on why we are only breaking the loop on 'auto' or 'scroll' only. Could there be an easy regression down the line where we set something to hidden, clip, visible, inherit, etc. that would cause this to become a crash.

nourmalaeb: Fair point. This was a convoluted way of working around the element we want to target being in `page.tsx` which is a server component. I refactored it so that the process builder edit area is in a separate client component that can manage scroll. Ready for re-review.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
```diff
@@ -380,7 +380,7 @@ export function TemplateEditorContent({
 
   return (
     <SidebarProvider isOpen={sidebarOpen} onOpenChange={setMobileSidebarOpen}>
-      <div className="flex h-full flex-col md:flex-row">
+      <div className="flex h-full flex-col overflow-hidden md:flex-row">
```
nourmalaeb, line 383: This fixes a separate Sidebar-height bug that was causing a double-scrollbar in some situations

scazan (approved):

PR #1155, opened 2026-05-12
fix(decisions): wire up review-phase "Learn more" CTA

nourmalaeb:
## Summary
- Replace dead `href="#"` placeholder in `ReviewPage` with `DecisionActionBar`, matching `VotingPage` / `StandardDecisionPage`.
- CTA label switches to "About this phase" when modal content comes from `currentPhase.additionalInfo` (or its translation); falls back to "About the process" for instance-level descriptions.
- Adds e2e assertion that the CTA renders and opens the modal.
- Empty description chain → button hidden (graceful).

---
  - https://app.asana.com/0/0/1214137998286187

### DIFF
--- a/apps/app/src/components/decisions/pages/ReviewPage.tsx
+++ b/apps/app/src/components/decisions/pages/ReviewPage.tsx
@@ -8,6 +8,8 @@ import { Header3 } from '@op/ui/Header';
 import { Suspense } from 'react';
 import { LuLeaf } from 'react-icons/lu';
 
+import { useTranslations } from '@/lib/i18n/routing';
+
 import { TranslatedText } from '@/components/TranslatedText';
 
 import { DecisionActionBar } from '../DecisionActionBar';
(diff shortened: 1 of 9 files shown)


apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -40,6 +45,18 @@ export function ReviewPage({
   const canReview = Boolean(instance.access?.review);
   const isAdmin = Boolean(instance.access?.admin);
 
+  const t = useTranslations();
+  const translation = useDecisionTranslation();
+  const description =
+    instance.description ?? instance.instanceData?.templateDescription;
+  const phaseAdditionalInfo =
+    translation?.additionalInfo ?? currentPhase.additionalInfo;
```
scazan, line 53: Not for this PR, but I see this often enough that I'm wondering if we really just need a function that handles this for us.

apps/app/src/components/decisions/pages/ReviewPage.tsx
```diff
@@ -66,11 +83,13 @@ export function ReviewPage({
...
-            <div className="flex justify-center pt-2">
-              <ButtonLink color="secondary" size="medium" href="#">
-                <TranslatedText text="Learn more" />
-              </ButtonLink>
-            </div>
+            <DecisionActionBar
+              instanceId={instance.id}
+              description={actionBarDescription}
+              label={actionBarLabel}
+              markup={!!translation?.additionalInfo}
```
scazan, line 90: This feels like a bit of an odd logic that we just know that additionalInfo from the translation will be markup enabled. Wondering if we should reconsider that. I believer the original DecisionActionBar markup was meant to be used for a special case where a pilot needed a custom HTML text.

scazan (approved):
Looks good. Just curious about the markup part there. Do we still need it for the translation?

scazan:
Shall we merge this one in? no blockers on it.

PR #1158, opened 2026-05-14
fix(decisions): show all valid proposals on results "All proposals" tab

valentin0h:
fix(decisions): show all valid proposals on results "All proposals" tab
feat(decisions): add cursor-paginated listAllProposals procedure
refactor(decisions): share relationship-data fetching between list services
feat(decisions): derive isSelected from the latest results selection

TODO (separate PRs):
- new e2e tests
- infinite scroll in proposals list (missing for existing lists)

## Demo
CleanShot 2026-05-21 at 12 04 11


apps/app/src/components/decisions/ProposalCard/ProposalCardComponents.tsx
```diff
@@ -344,20 +344,26 @@ export function ProposalCardStatus({
...
+  // "Funded" reflects whether the proposal landed in the latest results
+  // selection — independent of the editorial `status` column, which an admin
+  // can edit freely.
+  if (isSelected) {
+    return (
       <>
         <Bullet />
         <span className={cn('text-sm text-green-700', className)}>
-          {t('Shortlisted')}
+          {t('Funded')}
```
valentin0h: Did we want this to be `Selected`?

services/api/src/routers/decision/proposals/list.ts
```diff
@@ -22,4 +25,21 @@ export const listProposalsRouter = router({
 
       return proposalListSchema.parse(result);
     }),
+  /** Lists all proposals for a given process instance, no phase filter. */
+  listAllProposals: commonAuthedProcedure()
```
valentin0h, line 30: New procedure, it was getting unwieldy with `listProposals`

packages/common/src/services/decision/listAllProposals.ts
```diff
@@ -0,0 +1,201 @@
...
+  const [relationshipData, documentContentMap, selectedIds] = await Promise.all(
+    [
+      getProposalRelationshipData({ profileIds, currentProfileId }),
+      getProposalDocumentsContent(
+        pageItems.map((proposal) => ({
+          id: proposal.id,
+          proposalData: proposal.proposalData,
+          proposalTemplate,
+        })),
+      ),
```
valentin0h, line 148: Not versioned, always get latest

apps/app/src/components/decisions/ProposalCard/ProposalCardComponents.tsx
scazan: yes!

apps/app/src/components/decisions/ProposalCard/ProposalCardComponents.tsx
```diff
@@ -344,20 +344,24 @@ export function ProposalCardStatus({
...
+    );
+  }
+
+  return match(status, {
+    [ProposalStatus.APPROVED]: (
       <>
         <Bullet />
         <span className={cn('text-sm text-green-700', className)}>
-          {t('Funded')}
+          {t('Shortlisted')}
```
scazan, line 362: Just noting that we will remove the shortlisted status as well down the line

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -680,10 +680,28 @@ export const ProposalsList = ({
     return params;
   }, [instanceId, selectedCategory, sortOrder, phase]);
 
-  const [proposalsData] =
-    trpc.decision.listProposals.useSuspenseQuery(queryParams);
+  // Split per phase — TS can't unify the two procedures' output shapes
+  // inside one useSuspenseQueries call.
+  const [[resultsProposals]] = trpc.useSuspenseQueries((t) => [
+    ...(phase === 'results'
```
scazan, line 689: I would prefer this to be a bit higher up if we can. But fine for now.. I would like to mostly avoid conditionals at this level for phases as it will get messy (ProposalsResultslist or something). Also this file is getting way too much stuff in it and would be good to start breaking out.

Not a blocker for this one though.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -680,10 +680,28 @@ export const ProposalsList = ({
...
+            processInstanceId: queryParams.processInstanceId,
+            dir: queryParams.dir,
+            limit: queryParams.limit,
+            categoryId: queryParams.categoryId,
+          }),
+        ]
+      : []),
+  ]);
+  const [[phaseProposals]] = trpc.useSuspenseQueries((t) => [
+    ...(phase === 'results' ? [] : [t.decision.listProposals(queryParams)]),
```
scazan, line 701: Any reason not to just include this in the prior useSuspenseQueries?

packages/common/src/services/decision/getProposalRelationshipData.ts
```diff
@@ -0,0 +1,123 @@
...
+        })
+        .from(profileRelationships)
+        .where(
+          and(
+            eq(profileRelationships.sourceProfileId, currentProfileId),
+            inArray(profileRelationships.targetProfileId, profileIds),
+          ),
+        ),
+
+      db
```
scazan, line 72: Should we include vote counts in here? Thinking about the other issue of wrong vote counts showing. Was thinking we just need to make sure it is gated on a results record existing so one can't expose it early. I have a separate PR for that open but maybe it could make sense with what you have here.

packages/common/src/services/decision/getProposalRelationshipData.ts
```diff
@@ -0,0 +1,123 @@
...
+          count: countFn(),
+        })
+        .from(profileRelationships)
+        .where(inArray(profileRelationships.targetProfileId, profileIds))
+        .groupBy(
+          profileRelationships.targetProfileId,
+          profileRelationships.relationshipType,
+        ),
+
+      db
```
scazan, line 59: Maybe not, but have the sense these could be a single query.. however, it'd add maybe too much complexity.

services/api/src/routers/decision/proposals/list.ts
scazan: I'm wondering, since we have a new procedure, we could remove some of the legacy stuff and just keep calling the current endpoint for legacy processes. Then we can keep the new stuff clean.

scazan (approved):
Nice. Nice you did some cleanup on it too pulling into functions.

No blockers, just discussion items. Vote counts might be a good separate addition.

apps/app/src/components/decisions/ProposalsList.tsx
valentin0h: Yeah this file needs a big refactor 😅

apps/app/src/components/decisions/ProposalsList.tsx
valentin0h: Typing issues, I think there's a comment for it. They'll get batched

PR #1175, opened 2026-05-16
fix(decisions): show max points in Total Score on reviews

scazan:
## Summary

- `TotalScoreCard` in the review form rendered only the running sum (e.g. `8`), with no in-context anchor for how many points the rubric was worth. Reviewers had to infer the maximum from each criterion's pts label.
- Format the value as `score/total` (e.g. `8/10`). The maximum comes from `getRubricScoringInfo(rubricTemplate).totalPoints` — the same canonical helper already used by `ReviewSummaryPanel`, `ReviewerList`, `ReviewerDetail`, and `ReviewSelectionList`. No new aggregation logic is introduced.
- Empty-state (`null` score) becomes `–/{total}`. If the rubric has no scored criteria (e.g. yes/no + long-text only), `totalPoints === 0` and the card falls back to the bare score, preserving the previous behavior.
- The existing e2e assertion in `tests/e2e/tests/review-submit.spec.ts` is updated from `/^6$/` to `/^6\/8$/` (Innovation max=5 + Feasibility max=3 = 8).

Asana: https://app.asana.com/0/1209477276073375/1214705378291854

## Test plan

- [ ] Open a review form on a rubric with scored criteria — the Total Score card shows `{score}/{maxPoints}` (e.g. `0/10` initially, `8/10` after filling all criteria).
- [ ] Open a review form on a rubric with no scored criteria — the card shows just `–` (unchanged behavior).
- [ ] `pnpm w:app typecheck` passes.
- [ ] `pnpm test` (under `CI=1`) — all 676 @op/api + 52 @op/common + 2 @op/realtime tests pass.
- [ ] `pnpm exec playwright test review-submit.spec.ts --grep "full review journey"` passes.

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
@@ -42,7 +42,7 @@ export function TotalScoreCard({
   }, null);
 
   const scoreText = totalScore === null ? '–' : String(totalScore);
-  const display = totalPoints > 0 ? `${scoreText}/${totalPoints}` : scoreText;
+  const display = totalPoints > 0 ? `${scoreText}/${totalPoints}` : '–';
 
   return (
     <Surface
--- a/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewFormShell.tsx
@@ -1,4 +1,5 @@
 import type { RubricReviewData, RubricTemplateSchema } from '@op/common/client';
+import { getRubricScoringInfo } from '@op/common/client';
 import { Header3 } from '@op/ui/Header';
 import { Surface } from '@op/ui/Surface';
 import type { ReactNode } from 'react';
@@ -28,6 +29,7 @@ export function TotalScoreCard({
   values: RubricReviewData['answers'];
 }) {
   const criteria = getCriteria(rubricTemplate);
+  const { totalPoints } = getRubricScoringInfo(rubricTemplate);
 
   const totalScore = criteria.reduce<number | null>((total, criterion) => {
     const value = values[criterion.id];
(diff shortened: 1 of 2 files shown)


apps/app/src/components/decisions/Review/ReviewFormShell.tsx
```diff
@@ -39,6 +41,9 @@ export function TotalScoreCard({
     return (total ?? 0) + value;
   }, null);
 
+  const scoreText = totalScore === null ? '–' : String(totalScore);
+  const display = totalPoints > 0 ? `${scoreText}/${totalPoints}` : scoreText;
```
nourmalaeb: If `totalPoints` is 0, then doesn't that mean we don't need to show a score at all?

nourmalaeb (changes requested):
I'm not sure about the logic for displaying the total score. I think we should omit it entirely if there are no scored criteria in the rubric.

apps/app/src/components/decisions/Review/ReviewFormShell.tsx
scazan: Actually this is a strange one on second thought. this should theoretically not happen. We can fallback to '-' though in that case then?

scazan: even more odd.. what if scoreText > 0 and totalPoints is not... again.. would be odd but perhaps a situation where we'd still want to surface the score rather than hide it.. I'm split on this.

nourmalaeb: That's the thing. There are some criteria that don't get scored (like a text input). So you could have a rubric with no score. In which case I think it's best to just hide it.

scazan: Cool. I'll hide it for now.

nourmalaeb (approved):

PR #1180, opened 2026-05-18
feat(resources): backend for decision resources

scazan:
## Summary
- Adds `resources` and `resource_collections` tables with Drizzle relations and migration
- Adds resources service layer (collections, ordering, storage, access) and tRPC routers (CRUD, reorder, pin, upload, collections)
- Exposes link preview helper via the content router

## Suggested reading order
This PR is large but tightly scoped to one feature. For a faster review pass, read in this order rather than file-by-file:

1. **Schema** — `services/db/schema/tables/resources.sql.ts`, `resourceCollectionProfiles.sql.ts`, and the new migration under `services/db/migrations/`. Adds a single update for RESTRICT cascade instead.
2. **Access control** — `packages/common/src/services/resources/access.ts` and `schemas.ts`. The authorization model used everywhere downstream.
3. **Core service** — `packages/common/src/services/resources/resources.ts` and `collections.ts`. The bulk of the business logic.
4. **Supporting concerns** — `ordering.ts`, `storage.ts`, `channelScope.ts`, `constants.ts`, plus the shared `packages/common/src/utils/sorting.ts` util.
5. **tRPC routes** — `services/api/src/routers/resources/*`. Thin wrappers; mostly input validation + service calls.
6. **Tests** — `resources.test.ts`, `collections.test.ts`, `resourcesTestUtils.ts`. Exercise the surface through tRPC.

## Stacked PR
Frontend changes are in a follow-up PR here: https://github.com/oneprojectorg/common/pull/1181 . You can try it out if you like there as it is working but in need of style updates and more review of the code.


packages/common/src/services/resources/createDocumentResource.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type CreateDocumentInput = {
+  authUserId: string;
+  profileId?: string;
+  collectionId?: string;
+  title: string;
+  description: string | null;
+  storagePath: string;
+  fileName: string;
+  mimeType: string;
```
scazan, line 26: We allowed the user to self-report the MIME type for now. This avoids us having to read the first bytes of the file in storage. However, we could do that and I think we probably should do that in a follow-up.

scazan:
Superceded by https://github.com/oneprojectorg/common/pull/1230/commits

PR #1184, opened 2026-05-18
fix(decisions): upload proposal attachments via signed URL

nourmalaeb:
Proposal attachments larger than ~4.5MB were failing to upload because Vercel caps serverless function request bodies at 4.5MB and we were base64-encoding the file through a tRPC mutation. This PR routes uploads directly to Supabase storage via a signed URL, bypassing the Vercel proxy entirely.

- Avoid Vercel 4.5MB body cap by uploading directly to Supabase storage with a signed URL
- Server-side path prefix check prevents linking arbitrary storage objects
- Tests updated to use two-step flow + new negative test for path spoofing

#1188 applies the same pattern to the rest of the upload endpoints. It's split out to keep this PR small and focused on the proposal-attachment fix — the follow-up touches several more endpoints and is easier to review on its own.


apps/app/src/lib/i18n/dictionaries/en.json
```diff
@@ -908,6 +908,7 @@
   "File too large. Maximum size: {size}MB": "File too large. Maximum size: {size}MB",
   "File too large: {name}": "File too large: {name}",
   "Unsupported file type: {name}": "Unsupported file type: {name}",
+  "Upload failed": "Upload failed",
```
scazan, line 947: We need the other translations

services/api/src/routers/decision/uploadProposalAttachment.ts
```diff
@@ -42,96 +115,44 @@ export const uploadProposalAttachment = router({
...
-        throw new CommonError('Storage configuration missing');
-      }
-
-      const supabase = createServerClient(supabaseUrl, supabaseServiceRole, {
-        cookieOptions: {},
-        cookies: {
-          getAll: async () => [],
-          setAll: async () => {},
-        },
+      const profileId = await getCurrentProfileId(user.id);
```
scazan: We can do the assert and this get in parallel to make it a little faster. We only end up with an extra fetch when access is denied

services/api/src/utils/storage.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+}
+
+/**
+ * Asserts the caller-supplied path matches the expected prefix and contains no
+ * traversal segments. Throws UnauthorizedError on mismatch — a client trying
+ * to claim a path under another caller's namespace is an authz failure, not a
+ * server bug.
+ */
+export function validateStoragePath({
```
scazan: maybe an assertValidStoragePath()

services/api/src/routers/decision/uploadProposalAttachment.ts
```diff
@@ -42,96 +115,44 @@ export const uploadProposalAttachment = router({
...
-      const bucket = 'assets';
-      const filePath = `profile/${profileId}/proposals/${Date.now()}_${sanitizedFileName}`;
-
-      const { error: uploadError, data } = await supabase.storage
-        .from(bucket)
-        .upload(filePath, buffer, {
-          contentType: mimeType,
-          upsert: false,
+      try {
+        const storageObject = await findUploadedStorageObject({
```
scazan: maybe should be getUploadedStorageObject for consistency

services/api/src/utils/index.ts
```diff
@@ -87,7 +87,7 @@ export function sanitizeS3Filename(filename: string) {
   return sanitizeForS3(sanitized, '-');
```
scazan: Any reason not to skip the replace and just stick with sanitizeForS3?

services/api/src/utils/storage.ts
```diff
@@ -0,0 +1,176 @@
...
+      );
+
+    if (found) {
+      row = found;
+      break;
+    }
+
+    if (attempt < RETRIES - 1) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, BACKOFF_MS * (attempt + 1)),
```
scazan: Let's pull out a backoff util as we will use it all over the place (especially with AI stuff)

scazan (commented):
Adding a few comments before finishing review

services/api/src/utils/index.ts
nourmalaeb: this was pre-existing and my change was a small tweak, but i agree we can rely on `sanitizeForS3` here

nourmalaeb:
closed in favor of #1420

PR #1187, opened 2026-05-18
feat(analytics): add 8 PostHog events for decision flows and invites

scazan:
## Summary

Wires PostHog tracking for 8 events spanning the decision lifecycle and invite flows:

- `user_voted` — vote submission
- `user_reviewed_proposal` — single proposal review submitted
- `user_finished_review` — reviewer completes all assignments for a process
- `admin_set_process` — decision process published (DRAFT → PUBLISHED)
- `admin_invited_participants` — admin invited people to a profile/process
- `admin_gave_roles` — admin assigned decision permissions to a role
- `admin_set_rubric` — admin saved a review rubric
- `user_invited` — user invited others to the platform/organization

Asana: https://app.asana.com/0/1213315744811204/1214898054846008

### DIFF
--- a/packages/analytics/src/utils.ts
+++ b/packages/analytics/src/utils.ts
@@ -402,6 +402,114 @@ export async function trackUserVoted(
 }
 
 /**
+ * Track when a reviewer submits a review for one proposal
+ */
+export async function trackProposalReviewed(
+  userId: string,
+  processId: string,
+  proposalId: string,
+  additionalProps?: Record<string, any>,
+): Promise<void> {
+  await trackEventWithContext(
+    userId,
+    'user_reviewed_proposal',
+    getDecisionCommonProperties(processId, proposalId, additionalProps),
+  );
+}
+
+/**
+ * Track when a reviewer finishes their entire review assignment list for a process
+ */
+export async function trackReviewListFinished(
+  userId: string,
+  processId: string,
+  additionalProps?: Record<string, any>,
+): Promise<void> {
+  await trackEventWithContext(
+    userId,
+    'user_finished_review',
+    getDecisionCommonProperties(processId, undefined, additionalProps),
+  );
+}
+
+/**
+ * Track when an admin publishes a decision process (DRAFT -> PUBLISHED)
+ */
+export async function trackAdminSetProcess(
+  userId: string,
+  processId: string,
+  additionalProps?: Record<string, any>,
+): Promise<void> {
+  await trackEventWithContext(
+    userId,
+    'admin_set_process',
+    getDecisionCommonProperties(processId, undefined, additionalProps),
+  );
+}
+
+/**
+ * Track when an admin invites participants to a profile (e.g. a decision process)
+ */
+export async function trackAdminInvitedParticipants(
+  userId: string,
+  profileId: string,
+  invitationCount: number,
+  additionalProps?: Record<string, any>,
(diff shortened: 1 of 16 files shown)


packages/analytics/src/utils.ts
```diff
@@ -382,22 +382,129 @@ export async function trackProposalFollowed(
 }
 
 /**
- * Track when a user votes on a proposal
+ * Track when a user submits a vote.
+ *
+ * Voting is private: we only record that the user voted in a given process.
+ * Do NOT pass the proposalId(s), selection counts, schema details, or any
+ * other field that could expose what they voted for. The userId + processId
+ * pair is the maximum we are willing to send.
```
valentin0h: it's not adding value given the input params
```suggestion
```

packages/analytics/src/utils.ts
```diff
@@ -382,22 +382,129 @@ export async function trackProposalFollowed(
...
+ * Track when a user invites others to the platform / organization
+ */
+export async function trackUserInvited(
+  userId: string,
+  inviteCount: number,
+  additionalProps?: Record<string, any>,
+): Promise<void> {
+  await trackEventWithContext(userId, 'user_invited', {
+    invite_count: inviteCount,
+    timestamp: new Date().toISOString(),
```
valentin0h: Is this critical to set here rather than relying on the event timestamp?

packages/common/src/services/decision/decisionRoles.ts
```diff
@@ -279,5 +281,11 @@ export async function updateDecisionRoles({
 
   await invalidateProfileUserCacheForRole(roleId);
 
+  waitUntil(
+    trackAdminGaveRoles(user.id, roleId, {
+      decision_permissions: decisionPermissions,
+    }).catch((err) => console.error('Failed to track admin_gave_roles', err)),
```
valentin0h: Could add `.catch` to `trackEventWithContext` so we don't have to write it for each event.

packages/common/src/services/decision/submitReview.ts
```diff
@@ -79,11 +81,57 @@ export async function submitReview({
...
+          eq(
+            proposalReviewAssignments.reviewerProfileId,
+            context.assignment.reviewerProfileId,
+          ),
+          ne(
+            proposalReviewAssignments.status,
+            ProposalReviewAssignmentStatus.COMPLETED,
+          ),
+        ),
+      );
```
valentin0h: Does this need to be inside the transaction? It looks like it's an analytics read. I'd put inside the `waitUntil`

packages/common/src/services/decision/voting.ts
```diff
@@ -308,6 +310,19 @@ export const submitVote = async ({
...
+        trackUserVoted(
+          authUserId,
+          result.voteSubmission.processInstanceId,
+        ).catch((err) => console.error('Failed to track user_voted', err)),
+      );
+    } catch (err) {
+      // waitUntil can throw synchronously off-Vercel (no request context).
+      // Swallow here so analytics never causes "Failed to submit vote".
+      console.error('Failed to schedule user_voted tracking', err);
+    }
```
valentin0h, line 321: seems like we have both try/catch and .catch

valentin0h (approved):
Interesting that we're naming the events after the user role. That's usually a property on the user object.

PR #1189, opened 2026-05-19
fix(a11y): allow user zoom to satisfy WCAG 1.4.4

nourmalaeb:
## Summary
- Drop `maximumScale`, `minimumScale`, `userScalable`, and non-standard `height` from the root `viewport` export.
- Resolves axe `meta-viewport` violation (WCAG 1.4.4 Resize Text) across all 16 routes in the a11y baseline.
- Low-vision users can now pinch- and browser-zoom; `initialScale: 1` still prevents iOS input auto-zoom (inputs already render at ≥16px).
- Also adds a name to the link wrapping the Common logo in the site header.

### DIFF
--- a/apps/app/src/app/[locale]/start/layout.tsx
+++ b/apps/app/src/app/[locale]/start/layout.tsx
@@ -1,9 +1,9 @@
-import { Link, useTranslations } from '@/lib/i18n/routing';
+import { Link } from '@/lib/i18n/routing';
 
 import { CommonLogo } from '@/components/CommonLogo';
+import { TranslatedText } from '@/components/TranslatedText';
 
 const StartLayout = ({ children }: { children: React.ReactNode }) => {
-  const t = useTranslations();
   return (
     <div className="relative flex h-svh w-full flex-col items-center justify-center font-sans">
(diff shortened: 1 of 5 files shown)


apps/app/src/app/[locale]/start/layout.tsx
```diff
@@ -12,6 +13,7 @@ const StartLayout = ({ children }: { children: React.ReactNode }) => {
             <Link
               href="/"
               className="flex items-center gap-2 hover:no-underline"
+              aria-label={t('Home')}
```
Comment (scazan) [inline apps/app/src/app/[locale]/start/layout.tsx]: We should use the server version of translations here. `<TranslatedText />` as this isn't a client component.

apps/app/src/components/layout/split/FullScreenSplitMain.tsx
```diff
@@ -1,4 +1,4 @@
-import { Link } from '@/lib/i18n/routing';
+import { Link, useTranslations } from '@/lib/i18n/routing';
```
scazan: Same thing in the rest of the file. If it's not a 'use client' we should use the server translations.

scazan (approved):
Looks good, but we should update the translations to make sure they're the correct version for server/client.

PR #1190, opened 2026-05-19
experiment: scroll process builder via nuqs scroll option

nourmalaeb:
## Summary
- Tries `scroll: true` on nuqs `useQueryState('section', ...)` instead of a manual scroll-reset effect
- Compare against #1154

## Expected outcome
- `scroll: true` delegates to Next.js router which calls `window.scrollTo(0, 0)`
- Process builder scroll container is `<main>` inside `overflow-y-hidden` parent — `window` isn't the scroller
- Likely no observable scroll reset, but worth confirming locally


nourmalaeb:
Closing because would require refactoring the layout

PR #1192, opened 2026-05-20
fix(profile): prevent removing admin access from a profile owner

scazan:
## Summary
- Server: `updateProfileUserRoles` now rejects role updates that would strip `profile.ADMIN` from a profileUser flagged `isOwner=true`, with a `ValidationError("Cannot remove admin access from the owner of a profile")`.
- UI: the role-select dropdown on `ProfileUsersAccessTable` is disabled for owners, matching the server contract.
- Test: `updateUserRoles.test.ts` now covers the owner-demotion case (existing tests still pass).

## Asana
https://app.asana.com/0/1213315744811204/1214897805510041

### DIFF
--- a/packages/common/src/services/profile/updateProfileUserRole.ts
+++ b/packages/common/src/services/profile/updateProfileUserRole.ts
@@ -2,14 +2,15 @@ import { invalidate } from '@op/cache';
 import { and, db, eq, inArray } from '@op/db/client';
 import { profileUserToAccessRoles } from '@op/db/schema';
 import type { User } from '@op/supabase/lib';
-import { assertAccess, permission } from 'access-zones';
+import { assertAccess, checkPermission, permission } from 'access-zones';
 
 import {
   CommonError,
   NotFoundError,
   UnauthorizedError,
+  ValidationError,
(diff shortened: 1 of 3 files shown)


packages/common/src/services/profile/updateProfileUserRole.ts
```diff
@@ -68,6 +69,31 @@ export const updateProfileUserRoles = async ({
 
   assertAccess({ profile: permission.ADMIN }, currentProfileUser.roles ?? []);
 
+  if (targetProfileUser.isOwner) {
+    // Profile owners must always retain admin access on their own profile —
+    // the UI lets an owner switch themselves to a non-admin role and locks
+    // them out, so we enforce it server-side. Load the desired roles with
+    // their zone permissions and verify the union still grants profile ADMIN.
+    const desiredRolesWithPermissions = await db.query.accessRoles.findMany({
```
valentin0h:
```suggestion
    const requestedRolesWithPermissions = await db.query.accessRoles.findMany({
```

valentin0h (approved):

PR #1193, opened 2026-05-20
feat(decisions): add admin-only listVoters endpoint

scazan:
## Summary

- New `decision.listVoters` tRPC endpoint returning the profile slugs/names/avatars of users who have submitted a vote in a process.
- Admin-only: enforces `decisions: ADMIN` on the instance profile so participation lists don't leak.
- Uses `SELECT DISTINCT` joined to `decisionsVoteSubmissions` (the unique `(processInstanceId, submittedByProfileId)` constraint already prevents duplicates).
- Output schema mirrors `listProposalSubmitters` for reuse in face-pile UIs.

Asana task: https://app.asana.com/0/1210939178526797/1211900775493988

## Test plan

- [x] `pnpm typecheck` passes
- [x] New integration test `listVoters.test.ts` covers:
  - Returns voters who submitted a vote
  - Excludes members who never voted
  - Returns multiple distinct voters
  - Rejects non-admin callers (AccessControlException)
- [ ] Reviewer sanity-checks the admin-only auth

### DIFF
--- a/packages/common/src/services/decision/listVoters.ts
+++ b/packages/common/src/services/decision/listVoters.ts
@@ -1,0 +1,73 @@
+import { db, eq } from '@op/db/client';
+import {
+  decisionsVoteSubmissions,
+  objectsInStorage,
+  profiles,
+} from '@op/db/schema';
+import type { User } from '@op/supabase/lib';
+import { assertAccess, permission } from 'access-zones';
+
+import { CommonError, UnauthorizedError } from '../../utils';
+import { getProfileAccessUser } from '../access';
+
+export interface ListVotersInput {
+  processInstanceId: string;
+}
+
+/**
+ * Returns the profiles that have actually submitted a vote in the process.
+ * Admin-only: process admins see the participation list, no one else does.
+ */
+export const listVoters = async ({
+  input,
+  user,
+}: {
+  input: ListVotersInput;
+  user: User;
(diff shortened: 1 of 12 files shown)


packages/common/src/services/decision/listVoters.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Returns the profiles that have actually submitted a vote in the process.
+ * Admin-only: process admins see the participation list, no one else does.
+ */
+export const listVoters = async ({
+  input,
+  user,
+}: {
+  input: ListVotersInput;
```
valentin0h:
```suggestion
  processInstanceId: string;
```

packages/common/src/services/decision/schemas/voters.ts
```diff
@@ -0,0 +1,20 @@
...
+/** Minimal voter profile shape returned to process admins. */
+export const voterSchema = z.object({
+  slug: z.string(),
+  name: z.string().nullable(),
+  avatarImage: z
+    .object({
+      name: z.string(),
+    })
+    .nullable(),
+});
```
valentin0h: we could introduce a minimal profile schema to make use of here and for `proposalSubmitterSchema`.

valentin0h (approved):

PR #1195, opened 2026-05-20
fix(decisions): gate phase-transition email on selectionsAreConfirmed

valentin0h:
Replaces the temporary last-phase suppression from #1183 with a proper gate: the participant email fires only once the inbound transition's selections are confirmed.


services/workflows/src/functions/notifications/sendPhaseTransitionNotification.ts
```diff
@@ -52,6 +56,29 @@ export const sendPhaseTransitionNotification = inngest.createFunction(
...
+    const manualSelectionStatus = await step.run(
+      'check-manual-selection',
+      async () =>
+        resolveManualSelectionStatus({
+          instance: {
+            id: processInstanceId,
+            instanceData: processData.instanceData,
+            currentStateId: processData.currentStateId,
+          },
+        }),
```
valentin0h, line 76: `resolveManualSelectionStatus` should be scoped to the input phase. I can do that in a separate PR.

scazan: Nice. We should also rename this to resolveSelectionStatus as it can be used to resolve whether selections are confirmed or not (manually or automatically)

services/events/src/types.ts
```diff
@@ -57,6 +57,14 @@ export const Events = {
       toPhaseId: z.string().min(1),
     }),
   },
+  selectionsConfirmed: {
+    name: 'decision/selections-confirmed' as const,
```
scazan: This can happen in the transition as well so maybe easier to have it as an event `decision/manual-selections-confirmed`

scazan (approved):

PR #1198, opened 2026-05-20
feat(db): add process survey response + submitter tables

scazan:
## Summary

Adds two tables backing the end-of-process NPS survey, split intentionally so response content cannot be linked back to a submitter via schema-level joins:

- **`decision_process_survey_responses`** — anonymous response payload. Only an FK to `decision_process_instances` and a JSONB `internal_data` column. No FK to the submitter.
- **`decision_process_survey_submitters`** — tracks which profiles have submitted for a given process instance, with a unique `(process_instance_id, submitted_by_profile_id)` constraint to gate duplicate submissions. Holds no response content.


nourmalaeb (approved):
Wondering if we need timestamps for submitters? If the submitter and submission are created at a similar time then they can be linked.

PR #1199, opened 2026-05-21
fix(decisions): hide Start a proposal for org members without profile role

scazan:
## Summary
- Stop `resolveInstanceAccess` from leaking org-level action capabilities into `instance.access`. The seeded org MEMBER role grants `SUBMIT_PROPOSALS` and `VOTE`, so any org member viewing a decision instance in their org was getting `access.submitProposals = true` — even when the server `createProposal` mutation (and every other action) requires a profile-level grant. The "Start a proposal" button on the decision listing page rendered for users who could not actually submit; clicking it surfaced a "Failed to create proposal" toast.
- Org-level fallback now strips action capabilities (`admin`, `inviteMembers`, `review`, `submitProposals`, `vote`). Read-only viewing bits remain so org members can still load the page.
- Adds an e2e regression: a member of the decision owner organization, with no profile-level role on the instance, should not see the "Start a proposal" button.

## Test plan
- [ ] `pnpm typecheck` — passes
- [ ] E2E `decision-role-capabilities.spec.ts` — new test:
  - `start a proposal button is hidden for org members without profile-level role`
- [ ] Manual: as a non-admin member of the decision owner organization without a profile-level role, open the decision listing page and confirm no "Start a proposal" button next to "Learn more"; an admin with profile-level access still sees it.

Asana task: https://app.asana.com/0/1213315744811204/1214897514250826

### DIFF
--- a/apps/app/src/components/decisions/ProposalEditorLayout.tsx
+++ b/apps/app/src/components/decisions/ProposalEditorLayout.tsx
@@ -38,6 +38,7 @@ interface ProposalEditorLayoutProps {
   access?: {
     admin: boolean;
     inviteMembers: boolean;
+    submitProposals: boolean;
   };
   /** Active revision request when the editor is in revision mode */
   revisionRequest?: ProposalReviewRequest | null;
@@ -67,6 +68,10 @@ export function ProposalEditorLayout({
 
   const canShare = access?.admin || access?.inviteMembers;
   const isRevisionMode = Boolean(revisionRequest);
+  const isUpdateMode = isEditMode && !isDraft;
+  // Update mode goes through updateProposal (UPDATE perm), not submitProposal,
+  // so it's not gated on SUBMIT_PROPOSALS.
+  const showSubmitButton = isUpdateMode || (access?.submitProposals ?? false);
 
   return (
     <div className="flex min-w-0 flex-1 flex-col bg-white">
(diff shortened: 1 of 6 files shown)


scazan:
Closing this PR after deeper investigation.

**What I changed and why I'm reverting it:**

The original fix gated `submitProposals` (and other action capabilities) when access was resolved through the org-level fallback in `resolveInstanceAccess` (`packages/common/src/services/decision/getInstance.ts`). An integration test confirmed the org-fallback does inflate action capabilities for org members without a profile-level role — that part is a real server-side issue.

**Why it's the wrong fix for the reported bug:**

After re-examining the screenshot attached to the Asana task:

1. The user (Juanfe) already has 4 existing proposals visible on the page (3 submitted + 1 draft) with Edit/Delete affordances — so they clearly already have `submitProposals` permission.
2. The error toast in the screenshot reads "Failed to create proposal" / "Failed to create proposal" (title + body identical). That body text matches the generic catch-all at `packages/common/src/services/decision/createProposal.ts:290` (`throw new CommonError('Failed to create proposal')`), NOT any of the permission-rejection paths (which throw `UnauthorizedError('Not authorized')` and surface that message).
3. The decision profile shown is the decision's own profile, not the user's org profile — so the org-fallback path likely isn't even being exercised in this scenario.

**Conclusion:** The reported bug is a server error during `createProposal` being swallowed by a generic catch and surfaced as "Failed to create proposal" — it is not a permission-visibility issue. The original task framing ("Submit proposal button should not appear if you don't have 'submit proposal' permission") does not match what the screenshot actually shows.

Reverting and going back to the task owner for clarification before doing more work. The org-fallback inflation issue is real and worth tracking, but it's a separate ticket.

PR #1208, opened 2026-05-22
feat(decisions): show allocated vs requested budget on proposal card and detail view

valentin0h:
Surfaces the allocated amount alongside the original requested budget once a decision's selections are confirmed, so viewers can see the gap between ask and award.

## Demo

CleanShot 2026-05-22 at 12 17 04@2x
CleanShot 2026-05-22 at 12 18 08@2x

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx
@@ -32,7 +32,7 @@ function ProposalViewPageContent({
       proposal={proposal}
       canSeeRevisions={false}
       backHref={backHref}
-      selectionsAreConfirmed={false}
+      selection={null}
     />
   );
 }
--- a/apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx
@@ -32,6 +32,7 @@ function ProposalViewPageContent({
       proposal={proposal}
       canSeeRevisions={false}
       backHref={backHref}
+      selectionsAreConfirmed={false}
     />
   );
 }
--- a/apps/app/src/components/decisions/ProposalView.tsx
+++ b/apps/app/src/components/decisions/ProposalView.tsx
@@ -27,10 +27,12 @@ export function ProposalView({
   proposal: initialProposal,
   canSeeRevisions,
   backHref,
+  selectionsAreConfirmed,
 }: {
   proposal: Proposal;
   canSeeRevisions: boolean;
   backHref: string;
+  selectionsAreConfirmed: boolean;
 }) {
   const t = useTranslations();
   const locale = useLocale();
@@ -41,6 +43,15 @@ export function ProposalView({
 
   // Safety check - fallback to initial data if query returns undefined
   const currentProposal = proposal || initialProposal;
+
+  // Only fetch allocation once the instance has confirmed its selections —
+  // before that, the "latest" result row isn't unambiguously canonical.
+  const { data: selection } =
+    trpc.decision.getLatestSelectionForProposal.useQuery(
+      { proposalId: currentProposal.id },
+      { enabled: selectionsAreConfirmed },
+    );
+  const allocated = selection?.allocated ?? null;
 
   // Use relationship mutations hook for like/follow functionality
   const {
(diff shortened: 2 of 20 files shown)


services/api/src/routers/decision/proposals/getLatestSelection.ts
```diff
@@ -0,0 +1,24 @@
...
+    )
+    .output(proposalSelectionSchema.nullable())
+    .query(async ({ ctx, input }) => {
+      const { user } = ctx;
+      const { proposalId } = input;
+
+      return getLatestSelectionForProposal({
+        proposalId,
+        user,
+      });
```
valentin0h, line 22: We can add caching as soon as we're done with SEI.

apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx
```diff
@@ -32,6 +32,7 @@ function ProposalViewPageContent({
       proposal={proposal}
       canSeeRevisions={false}
       backHref={backHref}
+      selection={null}
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/profile/[slug]/decisions/[id]/proposal/[profileId]/page.tsx:35]: This one is optional. Maybe nicer to just omit it?

apps/app/src/components/decisions/ProposalView.tsx
```diff
@@ -27,10 +28,12 @@ export function ProposalView({
   proposal: initialProposal,
   canSeeRevisions,
   backHref,
+  selection,
 }: {
   proposal: Proposal;
   canSeeRevisions: boolean;
   backHref: string;
+  selection: ProposalSelection | null;
```
scazan, line 36: Related to my earlier comment. this can also be optional since we pass through to optional props.

packages/common/src/services/decision/getLatestSelectionForProposal.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  if (!proposal) {
+    throw new NotFoundError('Proposal', proposalId);
+  }
+
+  await assertInstanceProfileAccess({
+    user: { id: user.id },
+    instance: proposal.processInstance,
+    profilePermissions: { decisions: permission.READ },
```
scazan, line 40: It's kind of nice that we can use the read permission here, it potentially block admins from viewing proposals which is a cool feature down the road. But just to point out that the permission.ADMIN gives a separate flag and usually we also include READ on admin permissions. I like this being permission.READ for that reason.

packages/common/src/services/decision/getLatestSelectionForProposal.ts
```diff
@@ -0,0 +1,74 @@
...
+    orgFallbackPermissions: [
+      { decisions: permission.READ },
+      { decisions: permission.ADMIN },
+    ],
+  });
+
+  const latestResult = await db.query.decisionProcessResults.findFirst({
```
scazan, line 47: You could do a parallel call here with the above assertion.

services/api/src/routers/decision/proposals/getLatestSelection.ts
```diff
@@ -0,0 +1,24 @@
+import { getLatestSelectionForProposal } from '@op/common';
+import { proposalSelectionSchema } from '@op/common/client';
+import { z } from 'zod';
+
+import { commonAuthedProcedure, router } from '../../../trpcFactory';
+
+export const getLatestSelectionForProposalRouter = router({
```
scazan, line 7: Should we rename the file to this?

packages/common/src/services/decision/getLatestSelectionForProposal.ts
```diff
@@ -0,0 +1,74 @@
...
+    where: { processInstanceId: proposal.processInstanceId },
+    orderBy: (table, { desc }) => [desc(table.executedAt)],
+    columns: { id: true, success: true },
+  });
+
+  if (!latestResult?.success) {
+    return null;
+  }
+
+  const selection = await db.query.decisionProcessResultSelections.findFirst({
```
scazan, line 57: You could also just do a join right above I'd imagine. Get 'em both in one go.

scazan (approved):

packages/common/src/services/decision/schemas/selection.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * A proposal's selection record from the latest successful result run on its
+ * process instance. `allocated` is stored as numeric (string) in the DB to
+ * preserve precision.
+ */
+export const proposalSelectionSchema = z.object({
+  proposalId: z.uuid(),
+  allocated: z.string().nullable(),
+  selectionRank: z.number().nullable(),
```
scazan, line 11: Not sure we need to expose selectionRank currently. Perhaps let's omit it.

PR #1209, opened 2026-05-22
test(decisions): cover allocated vs requested budget paths

valentin0h:
Adds API + e2e coverage for the latest-selection endpoint and the allocated/requested budget UI added in #1208.

### DIFF
--- a/services/api/src/routers/decision/proposals/getLatestSelection.test.ts
+++ b/services/api/src/routers/decision/proposals/getLatestSelection.test.ts
@@ -568,5 +568,4 @@ describe.concurrent('getLatestSelectionForProposal', () => {
       selectionRank: 1,
     });
   });
-
 });
--- a/services/api/src/routers/decision/proposals/getLatestSelection.test.ts
+++ b/services/api/src/routers/decision/proposals/getLatestSelection.test.ts
@@ -1,0 +1,596 @@
+import { db } from '@op/db/client';
+import {
+  decisionProcessResultSelections,
+  decisionProcessResults,
+} from '@op/db/schema';
+import { randomUUID } from 'node:crypto';
+import { describe, expect, it } from 'vitest';
+
+import { appRouter } from '../..';
+import { TestDecisionsDataManager } from '../../../test/helpers/TestDecisionsDataManager';
+import {
+  createIsolatedSession,
+  createTestContextWithSession,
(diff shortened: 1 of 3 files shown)


services/api/src/routers/decision/proposals/getLatestSelection.test.ts
```diff
@@ -0,0 +1,571 @@
...
+    const caller = await createAuthenticatedCaller(setup.userEmail);
+
+    const result = await caller.decision.getLatestSelectionForProposal({
+      proposalId: skipped.id,
+    });
+
+    expect(result).toBeNull();
+  });
+
+  it('returns allocated + selectionRank from the latest successful result', async ({
```
scazan, line 160: Fine to have this but I'm not sure we'll use selectionRank on this table in the end.

services/api/src/routers/decision/proposals/getLatestSelection.test.ts
```diff
@@ -0,0 +1,571 @@
...
+    const caller = await createAuthenticatedCaller(setup.userEmail);
+
+    await expect(
+      caller.decision.getLatestSelectionForProposal({
+        proposalId: randomUUID(),
+      }),
+    ).rejects.toMatchObject({ cause: { name: 'NotFoundError' } });
+  });
+
+  it('throws UnauthorizedError when the caller is in a different organization', async ({
```
scazan, line 426:
```suggestion
  it('throws UnauthorizedError when the caller is in a different organization on legacy instances', async ({
```

services/api/src/routers/decision/proposals/getLatestSelection.test.ts
```diff
@@ -0,0 +1,571 @@
...
+    );
+
+    await expect(
+      outsiderCaller.decision.getLatestSelectionForProposal({
+        proposalId: proposal.id,
+      }),
+    ).rejects.toMatchObject({ cause: { name: 'UnauthorizedError' } });
+  });
+
+  it('allows an org member without profile access via the org-level fallback', async ({
```
scazan, line 460: For proposals, we probably want to not allow this in the end except when it's a "legacy process". Just noting that this test should eventually be correctly failing. We haven't quite solved it yet, though elsewhere.

scazan (approved):

PR #1216, opened 2026-05-25
feat(ui): Add ButtonGroup Component

scazan:
## Summary

- Adds `ButtonGroup` into `@op/ui` (`packages/ui/src/components/ui/button-group.tsx`) from intent-ui with some custom styling
- Adds `./ButtonGroup` to the `@op/ui` package exports.
- Adds a Storybook story with three examples: basic three-button group, two-option toggle (selected state styled with brand teal tokens), and a vertical orientation.

The component is a thin wrapper around its children: it sets `role="group"` and strips the rounded edges + border between adjacent items so a row (or column) of `<Button>`s reads as a single grouped control. Used in the resources panel for the Link/Document tabs but also to be used in filter toggles so needs to be agnostic to tabs vs radiogroups vs toggles.

Storybook: https://apps-workshop-qpz30ksyt-oneproject.vercel.app/?path=/docs/buttongroup--docs

Screenshot 2026-05-25 at 17 39 20
Screenshot 2026-05-25 at 17 39 21

### DIFF
--- a/packages/ui/src/components/ButtonGroup.tsx
+++ b/packages/ui/src/components/ButtonGroup.tsx
@@ -1,0 +1,49 @@
+import { twMerge } from 'tailwind-merge';
+import { tv } from 'tailwind-variants';
+import type { VariantProps } from 'tailwind-variants';
+
+const buttonGroupStyles = tv({
+  base: "flex w-fit items-stretch *:focus-visible:relative *:focus-visible:z-10 has-[>[data-slot=button-group]]:gap-2 [&>*]:shadow-none [&>[aria-pressed=false]]:bg-neutral-offWhite [&>[aria-pressed=false]]:text-neutral-charcoal [&>[aria-pressed=false]_svg]:text-neutral-gray4 [&>[aria-pressed=false]:hover]:bg-neutral-offWhite [&>[aria-pressed=true]]:bg-primary-tealWhite [&>[aria-pressed=true]]:text-primary-teal [&>[aria-pressed=true]:hover]:bg-primary-tealWhite has-[select[aria-hidden=true]:last-child]:[&>[data-slot=select-trigger]:last-of-type]:rounded-e-md [&>[data-slot=select-trigger]:not([class*='w-'])]:w-fit [&>input]:flex-1",
+  variants: {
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/ButtonGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <div
+      className={twMerge(
+        "flex items-center gap-2 rounded-md border bg-muted px-4 text-sm font-medium shadow-xs *:data-[slot=icon]:pointer-events-none [&_[data-slot=icon]:not([class*='size-'])]:size-5 sm:[&_[data-slot=icon]:not([class*='size-'])]:size-4",
+        className,
+      )}
+      {...props}
+    />
+  );
+}
```
valentin0h, line 49: Will this be used?

packages/ui/stories/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+        color="secondary"
+        size="small"
+        aria-pressed={value === 'bottom'}
+        onPress={() => setValue('bottom')}
+      >
+        Bottom
+      </Button>
+    </ButtonGroup>
+  );
+};
```
valentin0h, line 110: It doesn't look like this is the UI we want at first glance https://apps-workshop-qpz30ksyt-oneproject.vercel.app/?path=/story/buttongroup--vertical 

If not and not used for now let's just not add at this time.

valentin0h (approved):

packages/ui/src/components/ButtonGroup.tsx
scazan: It makes sense to keep it. it's the Intent UI component that we are importing here not custom code.

packages/ui/stories/ButtonGroup.stories.tsx
scazan: This is just the intent UI variant import (not custom code) so we should have it here. But here I'm just not styling it in a way that we will use it. So we aren't adding the styles for it yet. it'll be the same with shadcn.

PR #1220, opened 2026-05-27
perf(access): request-scoped cache for access-control fetches

scazan:
## Summary
- Wraps `getProfileAccessUser`, `getOrgAccessUser`, and `getUserSession` (`packages/common/src/services/access/index.ts`) in an `AsyncLocalStorage`-backed memo. The same `(profileId, authUserId)` / `(orgId, authUserId)` / `authUserId` lookups in a single tRPC request now hit the DB once instead of N times.
- `assertAccess` itself stays in the `access-zones` library — it is already a synchronous bitfield check. Only the roles fetch that precedes it is cached.
- The cache scope is entered by a new `withRequestCache` middleware on `commonProcedure` in `services/api/src/trpcFactory.ts`, so every tRPC procedure runs inside its own store. Outside a scope (scripts, tests), `memoRequest` calls the fetcher every time — no behavior change.
- Existing `@op/cache` invalidations (`updateProfileUserRole`, `removeProfileUser`, `deleteOrganization`, `updateOrganizationUser`) now also call `invalidateRequest(...)` so within-request reads after a role change see fresh data.

## Cache-poisoning safety
`packages/common/src/services/access/requestCache.test.ts` covers:
- isolation between separate `runWithRequestCache` scopes (no cross-request leak)
- isolation between concurrent requests using the same key (each scope dedupes independently)
- rejected fetchers drop the cache entry so the next caller retries (no poisoned rejection cached)
- a rejection on one key never leaks to other keys
- prefix-based `invalidateRequest` only evicts matching keys
- out-of-scope behavior: `memoRequest` always calls the fetcher; `invalidateRequest` is a no-op

### DIFF
--- a/packages/common/src/services/access/requestCache.ts
+++ b/packages/common/src/services/access/requestCache.ts
@@ -1,40 +1,76 @@
 import { AsyncLocalStorage } from 'node:async_hooks';
 
-type Store = Map<string, Promise<unknown>>;
+type Sub = Map<string, Promise<unknown>>;
+type Store = WeakMap<object, Sub>;
 
 const als = new AsyncLocalStorage<Store>();
 
 export const runWithRequestCache = <T>(fn: () => T): T =>
-  als.run(new Map(), fn);
+  als.run(new WeakMap(), fn);
 
-const getStore = (): Store | undefined => als.getStore();
-
-export async function memoRequest<T>(
-  key: string,
-  fetcher: () => Promise<T>,
-): Promise<T> {
-  const store = getStore();
-  if (!store) {
-    return fetcher();
+function stableStringify(value: unknown): string {
+  if (value === null || typeof value !== 'object') {
+    return JSON.stringify(value) ?? 'undefined';
   }
-  const hit = store.get(key);
-  if (hit) {
-    return hit as Promise<T>;
+  if (Array.isArray(value)) {
+    return `[${value.map(stableStringify).join(',')}]`;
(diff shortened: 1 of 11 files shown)


packages/common/src/services/access/requestCache.ts
```diff
@@ -0,0 +1,57 @@
...
+export const runWithRequestCache = <T>(fn: () => T): T =>
+  als.run(new Map(), fn);
+
+export interface MemoizedFn<TArgs extends unknown[], TReturn> {
+  (...args: TArgs): Promise<TReturn>;
+  invalidate(...args: TArgs): void;
+  invalidateAll(): void;
+}
+
+export function memoize<TArgs extends unknown[], TReturn>(
```
scazan, line 15: opting for a simple custom function over something like dataloader here for a simpler interface

services/api/src/routers/platform/admin/addUsersToOrganization.ts
```diff
@@ -116,6 +121,12 @@ export const addUsersToOrganizationRouter = router({
           }),
         ),
       ]);
+      for (const addedUser of joinResults) {
+        getOrgAccessUser.invalidate({
```
scazan: invalidations like this are only necessary when we have request-scope stale data. Usually after a DB mutation where we need to re-access data.

nourmalaeb (approved):

PR #1225, opened 2026-05-28
test(api): add gating coverage across routers

valentin0h:
Codify closed-network access expectations so future auth changes fail fast when they unintentionally widen API access.

General procedure example:

```ts
describeGating("account.login", {
  noJwt: async ({ callers }) => { /* no session */ },
  anonJwt: async ({ callers }) => { /* anonymous Supabase user */ },
  userJwt: async ({ callers }) => { /* authenticated, out-of-network user */ },
  networkJwt: async ({ callers }) => { /* authenticated in-network user */ },
});
```

Decision procedure example:

```ts
describeDecisionGating("createProposal", {
  noJwtNonPublic: async ({ callers }) => { /* rejects at auth gate */ },
  anonJwtNonPublic: async ({ callers }) => { /* rejects at auth gate */ },
  userJwtNonPublic: async ({ callers }) => { /* rejects at auth gate */ },
  networkJwtNonPublic: async ({ callers }) => { /* reaches decision authz */ },
});
```

Public decision cells like `noJwtPublic`, `anonJwtPublic`, `userJwtPublic`, and `networkJwtPublic` will follow when public-instance access lands.


packages/common/src/utils/error/index.ts
```diff
@@ -53,6 +53,18 @@ export class UnauthorizedError extends CommonError {
...
+ * Raised by protected API auth gates (`verifyAuthentication` and the
+ * `withAuthenticated*` middlewares) when a request fails before resolver-level
+ * authorization runs: no valid session, anonymous user, unconfirmed email,
+ * missing auth email, or non-allow-listed user.
+ *
+ * It intentionally maps like {@link UnauthorizedError} today, while letting
+ * tests and callers distinguish gate rejection from deeper permission checks
+ * that throw plain `UnauthorizedError`.
+ */
+export class AuthGateError extends UnauthorizedError {}
```
scazan:
```suggestion
export class UnauthenticatedError extends UnauthorizedError {}
```

scazan: AuthGate is a bit confusing since authorization gates exists and it's unclear if this is Authentication or Authorization. Is this more about an In-network gate?

I would also note here that we are extending what looks like an Authentication error type with an Authorization error type. Not an issue but it is getting mixed and blurry where it is used.

I suggest we setup an UnauthenticatedError with code 401.

services/api/src/test/helpers/gating/index.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Asserts a call is admitted by the authentication gate.
+ *
+ * The gate (`verifyAuthentication` / the `withAuthenticated*` middlewares)
+ * throws {@link AuthGateError} — and nothing else does. So the check is a
+ * single `instanceof` on the rejection's cause, with no message parsing: an
+ * `AuthGateError` means the gate blocked the caller, anything else (a
+ * resolve, input validation, not-found, or a deeper authorization
+ * `UnauthorizedError`) means the gate let them through.
+ */
+export const expectPassesAuthGate = async (promise: Promise<unknown>) => {
```
scazan: We should be explicit in our var naming here. expectPassesAuthGate should communicate that it is an  authentication pass vs an authorization pass.

services/api/src/test/helpers/gating/index.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Forgetting a key is a compile error.
+ */
+export type GatingCells = {
+  noJwt: GatingBody;
+  anonJwt: GatingBody;
+  userJwt: GatingBody;
+  networkJwt: GatingBody;
+};
+
+export const describeGating = (name: string, cells: GatingCells) => {
```
scazan:
```suggestion
export const describeNetworkGating = (name: string, cells: GatingCells) => {
```

I'm not sure about this naming but the suggestion is that we are more descriptive about what we are gating here.

services/api/src/test/helpers/gating/decision.ts
```diff
@@ -0,0 +1,54 @@
...
+        onTestFinished: (fn: () => void | Promise<void>) => void;
+      }) => {
+        await body({
+          task,
+          onTestFinished,
+          callers: createGatingCallers(onTestFinished),
+        });
+      };
+
+    it('no-JWT caller on non-public instance', wrap(cells.noJwtNonPublic));
```
scazan: This is meant to be read like a sentence and we do that in the rest of our tests. Wondering if we can keep that. At the moment: "It no-JWT caller on a non-public instance". Otherwise I'm not really sure if this should pass or fail when I have no JWT.

scazan (commented):
A few points:
- I think we should update the naming because it's easy to conflate the authentication vs authorization here. 
- We can also make use of standard HTTP codes for this (401 vs 403). 
- The descriptions in the test are a bit ambiguous as to what they are testing for describeGating. In some cases we are going to want to test that something passed an authentication vs authorization gate

services/api/src/test/helpers/gating/index.ts
```diff
@@ -0,0 +1,108 @@
...
+ * The gate (`verifyAuthentication` / the `withAuthenticated*` middlewares)
+ * throws {@link AccessTierError} — and nothing else does. So the check is a
+ * single `instanceof` on the rejection's cause, with no message parsing: an
+ * `AccessTierError` means the gate blocked the caller, anything else (a
+ * resolve, input validation, not-found, or a deeper resource-authorization
+ * `UnauthorizedError`) means the gate let them through.
+ *
+ * @see expectFailsTierGate for the inverse assertion.
+ */
+export const expectPassesTierGate = async (promise: Promise<unknown>) => {
```
scazan:
```suggestion
export const expectPassesAccessTierGate = async (promise: Promise<unknown>) => {
```

scazan (approved):
Approving! I still think this is a very strange way to use `it()`: ` it('no-JWT caller on non-public instance',` and I added one comment where the AccessTier naming was shortened again into just Tier. Keeping it consistent means it's easier to grep generally.

services/api/src/test/helpers/gating/index.ts
valentin0h: Done in a3eaab495 — renamed `expectPassesTierGate` → `expectPassesAccessTierGate` and, for symmetry, `expectFailsTierGate` → `expectFailsAccessTierGate`, matching the `describeAccessTierGating` vocabulary.

PR #1232, opened 2026-05-29
feat(profiles): handle nullable user email across app, api, and workflows

valentin0h:
Defensive null handling for profile/user email plus the encoder contract change (profileUserEncoder.email is now nullable). Self-typechecks against the current non-null schema; the schema flip lands in the stacked follow-up.

---
**Stack** (merge bottom-up):
1. #1232 — app/api/workflows nullable-email handling
2. #1233 — make email columns nullable
3. #1234 — anonymous-user signup support

### DIFF
--- a/packages/common/src/services/profile/inviteUsersToProfile.ts
+++ b/packages/common/src/services/profile/inviteUsersToProfile.ts
@@ -247,7 +247,11 @@ export const inviteUsersToProfile = async ({
   }
 
   const usersByEmail = new Map(
-    existingUsers.map((user) => [user.email.toLowerCase(), user]),
+    existingUsers
+      .filter(
+        (user): user is typeof user & { email: string } => user.email !== null,
+      )
+      .map((user) => [user.email.toLowerCase(), user]),
   );
 
   const existingProfileUserAuthIds = new Set(
--- a/packages/common/src/services/profile/inviteUsersToProfile.ts
+++ b/packages/common/src/services/profile/inviteUsersToProfile.ts
@@ -7,6 +7,7 @@ import { User } from '@op/supabase/lib';
 import { waitUntil } from '@vercel/functions';
 import { assertAccess, permission } from 'access-zones';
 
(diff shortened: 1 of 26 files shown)


services/workflows/src/functions/notifications/sendPhaseTransitionNotification.ts
valentin0h, line 1: We need to change the source of email, currently relying on profileUsers.email, but it should be swapped with users.email. To be addressed separately.

packages/common/src/services/profile/listProfileUsers.ts
valentin0h, line 1: I'll address some potential pagination after migration.

packages/common/src/services/profile/inviteUsersToProfile.ts
```diff
@@ -247,7 +247,11 @@ export const inviteUsersToProfile = async ({
   }
 
   const usersByEmail = new Map(
-    existingUsers.map((user) => [user.email.toLowerCase(), user]),
+    existingUsers
+      .filter(
+        (user): user is typeof user & { email: string } => user.email !== null,
```
nourmalaeb: Any reason not to do `existingUsers.filter(hasEmail)` here?

packages/common/src/utils/email.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  it('returns false for an undefined email', () => {
+    expect(hasEmail({ email: undefined })).toBe(false);
+  });
+
+  it('returns false when the email property is absent', () => {
+    expect(hasEmail({})).toBe(false);
+  });
+
+  it('preserves the rest of the row when used as a filter', () => {
```
nourmalaeb:
```suggestion
  it('preserves the rest of the rows when used as a filter', () => {
```

Missing the "s" in "rows"

nourmalaeb (approved):

PR #1236, opened 2026-05-29
feat(resources): drag-and-drop file/link upload into collections

scazan:
Adds drag-and-drop Document upload to the Resources panel. Instead of only clicking **Add Resource**, you can drag a file or link directly onto a collection. This also works for links, allows you to drop at a particular sort point, and can drop multiple files.

There were some missing i18n keys as well that come along with this PR

https://github.com/user-attachments/assets/a12fcf41-b500-42af-81b4-2ac289ce08af

First commit: e0b8f7752da68509dfe6dba558c441fa2804622b


packages/common/src/services/resources/createDocumentResource.ts
```diff
@@ -125,12 +129,21 @@ export const createDocumentResource = async (
...
+            resourceId: row.id,
+            upperNeighborId: input.upperNeighborId,
+            addedByProfileId,
+          })
+        : await insertResourceAtTop({
+            tx,
+            collectionId,
+            resourceId: row.id,
+            addedByProfileId,
+          });
```
valentin0h: Seems like we can work with just `insertResourceAt` in this case?

https://github.com/oneprojectorg/common/blob/d2093e907d94a6bb5c16a555044e31ecc63191e3/packages/common/src/services/resources/ordering.ts#L180

valentin0h (approved):

packages/common/src/services/resources/createDocumentResource.ts
scazan: ah duhh.. thanks.

PR #1237, opened 2026-06-01
feat(a11y): add localized per-page titles

nourmalaeb:
Gives every page a legible, localized `<title>` (WCAG 2.4.2). Before, every tab read `Common`.

## Approach

Server-rendered titles via `generateMetadata` on every page (server `page.tsx` + client child for interactive pages). Server titles resolve as part of navigation, so the tab and route announcer pick them up; client `document.title` effects lose the announcer race on suspense-gated soft navs.

- **Title template**: root layout sets `title.template = '%s | Common'`, `default: 'Common'`. Each page supplies its segment.
- **Localized**: static labels via `getTranslations`, reading `locale` from `params` so soft-nav titles stay translated. New keys across all 7 locales; `(Editing)` composes an atomic `Editing` key; `Review {title}` stays interpolated (word order is language-dependent).
- **Proposal title**: `getProposalDisplayTitle` reads the proposal profile name (kept in sync with the title by create/updateProposal), falling back to the document-resolved title.
- **No duplicate analytics**: the title-relevant reads (`getProposal`, `getDecisionBySlug`, `getInstance`, `getLegacyInstance`) fire a "viewed" event per call. Each is wrapped in a module-level React `cache()` (`lib/decisionRenderData.ts`) and called from **both** `generateMetadata` and the page render. React memoizes per request, so the resolver — and its analytics side effect — runs once per page view, and the shared `queryClient` is dehydrated for client hydration. This is the dedup the Next metadata docs recommend.
- **Prefetch + HydrationBoundary** kept for render UX on the proposal/decision pages.

## Known limitations / follow-ups (separate PRs)
- **Tracking decouple**: "viewed" tracking lives in read resolvers, so non-view callers (`VoteSuccessModal`, ProcessBuilder reads) still emit `process_viewed`. Moving it to an explicit client-fired event would make reads pure and fix that over-counting.
- **Proposal soft-nav announcement**: tab titles are correct everywhere, but on client-side navigation the announcer reads `document.title` at the route-commit and proposal metadata can resolve a beat later, so the screen-reader announcement for proposals can come through blank (other pages announce fine). Not a 2.4.2 failure — the page is titled — but worth closing.


valentin0h (commented):
I'd suggest we move the tracking to the client, which would also make it more accurate since `getProposal,` for example, is called from a variety of places. 

Then we can remove `skipTracking` from this PR and the request payloads. Happy to tackle that together if you want.

nourmalaeb:
> I'd suggest we move the tracking to the client, which would also make it more accurate since `getProposal,` for example, is called from a variety of places.
> 
> Then we can remove `skipTracking` from this PR and the request payloads. Happy to tackle that together if you want.

@valentin0h good suggest, I agree. I've removed `skipTracking` from this PR. We should move the tracking to the client in a separate PR.

apps/app/src/app/[locale]/(main)/org/[slug]/relationships/page.tsx
```diff
@@ -1,5 +1,29 @@
...
+  try {
+    const [client, t] = await Promise.all([
+      createClient(),
+      getTranslations({ locale }),
+    ]);
+    const profile = await client.profile.getBySlug({ slug });
+    const label = t('Relationships');
+    return { title: profile.name ? `${label} | ${profile.name}` : label };
+  } catch {
+    return {};
```
Comment (valentin0h) [inline apps/app/src/app/[locale]/(main)/org/[slug]/relationships/page.tsx:23]: We don't want to log here?

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/edit/page.tsx
```diff
@@ -78,6 +80,16 @@ export default function ProposalEditorLayout({
...
 
+  const proposalTitle = getProposalDisplayTitle(proposal);
+  useEffect(() => {
+    const parts = [
+      proposalTitle ? `${proposalTitle} (${t('Editing')})` : null,
+      decisionProfile.name,
+      APP_NAME,
+    ].filter(Boolean);
+    document.title = parts.join(' | ');
+  }, [proposalTitle, decisionProfile.name, t]);
```
Comment (valentin0h) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/edit/page.tsx]: Wondering why this wouldn't be in generateMetadata?

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/page.tsx
```diff
@@ -1,128 +1,65 @@
...
-  const { profileId, slug } = useParams<{
-    profileId: string;
-    slug: string;
-  }>();
+  // Shares the cache()-wrapped fetch with generateMetadata above, so the
+  // resolver (and its view event) runs once and the data is hydrated.
+  await Promise.all([
+    renderProposal(profileId),
+    renderDecisionBySlug(slug),
+  ]).catch(() => {});
```
Comment (valentin0h) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/page.tsx:66]: Should we swallow these errors?

apps/app/src/components/decisions/serverRenderData.ts
valentin0h, line 1: I would suggest

1. Move these memo functions to the files they are used, some duplication is ok.
2. I'm not sure `render` is self explanatory in this context, what about `fetch` or `get`.

apps/app/src/components/decisions/proposalContentUtils.ts
```diff
@@ -129,3 +129,25 @@ export function resolveProposalSystemFields(proposal: Proposal) {
...
+  proposal: Proposal,
+): string | undefined {
+  // Prefer the proposal profile name: it's kept in sync with the title by
+  // createProposal/updateProposal and is a plain column, so it resolves cheaply.
+  // The document-resolved title runs tiptap over the fragments — keeping it out
+  // of the hot path keeps generateMetadata fast enough to land in the initial
+  // response (not streamed late), so the route announcer reads the real title.
+  // Falls back to the document title, then undefined.
+  return (
+    proposal.profile?.name ||
```
valentin0h: Let's rely only on `profile.name`, do you see any issues with that for this PR? 

I'd remove `getProposalDisplayTitle` completely.

valentin0h (commented):
Added some comments before approval

valentin0h (approved):

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/edit/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/edit/page.tsx]: The proposal editing page is a client component, can't use `generateMetadata` on client components. With the y.js and tiptap and all that I didn't want to change things here at this point. The page title will end up correct in the browser window, but we should fix this in the future

apps/app/src/components/decisions/serverRenderData.ts
nourmalaeb: Done — deleted `serverRenderData.ts`, inlined each cache() wrapper into its page, renamed `render*` → `fetch*`

apps/app/src/components/decisions/proposalContentUtils.ts
nourmalaeb: Agree let's simplify 👍

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/proposal/[profileId]/page.tsx:66]: Intentional — this prefetch only warms the cache for hydration. The client component runs the same query via `useSuspenseQuery` and its `APIErrorBoundary` owns any real error (404 → `notFound`), so a failed warmup must not crash the route. It also does not change error behavior vs. before the prefetch existed — worst case we lose the warmup. Added a comment to make that explicit.

apps/app/src/app/[locale]/(main)/org/[slug]/relationships/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(main)/org/[slug]/relationships/page.tsx:23]: Same theme as the swallow on the proposal page. On a metadata fetch error we return `{}` so the title falls back to the default — the page render fetches the same data right after and surfaces the real error through its boundary / `error.tsx`, so logging here would just duplicate that. Left the catch as-is.

apps/app/src/components/decisions/serverRenderData.ts
nourmalaeb: Done — deleted `serverRenderData.ts`, inlined each `cache()` wrapper into the page that uses it, and renamed `render*` → `fetch*`. The dedup still holds: `cache()` only needs to memoize within a single request, and a route's `generateMetadata` + page body always share a file, so no request spans two of these — the three inlined copies of `fetchProposal` dedup exactly like the shared one did.

apps/app/src/components/decisions/proposalContentUtils.ts
nourmalaeb: Done — removed `getProposalDisplayTitle`; all sites now read `proposal.profile?.name` directly. No issue for this PR: `profile.name` is kept in sync with the title by create/updateProposal, and the `|| t('Untitled Proposal')` fallback covers the empty case.

PR #1239, opened 2026-06-01
fix(api): scope profile search results to the fields the UI needs

valentin0h:
Picks the profile search encoders down to the fields the search UI actually reads, so the endpoint doesn't accidentally expose more data than needed. Derives the profile encoder from baseProfileEncoder so it stays in sync with the public profile shape.


scazan (approved):
Great

PR #1240, opened 2026-06-01
refactor(api): split network/authenticated/open tRPC procedures

valentin0h:
Makes each endpoint's auth posture explicit ahead of opening any public access. Renames the closed-network procedure (commonAuthedProcedure -> networkAuthenticatedProcedure) and adds inert authenticatedConfirmedProcedure, authenticatedProcedure, and openProcedure scaffolding with no endpoint behavior change yet.

### DIFF
--- a/services/api/src/routers/organization/deleteOrganizationUser.ts
+++ b/services/api/src/routers/organization/deleteOrganizationUser.ts
@@ -1,7 +1,7 @@
 import { deleteOrganizationUser } from '@op/common';
 import { z } from 'zod';
 
-import { commonNetworkProcedure, router } from '../../trpcFactory';
+import { networkAuthenticatedProcedure, router } from '../../trpcFactory';
 
 const inputSchema = z.object({
   organizationId: z.uuid(),
@@ -20,7 +20,7 @@ const outputSchema = z.object({
 });
 
 export const deleteOrganizationUserRouter = router({
-  deleteOrganizationUser: commonNetworkProcedure({
+  deleteOrganizationUser: networkAuthenticatedProcedure({
     rateLimit: { windowSize: 60, maxRequests: 5 },
   })
     .input(inputSchema)
--- a/services/api/src/routers/organization/deleteOrganizationUser.ts
+++ b/services/api/src/routers/organization/deleteOrganizationUser.ts
@@ -1,7 +1,7 @@
 import { deleteOrganizationUser } from '@op/common';
 import { z } from 'zod';
 
-import { commonAuthedProcedure, router } from '../../trpcFactory';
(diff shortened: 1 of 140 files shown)


services/api/src/trpcFactory.ts
```diff
@@ -68,29 +71,69 @@ export const commonProcedure = t.procedure
...
+
+/**
+ * Confirmed-user procedure: admits any confirmed, non-anonymous user (a real
+ * account whose email/phone is confirmed) via {@link withConfirmedUser}, but
+ * applies no closed-network/allow-list gating. Sits one tier below
+ * {@link networkAuthenticatedProcedure} (which adds the `@oneproject.org` /
+ * invite allow list) and one above {@link authenticatedProcedure} (which also
+ * admits anonymous sessions). Default rate limit is 10 requests per 10 seconds.
+ */
+export function confirmedProcedure(opts?: RateLimitedProcedureOptions) {
```
valentin0h: maybe `authenticatedConfirmedProcedure`

services/api/src/routers/organization/deleteOrganizationUser.ts
```diff
@@ -20,7 +20,7 @@ const outputSchema = z.object({
 });
 
 export const deleteOrganizationUserRouter = router({
-  deleteOrganizationUser: commonAuthedProcedure({
+  deleteOrganizationUser: networkAuthenticatedProcedure({
     rateLimit: { windowSize: 60, maxRequests: 5 },
```
scazan, line 24: Not related to this PR at all but since I'm here... just noting that we could probably remove all these rate limit overrides unless you have a specific need. At one point we didn't have them in the middleware so I'm guessing these are leftovers from then.

services/api/src/types.ts
```diff
@@ -29,6 +29,12 @@ export interface TContextWithUser {
   user: User;
 }
 
+/** Context after optional user resolution: `user` is the resolved Supabase
+ * identity, or `undefined` when the caller has no valid session. */
+export interface TContextWithMaybeUser {
```
scazan, line 34: ooh, I almost expect to see a Maybe type returned here with the Maybe naming.
Got me thinking. :)

services/api/src/trpcFactory.ts
scazan: yeah, I like that framing better.

scazan (approved):
Exciting work!

services/api/src/routers/organization/deleteOrganizationUser.ts
valentin0h: I recall that, when migrating to the then-new procedures, some were different from the default values, which is why they are still there. I'll do another pass on all now that some will be public.

PR #1243, opened 2026-06-02
feat(access-control): add global public user

valentin0h:
feat(db): seed the GLOBAL_USER_PUBLIC sentinel for no-JWT public callers
feat(access-control): exclude the sentinel from listAllUsers so it never surfaces in user lists
refactor(access-control): move the listAllUsers query into an @op/common service

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -220,18 +220,15 @@ export const genericEmail = 'support@oneproject.org';
 
 export const adminEmails = ['scott@oneproject.org'];
 
-// Sentinel "global" user identities used by the access-control substitution
-// layer. GLOBAL_USER_PUBLIC stands in for no-JWT public callers and
-// GLOBAL_USER_ANONYMOUS for anon-JWT callers when resolving role grants. Their
-// UUIDs are hardcoded so every environment references the same rows; the rows
-// are created by the database seed (services/db/seed-global-users.ts). They are
-// quarantined: cannot sign in, and never surface in user/profile lists.
+// Sentinel "global" user identity used by the access-control substitution
+// layer. GLOBAL_USER_PUBLIC stands in for no-JWT public callers when resolving
+// role grants. Anonymous (anon-JWT) callers are not substituted: they already
+// have a real auth.users identity, so their own id carries their grants and
+// participation. Its UUID is hardcoded so every environment references the same
+// row; the row is created by the database seed (services/db/seed-global-users.ts).
+// It is quarantined: cannot sign in, and never surfaces in user/profile lists.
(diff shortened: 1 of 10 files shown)


services/db/seed-global-users.ts
```diff
@@ -0,0 +1,47 @@
...
+ * It is identity-only — an `auth.users` row + `public.users` mirror, no
+ * profile/owner/Admin scaffolding — so it can never sign in or surface in
+ * search, invites, or people lists. We direct-INSERT (the Admin API can't pin
+ * a UUID) and suppress `on_auth_signup_create_user` via SET LOCAL
+ * session_replication_role rather than DISABLE TRIGGER, since `auth.users` is
+ * owned by `supabase_auth_admin`. Idempotent: safe to re-run.
+ */
+export async function seedGlobalUsers(): Promise<void> {
+  // TODO: protect this row from deletion.
+  const sentinels = [{ id: GLOBAL_USER_PUBLIC, name: 'Public' }];
```
valentin0h, line 22: Maybe a better naming would be `Visitor`

scazan: Is the name surfaced anywhere? You could also just have both. "Public Visitor" but Public is also fine. it communicates well.

packages/core/src/config.ts
```diff
@@ -220,6 +220,16 @@ export const genericEmail = 'support@oneproject.org';
...
 export const adminEmails = ['scott@oneproject.org'];
 
+// Sentinel "global" user identity used by the access-control substitution
+// layer. GLOBAL_USER_PUBLIC stands in for no-JWT public callers when resolving
+// role grants. Anonymous (anon-JWT) callers are not substituted: they already
+// have a real auth.users identity, so their own id carries their grants and
+// participation. Its UUID is hardcoded so every environment references the same
+// row; the row is created by the database seed (services/db/seed-global-users.ts).
+// It is quarantined: cannot sign in, and never surfaces in user/profile lists.
+export const GLOBAL_USER_PUBLIC = '00000000-0000-4000-a000-000000000001';
```
scazan, line 230: Do we need a migration for this as well?

scazan (approved):
nice

packages/core/src/config.ts
valentin0h: I'm thinking of not having this in migrations. I see it more as a seed. We run it separately in each env for existing envs.

services/db/seed-global-users.ts
valentin0h: It shouldn't surface anywhere. Keeping for now, easy to change.

PR #1245, opened 2026-06-02
refactor(common): collapse profile fetch-and-assert into assertProfileAccess

valentin0h:
refactor(common): collapse profile fetch-and-assert into a single assertProfileAccess utility
refactor(common): instance profile access uses the new assertProfileAccess
refactor(common): unify profile-access denials on AccessControlException


packages/common/src/services/assert/assertProfileAccess.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function assertProfileAccess(
+  { user, profileId }: { user: { id: string }; profileId: string },
+  permissions: AccessZonePermissionInput,
+  notMemberMessage?: string,
+): Promise<ProfileUserWithNormalizedRoles> {
+  const profileUser = await getProfileAccessUser({ user, profileId });
+
+  if (!profileUser) {
+    throw new AccessControlException({
+      message: notMemberMessage ?? 'Not authenticated',
```
scazan: This one is a bit weird because we are authenticated at this point. We just don't have authorization on the profile. We should probably be throwing an UnauthorizedError error here like we had previously.

packages/common/src/services/assert/assertProfileAccess.ts
```diff
@@ -0,0 +1,39 @@
...
+ * returning the resolved profile-access user so callers can reuse it.
+ *
+ * @param notMemberMessage - Optional message for the thrown exception when the
+ *   user has no role on the profile. Defaults to the access-zones denial message.
+ * @throws AccessControlException if the user is not a member of the profile or
+ *   their roles don't satisfy the permissions — every denial throws the same
+ *   exception type (only the message differs when `notMemberMessage` is given).
+ */
+export async function assertProfileAccess(
+  { user, profileId }: { user: { id: string }; profileId: string },
```
scazan: Just food for thought here.. why not just make them all named parameters into the single object and pass the User type with Partial:

```suggestion
  { user, profileId, permissions, notMemberMessage }: { user: User; profileId: string; permissions: AccessZonePermissionInput; notMemberMessage?: string },
```
We probably want the User type here as well since we could inadvertently pass the wrong "user" type here (or anything really that has an id). That would allow us to throw a type error in that situation and not be passing a non authUser... It also nicely signals that there is some tension here with the dual users (our Common user and the Auth user from Supabase).

scazan (commented):
The two things I think we should do before merge are:
- I think we should update that type to User as it could block an easy mistake to make. 
- I think it should be an UnauthorizedError here as a missing profileUser doesn't imply an authentication error but a real no permissions error.
Open to discuss though.
The other comment is more about the parameter shape since it is mixing named params with ordered params. When that happens I switch to all named params.. in most cases I think outside of a single param I prefer named params on everything.

packages/common/src/services/assert/assertProfileAccess.ts
valentin0h: This is mimicking what assertAccess throws. Do you think we should differentiate between a user who is a member and doesn't have permission and a user who is not a member? Could be good, but don't see an immediate use case. Ok to go either way.

packages/common/src/services/assert/assertProfileAccess.ts
valentin0h: Yes, `User` sounds great.

valentin0h: I'll stack a new PR on this because it's a bit more involved than just changing the type there.

packages/common/src/services/assert/assertProfileAccess.ts
scazan: Yeah, and looking closer AccessControlException _does_ make sense (same code as Unauthorized anyway) but we probably want this message here to be "Not authorized" instead of "Not authenticated" (what originally made me stop on this line). I'm thinking that maybe we should be catching and rethrowing the Common UnauthorizedError here rather than the library AccessControlException to keep it consistent with the Common errors. Then this function only throws the one type and Common doesn't need to know about AccessControlException from the external dependency. We can think of a more elegant solution as well since we control the library.

scazan (approved):
Approving in advance of the changes to unblock

packages/common/src/services/assert/assertProfileAccess.ts
valentin0h: Sounds good. The library does indeed throw that error message, I think. I've been meaning to discuss changing that. But rethrowing is a better solution.

PR #1248, opened 2026-06-02
feat(decisions): add decision overview page shell

nourmalaeb:
Scaffolds a decision **overview** page behind the `decision_overview` flag.

- New `(decision-view)` route group with a shared layout — persistent header + Overview/Current Phase toggle (`@op/sense` Tabs). Routes: `/overview` and `/current`. Instance is prefetched once so tab switches are instant.
- Canonical root `/decisions/[slug]` is unchanged (still the current-phase page, no flag code).
- Whole branch gated **server-side** (`posthog-node`, distinct id from the PostHog cookie); flag off → redirect to root before render, no flash.

Overview body is still a placeholder — next pass.

---
  - https://app.asana.com/0/0/1215289426908478


apps/app/src/components/decisions/DecisionViewToggle.tsx
```diff
@@ -0,0 +1,53 @@
+'use client';
+
+import { Tabs, TabsList, TabsTrigger } from '@op/sense/Tabs';
```
nourmalaeb, line 3: Using the new shadcn components from the `@op/sense` package 😎

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const [{ decisionProfile }, t] = await Promise.all([
+      loadDecision(slug),
+      getTranslations({ locale }),
+    ]);
+
+    // Title the page with the active phase's name; fall back to a generic
+    // label if it can't be resolved. Phase data rides along on the decision
+    // profile, so no extra instance fetch is needed.
+    const { instanceData, currentStateId } = decisionProfile.processInstance;
+    const currentPhase = instanceData?.phases?.find(
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx:27]: We can use our assertion here instead: ```  const currentPhase = assertInstancePhase({
    instance: decisionProfile.processInstance;,
    phaseId: currentStatusId,
  });```

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
```diff
@@ -0,0 +1,61 @@
...
+  params: Promise<{ slug: string; locale: string }>;
+}): Promise<Metadata> {
+  const { slug, locale } = await params;
+
+  try {
+    const [{ decisionProfile }, t] = await Promise.all([
+      loadDecision(slug),
+      getTranslations({ locale }),
+    ]);
+    const label = t('Overview');
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx:22]: Hmmm.. not sure this will totally work if we have more complex natural keys with punctuation. Luckily here you don't so it'll work but noting that we do some parsing for translation keys with punctuation which is skipped if use the server version.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts
```diff
@@ -0,0 +1,43 @@
+import { createClient } from '@op/api/serverClient';
+import { CommonError } from '@op/common';
+import { forbidden, notFound } from 'next/navigation';
+import { cache } from 'react';
+
+/**
+ * Resolve a decision profile by slug for the (decision-view) route group.
+ *
+ * Wrapped in React `cache` so the layout and the active page (overview or
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts]: We don't need this comment

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/error.tsx
```diff
@@ -0,0 +1,7 @@
+'use client';
+
+import PageError, { ErrorProps } from '@/components/screens/PageError';
+
+export default function Error(props: ErrorProps) {
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/error.tsx]: Just curious if the reasoning we don't let it inherit the higher one here is for the header?

apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -84,19 +90,22 @@ function DecisionHeaderContent({
...
-            <DecisionProcessStepper
-              phases={phases}
-              currentStateId={instance.currentStateId || ''}
-              instanceId={instanceId}
-              isAdmin={isAdmin}
-              className="mx-auto"
-            />
+        {showStepper ? (
+          <div className="flex flex-col overflow-x-auto sm:items-center">
+            <div className="w-fit rounded-b border border-t-0 bg-white px-12 py-4 sm:px-32">
```
scazan: Is this our standard header? Maybe good ot pull it out to a component

apps/app/src/components/decisions/DecisionStepperBar.tsx
```diff
@@ -0,0 +1,50 @@
...
+    phase: {
+      startDate: p.startDate,
+      endDate: p.endDate,
+    },
+    advancementMethod: p.rules?.advancement?.method,
+  }));
+
+  return (
+    <div className="flex flex-col overflow-x-auto sm:items-center">
+      <div className="w-fit rounded-b border border-t-0 bg-white px-12 py-4 sm:px-32">
```
scazan, line 24: Seeing this multiple times so good to go ahead and pull it out.

apps/app/src/components/decisions/DecisionViewToggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * Each segment is a route link, so the toggle anchors the user as they move
+ * between /decisions/[slug] (overview) and /decisions/[slug]/current. The
+ * active segment is derived from the path — the toggle lives in the shared
+ * layout, so it has no per-page prop telling it which tab is active.
+ */
+export function DecisionViewToggle({ decisionSlug }: DecisionViewToggleProps) {
+  const t = useTranslations();
+  const pathname = usePathname();
+  const activeView =
+    pathname.split('/').pop() === 'current' ? 'current' : 'overview';
```
scazan: It feels a bit shaky to split on the pathname here... something smells funny here that I need to think about for a sec. just commenting on it.

apps/app/src/lib/getServerFeatureFlag.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  return client;
+}
+
+/**
+ * Read the PostHog distinct id the browser SDK persists, so server-side flag
+ * evaluation targets the same identity the client does. posthog-js stores it
+ * as JSON under `ph_<project_key>_posthog`.
+ */
+async function getDistinctId(): Promise<string | null> {
```
scazan, line 23: I'm pretty sure we are already doing this for other server-side posthog events that exist already. We should re-use it.

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Overview tab for a decision process. Rendered at the decision root
+ * (/decisions/[slug]) by DecisionRootView when the overview flag is on; the
+ * current phase moves to /current.
+ *
+ * Scaffold: renders the instance title + a placeholder body. Flesh out with
+ * the full process summary (hero, proposals, phases, About) next.
+ */
+export function DecisionOverview({ instanceId }: DecisionOverviewProps) {
```
scazan: It's really nice to keep the standard of `DecisionOverviewSuspense` so it's visible to see that this component will suspend. It really clearly signals the boundary and invites a Suspense boundary wrapper in the consuming component.

apps/app/src/components/decisions/DecisionInstanceHeader.tsx
```diff
@@ -30,27 +32,47 @@ export const DecisionInstanceHeader = ({
   decisionSlug?: string;
   isAdmin?: boolean;
   canReadUpdates?: boolean;
+  /**
+   * Optional content for the header's center column (e.g. the Overview /
+   * Current Phase toggle). When provided, the title moves beside the Back
+   * link so the center stays reserved for the slot; otherwise the title is
+   * centered as before.
+   */
+  centerSlot?: ReactNode;
```
scazan, line 41: Just wondering if using a slot here can be avoided by composing the components a little better. as in pass children and then compose these components further up. It keeps the component flexible and free from logic branches that can spiral out into complexity.

scazan (approved):

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx:22]: We control the page title, we can ensure the keys (`Overview`, `Current Phase`) have no punctuation so that the server `getTranslations` path works.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/error.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/error.tsx]: Yeah it's for the header, but actually we don't need to preserve the header when the page errors out. I'll remove this and let it inherit the higher one.

apps/app/src/components/decisions/DecisionStepperBar.tsx
nourmalaeb: Done — pulled the stepper-bar wrapper into DecisionStepperBar (now presentational, takes phases/currentStateId/instanceId/isAdmin). Both DecisionHeader variants render it, so the markup lives in one place.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx:27]: That's a server side util, the types don't match cleanly and I don't want to use an assertion so the find works here. Punted for now because of the type split, + it throws and we want the soft fallback here.

apps/app/src/components/decisions/DecisionInstanceHeader.tsx
nourmalaeb: Good call. This reshapes the shared header API (used by the old root page, legacy, and results views), so I'd rather do it as a focused follow-up than fold it into this batch. Tracking it.

apps/app/src/lib/getServerFeatureFlag.ts
nourmalaeb: Found it 🙏

PR #1252, opened 2026-06-04
refactor(common): collapse org fetch-and-assert into assertOrgAccess

valentin0h:
Mirror the profile-side assertProfileAccess utility for organizations so org-access denials are fetched, checked, and unified on UnauthorizedError in one place.

### DIFF
--- a/packages/common/src/services/profile/listProfileUserInvites.ts
+++ b/packages/common/src/services/profile/listProfileUserInvites.ts
@@ -17,10 +17,10 @@ export const listProfileUserInvites = async ({
   user: User;
   query?: string;
 }) => {
-  await Promise.all([
-    assertProfileAdmin({ user, profileId }),
-    assertProfile(profileId),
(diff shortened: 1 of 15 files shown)


packages/common/src/services/profile/listProfileUserInvites.ts
```diff
@@ -17,10 +17,10 @@ export const listProfileUserInvites = async ({
...
   query?: string;
 }) => {
-  await Promise.all([
-    assertProfileAdmin({ user, profileId }),
-    assertProfile(profileId),
-  ]);
+  // Check existence before access so a nonexistent profile is a 404
+  // regardless of the caller's permissions.
+  await assertProfile(profileId);
+  await assertProfileAdmin({ user, profileId });
```
valentin0h, line 23: I'll address this separately since we have this race error condition in a few places

scazan (approved):

PR #1254, opened 2026-06-04
refactor(common): widen access-layer caller to optional AccessUser

valentin0h:
Lays the rails for the upcoming no-JWT/public-user substitution (auth-refactor PR4) by making the access-layer caller type optional, without changing current behavior — nothing passes undefined yet, and resolvers fail closed on a missing caller id.

### DIFF
--- a/packages/common/src/services/access/index.ts
+++ b/packages/common/src/services/access/index.ts
@@ -22,15 +22,30 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
   profile: ProfileMinimal;
 };
 
+/**
+ * The caller identity the access layer needs. A subset of the Supabase `User`
+ * (only the id is read today); widen the `Pick` as later auth work needs more
+ * fields. Optional throughout the access layer so a future no-JWT (public)
+ * caller can be represented as `undefined` — resolvers fail closed on it.
+ */
+export type AccessUser = Pick<User, 'id'>;
+
 // gets a user assuming that the user is authenticated
 export const getOrgAccessUser = memoize(
   async ({
     user,
     organizationId,
   }: {
-    user: { id: string };
+    user?: AccessUser;
     organizationId: string;
   }): Promise<OrgUserWithNormalizedRoles | undefined> => {
+    // No caller identity → fail closed. Never let an undefined id reach the
+    // query: Drizzle skips undefined filter conditions, which would drop the
+    // authUserId constraint and match an arbitrary member (auth bypass).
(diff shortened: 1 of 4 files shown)


packages/common/src/services/access/index.ts
```diff
@@ -22,15 +22,30 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
...
-    user: { id: string };
+    user?: AccessUser;
     organizationId: string;
   }): Promise<OrgUserWithNormalizedRoles | undefined> => {
+    // No caller identity → fail closed. Never let an undefined id reach the
+    // query: Drizzle skips undefined filter conditions, which would drop the
+    // authUserId constraint and match an arbitrary member (auth bypass).
+    if (!user?.id) {
+      throw new UnauthorizedError();
+    }
```
valentin0h, line 47: To be changed in the following PRs that will substitute with the public user.

packages/common/src/services/access/index.ts
```diff
@@ -79,7 +94,7 @@ export const getOrgAccessUser = memoize(
       },
     });
   },
-  ({ user, organizationId }) => `${user.id}:${organizationId}`,
+  ({ user, organizationId }) => `${user?.id}:${organizationId}`,
```
valentin0h, line 97: In the following PR, it won't be `user?.id.`

packages/common/src/services/access/index.ts
```diff
@@ -22,15 +22,30 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
   profile: ProfileMinimal;
 };
 
+/**
+ * The caller identity the access layer needs. A subset of the Supabase `User`
+ * (only the id is read today); widen the `Pick` as later auth work needs more
+ * fields. Optional throughout the access layer so a future no-JWT (public)
+ * caller can be represented as `undefined` — resolvers fail closed on it.
+ */
+export type AccessUser = Pick<User, 'id'>;
```
valentin0h, line 31: This felt balanced naming.

scazan: I like the naming! I'm wondering though if we should be Picking roles here as well since it's one of the main use cases for it. It's the defining aspect of these and the shape is shared in profileUsers and the (now legacy) orgUsers.

scazan (approved):

packages/common/src/services/access/index.ts
valentin0h: What do you mean with roles in this case?

PR #1256, opened 2026-06-05
refactor(analytics): track proposal/process views client-side

valentin0h:
The proposal_viewed and process_viewed events fired inside the getProposal/getInstance query procedures, so they re-fired on every prefetch, refetch, and cache invalidation (including when the proposal editor or review summary fetched the same query). Moving them into the view components fires them once per viewed entity and lets the client-side common properties (location/timezone/language) populate.

### DIFF
--- a/packages/analytics/src/utils.ts
+++ b/packages/analytics/src/utils.ts
@@ -1,5 +1,10 @@
 import PostHogClient from '@op/analytics/client';
 
+import {
+  type DecisionCommonProperties,
+  getDecisionCommonProperties,
+} from './client-utils';
+
 const posthog = PostHogClient();
 
 /**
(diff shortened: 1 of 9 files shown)


packages/analytics/src/utils.ts
```diff
@@ -365,7 +353,11 @@ export async function trackProposalLiked(
   await trackEventWithContext(
     userId,
     'proposal_liked',
-    getDecisionCommonProperties(processId, proposalId, additionalProps),
+    getDecisionCommonProperties({
+      decisionInstanceId: processId,
```
scazan, line 357: Love these becoming named params.

scazan (approved):

PR #1257, opened 2026-06-05
feat(api): open public-decision reads + decision posts to no-JWT/anonymous callers

valentin0h:
feat(api): open public-decision reads and decision posts to no-JWT/anonymous callers

### DIFF
--- a/packages/common/src/services/access/index.ts
+++ b/packages/common/src/services/access/index.ts
@@ -1,4 +1,5 @@
 import { cache } from '@op/cache';
+import { GLOBAL_USER_PUBLIC } from '@op/core';
 import { db, eq } from '@op/db/client';
 import { organizations, users } from '@op/db/schema';
 import type { User } from '@op/supabase/lib';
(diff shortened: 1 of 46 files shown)


packages/common/src/services/access/index.ts
```diff
@@ -30,6 +31,77 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
...
+ * is part of the identity, so a stale `[organizationId, user.id]` key would miss
+ * and serve removed/demoted members their old roles until TTL.
+ */
+export const orgUserCacheKey = ({
+  user,
+  organizationId,
+}: {
+  user?: AccessUser;
+  organizationId: string;
+}): [string, string] => [organizationId, resolveAccessUserIds(user).join(':')];
```
scazan, line 66: I'm wondering why we need a special key here since all users should probably inherit the rights of the GLOBAL_PUBLIC_USER. In that case, this is still just their role and no need to collect the ids into a separate cache key. This doesn't really have an effect I'd imagine unless there are spots where we don't merge in the role in which case we have two cache keys to invalidate which would be harder to deal with.

packages/common/src/services/access/index.ts
```diff
@@ -30,6 +31,77 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
...
+    roles: Parameters<typeof getNormalizedRoles>[0];
+  },
+>(
+  rows: TRow[],
+  user?: AccessUser,
+): { baseRow: TRow; normalizedRoles: NormalizedRole[] } => {
+  const ownRow = rows.find((row) => row.authUserId === user?.id);
+  return {
+    baseRow: ownRow ?? rows[0]!,
+    normalizedRoles: rows.flatMap((row) => getNormalizedRoles(row.roles)),
```
scazan, line 101: `their own u the public..` seems like some missing letters there. not sure what `u` is.

packages/common/src/services/access/index.ts
```diff
@@ -30,6 +31,77 @@ export type ProfileUserWithNormalizedRoles = ProfileUserBase & {
...
+export const profileUserCacheKey = ({
+  user,
+  profileId,
+}: {
+  user?: AccessUser;
+  profileId: string;
+}): [string, string] => [profileId, resolveAccessUserIds(user).join(':')];
+
+/**
+ * Collapse the grant rows matched for a caller (their own ∪ the public
```
scazan, line 84: This "Collapse" comment threw me a little because our access library also has a concept of collapsing roles but it's not the same here. We are collecting roles rather than collapsing them.

packages/common/src/services/decision/listProposals.ts
```diff
@@ -157,18 +158,28 @@ export const listProposals = async ({
...
-    throw new UnauthorizedError('User must be authenticated');
+  // Resolve the caller's profile once; it's reused for ballot auth, the
+  // HIDDEN visibility filter, and owner/editable checks further down. Public
+  // (no-JWT) and anonymous callers have no account profile — treat as none.
+  let currentProfileId: string | undefined;
+  if (user) {
+    try {
+      currentProfileId = await getCurrentProfileId(user.id);
+    } catch {
+      currentProfileId = undefined;
```
scazan, line 173: Is this actually necessary?

packages/common/src/services/decision/listProposals.ts
```diff
@@ -157,18 +158,28 @@ export const listProposals = async ({
...
+    }
   }
 
-  // Resolve the caller's profile once; it's reused for ballot auth, the
-  // HIDDEN visibility filter, and owner/editable checks further down.
-  const currentProfileId = await getCurrentProfileId(user.id);
+  // Caller's own grants unioned with public (GLOBAL_USER_PUBLIC) grants — used
+  // for the draft and HIDDEN visibility subqueries below.
+  // INVARIANT: public grants must only be placed on the process/decision
```
scazan, line 179: oof. this feels a bit risky. I wonder if we should limit HIDDEN and draft to edit permissions in that case. Because read access makes sense here on the proposal directly.

packages/common/src/services/decision/voting.ts
```diff
@@ -337,17 +337,21 @@ export const submitVote = async ({
...
-
   try {
-    const profileId = await getIndividualProfileId(authUserId);
+    // Public / anonymous callers have no individual profile (→ no ballot).
+    let profileId: string | undefined;
+    if (user) {
+      try {
+        profileId = await getIndividualProfileId(user.id);
+      } catch {
+        profileId = undefined;
```
scazan, line 352: same thing. I wonder what this does since profileId is already undefined.

packages/common/src/services/decision/voting.ts
```diff
@@ -382,27 +386,33 @@ export const getVotingStatus = async ({
...
-        voteProposals: {
-          with: {
-            proposal: {
-              with: {
-                profile: {
-                  columns: {
-                    name: true,
+    let voteSubmission = null;
+    if (profileId) {
+      voteSubmission = await db._query.decisionsVoteSubmissions.findFirst({
```
scazan, line 391: Should you switch to relations v2 here?

services/api/src/routers/decision/proposals/get.ts
```diff
@@ -1,14 +1,18 @@
...
 
-import { networkAuthenticatedProcedure, router } from '../../../trpcFactory';
+import { openProcedure, router } from '../../../trpcFactory';
 
 export const getProposalRouter = router({
-  getProposal: networkAuthenticatedProcedure()
+  /**
+   * NOTE: not wrapped in a shared `cache()` here. The cache key is keyed by
+   * profileId only (no caller identity), so a cache hit would serve the
+   * proposal to a non-member and bypass the authz inside `getProposal`. The
```
scazan, line 12: `authz` let's just use the few extra characters to make this shortening clear :) As much as possible we should avoid acronyms and unnecessary shortenings.

scazan (approved):

PR #1261, opened 2026-06-07
feat(db): add location geometry column to proposals

bloudermilk:
## Summary

Database-only changes for the upcoming proposal **Location** field ([Asana task](https://app.asana.com/0/0/1214496568127695/f)), split out so migrations can land and deploy ahead of the application code.

- Add `location geometry(point, 4326)` to `decision_proposals` — a denormalized projection of `proposalData.location` (`{ lat, lng }`) for PostGIS queries
- Column flows into `decision_proposal_history` automatically via the shared `proposalColumns` spread
- GiST index `proposals_location_gist_idx`
- Custom migration updates the `create_proposal_history()` trigger to copy `NEW.location` into history snapshots

The column is nullable and nothing references it yet, so this is safe to merge independently.

## Stacked PRs

This is **PR 1 of 2**. The application code (template builder, participant stub, write-path sync) follows in the stacked PR from `issue-1214496568127695`.

## Test plan

- `pnpm w:db generate` output reviewed; both migrations applied + verified against the local docker stack (columns, index, and trigger confirmed via psql)
- Unit tests for the projection helper live in the follow-up PR alongside the code


bloudermilk:
No longer necessary as we're using profiles location rather than a new column.

PR #1262, opened 2026-06-07
feat(decisions): add location field to proposal templates

bloudermilk:
## Summary

Adds a **Location** field to decision-making proposal templates. Asana tasks:
* https://app.asana.com/1/1108348036672214/project/1215095416548888/task/1214496568127695?focus=true
* https://app.asana.com/1/1108348036672214/project/1215095416548888/task/1214497793161444?focus=true
* https://app.asana.com/1/1108348036672214/project/1215095416548888/task/1214497793161424?focus=true

## Open Issues

- [x] Should we revert Posthog CORS workaround or get permission to integrate
- [ ] I plan to refactor map components as there's unnecessary duplication now
- [x] How to handle marker placements that don't resolve to an address?

## Pre-deploy checklist

- [ ] Merge and run #1294 migrations
- [ ] Set `NEXT_PUBLIC_MAPTILER_API_KEY` to an organization-owned API key
- [ ] Enable `gis_maps` feature flag

## Post-deploy checklist

- [ ] Seed Columbus city council districts [geojson](https://opendata.columbus.gov/datasets/columbus::columbus-city-council-districts/explore?location=39.967058%2C-82.978761%2C15) with `import:boundaries`


bloudermilk:
@scazan first pass at proposal location backend and template UX. Some assumptions I made which should be verified:
* Not all proposals will have a location, so it is added manually rather than automatically with every template
* Only one location field can be added to a proposal
* When added to a template, it is always a required field

Also of note:
* The migrations are defined in a separate PR #1261 which should be merged first, but are also included in this branch
* This PR does not address the map component/location selection UI

Looking forward to any feedback you might have before I move forward with the map UI and other scope.

bloudermilk:
### Added: automatic council-district tagging

This branch now also includes the district-boundary feature (commits `feat(db): add decision_boundaries schema` + `feat(decisions): auto-tag proposals by district boundary with location validation`):

- Proposals are auto-tagged with the council-district category whose boundary contains the placed pin (`ST_Contains`), surfaced on the map badge and the existing category filter.
- Location is validated against boundaries on submit (required when the template collects a location).
- A `pnpm --filter @op/common import:boundaries` CLI imports district GeoJSON.

**Deploy ordering:** depends on #1294 (the `decision_boundaries` migration) being **merged and deployed first**. This branch carries the Drizzle *schema* so it compiles, but the migration itself ships in #1294 — merge #1294, then sync this branch with `dev` before merging.

bloudermilk:
This PR is far from perfect but it's getting big enough that a review would be really helpful!

apps/app/src/components/collaboration/CollaborativeDropdownField.tsx
```diff
@@ -20,6 +20,18 @@ interface CollaborativeDropdownFieldProps {
...
+   * boundary-resolved council district). `undefined` means "no auto-resolution
+   * settled yet — don't touch the selection"; `null` means "resolved to no
+   * value". When the auto value changes, the previously auto-applied value is
+   * swapped out; a value the user chose manually is otherwise preserved.
+   *
+   * Routing this through the field (rather than a second collaborative writer
+   * on the same fragment) keeps a single writer per fragment — two writers race
+   * under Yjs and duplicate the stored text.
+   */
+  autoValue?: string | null;
```
bloudermilk: Without knowing the internals of Yjs I had a hard time verifying the need for this, but the bug it fixed was real. The District auto-selection feature writes to the categories field on the client when matched, which was causing strange behavior. Maybe there's a more paradigmatic way to handle this feature?

services/api/src/routers/taxonomy/reverseGeocode.ts
```diff
@@ -0,0 +1,102 @@
...
+      throw new Error(
+        `Google Geocoding API error: ${data.error_message || data.status || response.statusText}`,
+      );
+    }
+
+    const place = data.results?.[0];
+
+    // Open water, deserts, etc. return no address — the caller falls back to a
+    // bare coordinate.
+    if (!place) {
```
bloudermilk: Should we avoid persisting locations that can't be reverse geocoded?

services/api/src/routers/taxonomy/index.ts
```diff
@@ -1,5 +1,10 @@
 import { mergeRouters } from '../../trpcFactory';
 import { getGeoNames } from './geoNames';
+import { reverseGeocode } from './reverseGeocode';
```
bloudermilk, line 3: Is this the correct router for this functionality? Reverse geocoding is used for persisting locations. I'm guessing geonames is in here because those get persisted as a taxonomy?

packages/ui/src/components/Menu.tsx
```diff
@@ -50,7 +50,7 @@ export const menuItemStyles = tv({
     },
     isDisabled: {
       false: 'text-neutral-black',
-      true: 'text-neutral-gray2',
+      true: 'cursor-default text-neutral-gray2',
```
bloudermilk, line 53: Menu headings were using `cursor-pointer` but had no associated action.

packages/common/src/services/decision/validateProposalAgainstTemplate.ts
```diff
@@ -50,10 +54,12 @@ export async function validateProposalAgainstTemplate(
...
     schemaValidator.assertProposalData(proposalTemplate, validationData);
-  } else {
-    schemaValidator.assertProposalData(proposalTemplate, {
-      ...storedProposalData,
-      ...(shouldInjectTitle ? { title } : {}),
-    });
+    return validationData;
   }
+
+  schemaValidator.assertProposalData(proposalTemplate, {
```
bloudermilk, line 69: Not actually clear to me whether/why this change was needed.

bloudermilk (commented):
I left some comments in areas with known quirks or questions. To be frank, between this being a new project for me and using Claude to generate most of the code, there are a lot of aspects I don't fully understand. I'm very curious to learn how code reviews work on your team and am very open to any feedback you all have on the results or process.

services/api/src/routers/taxonomy/reverseGeocode.ts
scazan: We should certainly store them, however, it's ok to not include the placeId for sure.

apps/app/src/components/decisions/location/dynamicMap.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Client-only handle to {@link MapCanvas}. `ssr: false` keeps `maplibre-gl` out
+ * of the server bundle entirely — it is fetched and compiled on the client only
+ * when a location field actually mounts.
+ */
+export const MapCanvas = dynamic(() => import('./MapCanvas'), {
+  ssr: false,
+  loading: () => (
+    <div className="h-44 sm:h-80 w-full animate-pulse border border-neutral-gray1 bg-neutral-gray1" />
```
scazan: We can use our `<Skeleton>` component here

apps/app/src/components/decisions/location/LocationMapField.tsx
```diff
@@ -0,0 +1,172 @@
...
+      // The user placed the pin directly — clear any stale search result.
+      setSearchResetToken((token) => token + 1);
+
+      // Commit the dropped coordinate immediately so the controlled marker
+      // sticks at the new spot instead of snapping back to its prior position
+      // while reverse geocoding is in flight.
+      onChange({ lat: lngLat.lat, lng: lngLat.lng });
+
+      try {
+        const { geoname } = await utils.taxonomy.reverseGeocode.fetch({
```
scazan: We should lean into useSuspenseQuery and useQuery instead of fetch. This bakes in react-query so we get all the benefits of caching. We can avoid the try/catch there as a result.

apps/app/src/components/decisions/location/LocationMapField.tsx
```diff
@@ -0,0 +1,172 @@
...
+          <div className="flex min-w-0 flex-col gap-0.5">
+            <span
+              className={
+                isWithinArea ? 'text-neutral-black' : 'text-functional-red'
+              }
+              dir="auto"
+            >
+              {/* While reverse geocoding is in flight `address` is undefined —
+                  render nothing rather than exposing raw coordinates. */}
+              {value ? value.address : t('No location selected')}
```
scazan, line 163: Will this show "No location selected" for all pins that can't be reverse geocoded?

apps/app/src/components/decisions/location/LocationMapField.tsx
```diff
@@ -0,0 +1,172 @@
...
+            </span>
+            {!isWithinArea && (
+              <span className="text-sm text-functional-red">
+                {t('This address is outside the allowed proposal area.')}
```
scazan, line 167: Is there a way we can block this instead? maybe don't allow a pin to be placed outside the districts?

apps/app/src/components/decisions/location/mapConfig.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/** Fallback camera target before a location is chosen (Bexley, OH). */
+export const DEFAULT_MAP_CENTER: LngLat = { lng: -82.9371, lat: 39.9686 };
+
+/**
+ * Starting camera for the Location field's "Map view" config before an admin
+ * has positioned it: the whole globe, centered on Columbus, OH.
+ */
+export const DEFAULT_LOCATION_FIELD_MAP_VIEW: MapDefaultView = {
+  center: { lng: -82.9988, lat: 39.9612 },
```
scazan, line 24: We probably want this to be configured via either the process or set through "Use my location". More of a comment for the future here and maybe to add a TODO:

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
```diff
@@ -161,6 +162,12 @@ export function TemplateEditorContent({
     [template],
   );
 
+  // Single-instance field types that can't currently be added
+  const disabledTypes = useMemo<FieldType[]>(
```
scazan, line 166: any reason we want to limit this? We can already gate it via the feature flag

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/FieldCard.tsx
```diff
@@ -203,7 +208,8 @@ export function FieldCard({
               <span className="text-neutral-charcoal">{t('Required?')}</span>
               <ToggleButton
                 size="small"
-                isSelected={field.required}
+                isSelected={isLocation || field.required}
```
scazan, line 211: We can keep this optionally required. They'll want to keep it required when they setup their instance. Same with below since not-required should be an option for this.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/FieldConfigLocation.tsx
```diff
@@ -0,0 +1,62 @@
...
+}: FieldConfigProps) {
+  const t = useTranslations();
+
+  const initialView = useRef(
+    fieldSchema['x-map-default'] ?? DEFAULT_LOCATION_FIELD_MAP_VIEW,
+  );
+
+  return (
+    <div className="space-y-2">
+      <h4 className="text-sm text-neutral-charcoal">{t('Map view')}</h4>
```
scazan: Small thing, we can use the `<HeaderX>` components here. If there is no `<Header4>` then one can be added. Most headers end up with the text-title-sm instead but it could be that there is an exception here (would have to check the figma)

apps/app/src/components/decisions/proposalEditor/ProposalFormRenderer.tsx
```diff
@@ -407,12 +487,30 @@ export function ProposalFormRenderer({
...
+  const gisMapsEnabled = useFeatureFlag('gis_maps');
   const formGapClass = mode === 'preview-template' ? 'gap-4' : 'gap-8';
 
   const titleField = fields.find((f) => f.key === 'title');
   const categoryField = fields.find((f) => f.key === 'category');
   const budgetField = fields.find((f) => f.key === 'budget');
-  const dynamicFields = fields.filter((f) => !f.isSystem);
+  // The location field lives behind the `gis_maps` flag. Filtering it out here
+  // also leaves `locationField` undefined, so no district is auto-resolved.
+  const dynamicFields = fields.filter(
```
scazan, line 464: small nit.. this could be done in one go with a reduce :) This way it could also support multiple locations

apps/app/src/components/decisions/proposalEditor/ReadonlyProposalFields.tsx
```diff
@@ -93,6 +93,37 @@ export function ReadonlyDropdownField({
...
+}: {
+  /** Preformatted coordinates, e.g. `"12.34, 56.78"`. */
+  value: string | null;
+  title?: string;
+  description?: string;
+  placeholder: string;
+}) {
+  const content = (
+    <Button variant="pill" color="pill" className="justify-start text-start">
+      {value ?? placeholder}
```
scazan: Is the LocationField only rendering as a pill button? Should this be a map?

apps/app/src/components/decisions/proposalTemplate.ts
```diff
@@ -493,6 +516,10 @@ export function ensureLockedFields(
   } else {
     currentRequired.delete('category');
   }
+  // The location field is always required when present
```
scazan, line 519: We don't want to require by default. Better to have a toggle like we do with the others. On this note, I'm wondering if we can simply allow multiple locations so that this becomes a bit simpler (the profiles support multiple locations in the backend already). We can simply allow it to be a dynamic field like the others. Where we do they "sync" we can, for now, just replace all with the new one(s). This would open up for a lot of situations in other processes.

packages/common/src/services/decision/syncProposalProfileLocation.ts
```diff
@@ -0,0 +1,86 @@
...
+ * the join in place.
+ *
+ * Never throws on a bad value — a malformed location degrades to "no link"
+ * instead of failing the surrounding write. Must run inside the same
+ * transaction as the proposal write (`tx`).
+ */
+export async function syncProposalProfileLocation(
+  tx: DbClient,
+  proposalProfileId: string,
+  proposalData: unknown,
```
scazan: We should definitely type this one (generally we keep the type assertions as close to the DB call as possible)

packages/common/src/services/decision/validateProposalAgainstTemplate.ts
scazan: Seems mostly like a stylistic update. looks good to me though

services/api/src/routers/decision/resolveBoundary.ts
```diff
@@ -0,0 +1,33 @@
...
+  name: z.string(),
+  taxonomyTermId: z.string().nullable(),
+});
+
+const resolveBoundaryOutputSchema = z.object({
+  boundary: resolvedBoundaryEncoder.nullable(),
+});
+
+export const resolveBoundaryRouter = router({
+  resolveBoundary: networkAuthenticatedProcedure()
```
scazan: Small update. This should use `authenticatedProcedure` instead. This is some new stuff pretty unique to our codebase right now.

services/api/src/routers/taxonomy/index.ts
scazan: yeah, this makes sense for now. We can always move it.

services/api/src/routers/taxonomy/reverseGeocode.ts
```diff
@@ -0,0 +1,102 @@
...
+    )
+    .output(
+      z.object({
+        geoname: ReverseGeoNameSchema.nullable(),
+      }),
+    )
+    .query(async ({ input }) => {
+      const { lat, lng } = input;
+
+      const geoname = await cache({
```
scazan: perfect with the caching here!

services/api/src/routers/taxonomy/reverseGeocode.ts
```diff
@@ -0,0 +1,102 @@
...
+  address: z.string(),
+  lat: z.number(),
+  lng: z.number(),
+  countryCode: z.string().optional(),
+  countryName: z.string().optional(),
+});
+
+type ReverseGeoName = z.infer<typeof ReverseGeoNameSchema>;
+
+const reverseGeocodePoint = async ({
```
scazan: This should be moved to the packages/common/src/services. We are keeping routers slim and database-less.

.env.docker
```diff
@@ -42,3 +42,7 @@ SSR_SECRETS_KEY={"alg":"A256CBC","ext":true,"k":"xUWFSEEnQPIlAkPqTxnIsZz5VRacNPE
 
 # S3 asset root — point at Supabase Storage for local dev
 S3_ASSET_ROOT=http://localhost:3121/storage/v1/object/public/assets
+
+# ── Maps ─────────────────────────────────────────────────────────────────────
+# MapTiler client key for map tiles/styles. Publishable client key — safe to commit.
+NEXT_PUBLIC_MAPTILER_API_KEY=uNCePYY380rbdWbgIYS7
```
scazan: We probably want to copy this over instead from the local environment. If others are hosting Common it needs to avoid using our key here.

apps/app/src/components/collaboration/CollaborativeDropdownField.tsx
scazan: Could we just have this done on update and submit? We can avoid doing the district auto-selection on the client end entirely maybe.

scazan (commented):
Great work! left some comments mostly to help guide some things.

apps/app/src/components/decisions/location/LocationMapField.tsx
bloudermilk: @scazan currently we show the raw latitude/longitude values if we can't reverse geocode. This is actually related to the question I asked in my own review. Do we want to expose latitude/longitude to the user? Or prevent people from tagging locations that can't be reverse geocoded to an address?

apps/app/src/components/decisions/location/LocationMapField.tsx
bloudermilk: @scazan IMO the UX as-designed (with the error message) is more intuitive. Interfering with the ability to place a marker or filtering the addresses that show up in search could be confusing. The way it is now, users are free to enter any address or place a marker anywhere then they get clear feedback on whether the location is valid or not. 

Judging by the complex boundary shapes of these districts, I wouldn't be surprised if this ends up being an education moment for some citizens. There are a bunch of gaps in the middle that you might expect to be in the city which aren't.

apps/app/src/components/decisions/location/mapConfig.ts
bloudermilk: @scazan FYI we do support configuring the default map location for each process in the admin. This location is what the admin would see when they first load up a fresh template. It's zoomed out to the whole globe, so centering on Columbus ends up not being noticeable. Agreed a future improvement would be to geolocate the admin 💯

Would you like me to track that as a TODO or an Asana ticket?

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/TemplateEditorContent.tsx
bloudermilk: This is the mechanism to prevent admins from adding multiple Location fields to a process. This behavior wasn't specified and supporting multiple locations added a complexity that didn't qualify as MVE in my opinion.

.env.docker
bloudermilk: Will move to `.env.local` / `.env`

apps/app/src/components/collaboration/CollaborativeDropdownField.tsx
bloudermilk: @scazan let's give it a shot and see how it feels. The UX to get right is the immediate district feedback in the map component, so the participant knows whether or not their location is valid.

I'll try to apply auto-tagging on save server-side. This should be fine as long as the client updates to show the added category. Otherwise it could lead to the client overwriting the tagged category.

I'll do this in its own commit so it's easy to roll back.

apps/app/src/components/decisions/location/LocationMapField.tsx
scazan: ah great. no I think that raw lat/lng makes sense.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/FieldCard.tsx
bloudermilk: @scazan sorry, I'm interpreting these two statements as contradictory:

> We can keep this optionally required. They'll want to keep it required when they setup their instance.

Seeing two possible pathways:
A) My original interpretation of the spec: admins always *require* a location to be provided if a location field is present. Hence showing the toggle and enforcing that it stays on.
B) We could allow that Location itself is optional, while *required* logic is enforced exclusively at the district category level.

apps/app/src/components/decisions/proposalTemplate.ts
bloudermilk: @scazan if you want I can spend some time on this today but it's definitely additional scope and potentially taking time away from other features.

Could you clarify this statement?
> Where we do they "sync" we can, for now, just replace all with the new one(s).

apps/app/src/components/decisions/proposalEditor/ReadonlyProposalFields.tsx
bloudermilk: @scazan I will improve the preview mode, time-permitting!

https://app.asana.com/1/1108348036672214/project/1215095416548888/task/1215823399823188

apps/app/src/components/collaboration/CollaborativeDropdownField.tsx
scazan: ahh I see. Well let's go with the MVE generally and we can iterate as needed.

apps/app/src/components/decisions/location/mapConfig.ts
scazan: Definitely just a TODO. but added this mostly for myself :)

apps/app/src/components/decisions/proposalTemplate.ts
bloudermilk: Keeping this as-is for MVE given the scope ambiguity of having this field be optional or multiple.

bloudermilk (commented):
@scazan I fixed most of the smaller changes. Added one Asana story for an admin-facing improvement, one TODO you requested, and deferring some higher-lift requests for now.

Reminder that you will need to generate and configure the `NEXT_PUBLIC_MAPTILER_API_KEY` across CI/hosts.

apps/app/src/components/decisions/ProcessBuilder/stepContent/template/FieldCard.tsx
scazan: ah sorry, I meant that in the app you would toggle required on/off but that the first user will be toggling that on by chance.

In that way this should require less work. just field.required . no special functionality here is what I meant including enforcing it at the district level.

you can definitely leave it though and we can iterate internally if you want (same with all of the comments). definitely dont want to pull time away from other things.

scazan (approved):

PR #1264, opened 2026-06-08
feat(db): moderation_flags table

scazan:
Async-moderation DB schema only: a polymorphic `moderation_records` audit table (subjectType+subjectId, status, provider link) plus a denormalized `moderationHiddenAt` flag on proposals/posts/users, kept separate from `visibility`. No behavior is wired yet — that lands in the stacked sync/provider/inngest PRs.

### DIFF
--- a/services/db/schema/tables/moderationFlags.sql.ts
+++ b/services/db/schema/tables/moderationFlags.sql.ts
@@ -1,0 +1,128 @@
+import { sql } from 'drizzle-orm';
+import {
+  index,
+  jsonb,
+  pgEnum,
+  pgTable,
+  text,
+  timestamp,
+  uniqueIndex,
+  uuid,
+} from 'drizzle-orm/pg-core';
+
+import {
+  autoId,
+  enumToPgEnum,
+  serviceRolePolicies,
+  timestamps,
+} from '../../helpers';
+import { profiles } from './profiles.sql';
+
+/** What kind of item a flag points at (comments are posts). */
+export enum ModerationItemType {
+  PROPOSAL = 'proposal',
+  POST = 'post',
+  USER = 'user',
+}
+
+/** Lifecycle of a single flag (distinct from the item's own status). */
+export enum ModerationFlagStatus {
+  FLAGGED = 'flagged', // open flag
+  CONFIRMED = 'confirmed', // admin confirmed the flag
+  DISMISSED = 'dismissed', // false positive
+  DISPUTED = 'disputed', // owner contested; awaiting admin
+}
+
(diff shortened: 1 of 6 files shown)


services/db/schema/tables/moderationFlags.sql.ts
```diff
@@ -0,0 +1,114 @@
...
+ * The flagged subject is a profile (`profileId` FK), with `entityType` as the
+ * shared discriminator for the profile kind.
+ */
+export const moderationFlags = pgTable(
+  'moderation_flags',
+  {
+    id: autoId().primaryKey(),
+
+    entityType: entityTypeEnum('entity_type').notNull(),
+    profileId: uuid('profile_id')
```
scazan: This would mean that we need to have an associated profile on every moderatable entity. This means Posts now get a profileId. Luckily it's only for new posts atm. Up for discussion though. Alternatively, we could to a "type" here and have an enum for moderatable content types.

services/db/schema/tables/moderationFlags.sql.ts
```diff
@@ -0,0 +1,114 @@
...
+    index('moderation_flags_status_created_at_idx').on(
+      table.status,
+      table.createdAt,
+    ),
+    // flaggedBy FK support + "flags raised by X".
+    index('moderation_flags_flagged_by_idx').on(table.flaggedByProfileId),
+  ],
+);
+
+export type ModerationFlag = InferModel<typeof moderationFlags>;
```
valentin0h:
```suggestion

```

valentin0h: Use `$inferSelect` if needed.

services/db/schema/tables/moderationFlags.sql.ts
valentin0h: yeah making everything that should be moderated a profile won't scale I think. good candidate for polymorphic table as you mentioned.

valentin0h (approved):

services/db/schema/tables/moderationFlags.sql.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * The flagged subject is a profile (`profileId` FK), with `entityType` as the
+ * shared discriminator for the profile kind.
+ */
+export const moderationFlags = pgTable(
+  'moderation_flags',
+  {
+    id: autoId().primaryKey(),
+
+    entityType: entityTypeEnum('entity_type').notNull(),
```
valentin0h: What's the need for having this?

PR #1265, opened 2026-06-08
feat(moderation): synchronous text moderation

scazan:
Synchronous text-moderation gate that scores submitted posts/comments and blocks disallowed content before any write, backed by a pluggable provider (Hive/Lasso/Checkstep) selected via `MODERATION_PROVIDER`. Stacked on the moderation_flags schema PR (#1264).

This uses multiple providers for us to test with but ultimately we will settle on a single provider. This keeps it easily swappable which could be useful for users who self-host.


packages/common/src/services/posts/createPost.ts
```diff
@@ -267,6 +268,9 @@ export const createPost = async (input: CreatePostServiceInput) => {
     },
   });
 
+  // Moderation gate: block disallowed text before any row is written.
+  await assertTextContentModerated(content);
```
valentin0h: Feels like we should move this as high as possible in the function, as it's independent of all other logic? 

Sth like

```ts
const assertTextContentModeratedPormise = assertTextContentModerated();

... logic ...

await assertTextContentModeratedPormise; // before any writes
```

valentin0h (approved):
Noice 👍

packages/common/src/services/posts/createPost.ts
scazan: my only concern there is that it'll be a publicly accessible endpoint with cost implications since it's an external provider. We make sure you can actually do it before we hit any external provider. It would decrease latency to do this in parallel though. I'll move it higher for now and rethink after because we might actually have to move this entirely out of this funciton in the end (waiting on lawyers to respond about whether we can do this)

valentin0h: Yeah I ideally this shouldn't be "sync".

scazan: I don't know.. it's good for not allowing someone to submit in the first place. Good way for a hard block on certain red-lines.

PR #1267, opened 2026-06-08
feat(moderation): Hive/Lasso/Checkstep adapters behind MODERATION_PROVIDER

scazan:
Replaces the placeholder HTTP provider with three real vendor adapters (Hive, Lasso, Checkstep) selected by a single MODERATION_PROVIDER env var, with the optional submitForReview record link for the two platform vendors. Stacked on the sync-gate PR.


scazan:
Folded into #1265 — the sync gate and the Hive/Lasso/Checkstep adapters ship as one feature, since the gate isn't functional without a real provider.

PR #1269, opened 2026-06-08
feat(moderation): Moderation flow — automated review, Inngest, webhook + notify

scazan:
> **Automated moderation — base of the stack.** User-initiated reporting (`flagItem`/`submitUserFlag`, the access gate, and the `moderation.flagItem` tRPC route) is split into the stacked **#1320**; the proposal Report UI + flagged indicators into **#1314** (on top of #1320); the read-side hiding + `isFlagged` indicators merged earlier in **#1296**. This PR is the automated server-side async flag pipeline.

Adds the asynchronous moderation path on top of the provider stack: `content/submitted` + `content/flagged` events, the admin-resolution (`resolveModeration`) service, the unit-tested `reviewContentAsync` orchestrator wired into a `moderateContentAsync` Inngest function, webhook-based provider verdicts with per-task/per-round accounting, and an author notification email. The flags this produces are what #1296 (merged) hides from reads. Flag administration happens in the provider's application, so there is no in-platform review queue: the provider webhook **is** the resolution path (overturns dismiss, upholds keep hiding), and `resolveModeration` remains as the escape hatch for flags no webhook will resolve.

## Components

| Layer | Piece | Role |
|---|---|---|
| `services/db` | `moderation_flags` | One open flag per item (partial unique index); lifecycle `pending → flagged → confirmed/dismissed` |
| `services/db` | `moderation_submissions` | One row per provider task per round; verdicts aggregate across an item's tasks |
| `packages/common` | `reviewContentAsync` | Automated review: plan round → record → submit |
| `packages/common` | `applyModerationVerdict` | Pure decision flow over the task aggregate (DI, unit-tested) |
| `packages/common` | `providers/{hive,lasso,checkstep}` | Vendor adapters: sync scoring, async submit, webhook parse + HMAC verify |
| `services/api` | webhook request handler | Webhook entry point (the `moderation.flagItem` tRPC route is in #1320) |
| `apps/api` | `POST /api/v1/moderation/webhooks` | Raw-body webhook route |
| `services/workflows` | `moderateContentAsync`, `sendContentFlaggedNotification` | Inngest functions |

## Flow 1 — automated review on content write

The full item (text + media) is reviewed **asynchronously** — there is no synchronous gate on the write path; the post/submit always succeeds and a `content/submitted` event kicks off provider review. Only a provider verdict hides the item (a passed/`flagged`/`confirmed` verdict, never the write itself). For proposals every prose field is resolved via `extractProposalText` — templates define arbitrary text fields — against the template-validated record, which carries the latest TipTap fragment text per field:

```mermaid
sequenceDiagram
    autonumber
    actor U as Author
    participant API as createPost / submitProposal
    participant W as Inngest: moderateContentAsync
    participant DB as moderation_submissions
    participant P as Provider (Hive / Lasso / Checkstep)

    U->>API: write post / submit proposal
    API--)W: emit content/submitted (no sync gate)
    W->>W: step "round-id": mint roundId (memoized across retries)
    W->>W: step "resolve-media": sign attachment URLs
    W->>DB: recordSubmissionRound(roundId, one pending row per task)
    W->>P: submitForReview(refs embed roundId)
    P-->>W: provider task ids
    Note over P: human / model review happens asynchronously
```

The round is recorded **before** the provider is called, so a verdict can never arrive for a task we haven't recorded. Because the Inngest step memoizes `roundId`, a retry after a partial submit failure reuses the same round and a verdict that already landed is preserved (`onConflictDoNothing` on re-insert).

## Flow 2 — user report

The user-initiated report path (`moderation.flagItem` → `submitUserFlag` →
`assertModerationItemAccess` → pending flag → provider submit) is split into the
stacked **#1320**. It reuses the same stores, provider integration, and verdict
pipeline this PR introduces — a user report just creates the `pending` flag that
the webhook verdict (above) later confirms or clears.

## Flow 3 — provider verdict webhook

```mermaid
sequenceDiagram
    autonumber
    participant P as Provider
    participant RT as POST /api/v1/moderation/webhooks
    participant H as handleModerationWebhook
    participant S as recordSubmissionVerdict
    participant V as applyModerationVerdict
    participant N as Inngest: sendContentFlaggedNotification

    P->>RT: verdict callback (?secret=…, vendor signature headers)
    RT->>H: rawBody + headers + provided secret
    H->>H: gate 1 — shared secret, timing-safe (401)
    H->>H: gate 2 — vendor HMAC verifyWebhook (401)
    H->>H: parseWebhook → { itemType, itemId, roundId, mediaId, verdict }
    H->>V: recordModerationVerdict
    V->>S: UPDATE task row WHERE item + mediaId + roundId (rows locked FOR UPDATE)
    alt no matching row (forged ref / stale round / already resolved)
        S-->>V: null
        V-->>RT: noop → 200
    else task matched
        S-->>V: aggregate (flagged-if-any, clear-only-when-all-resolved)
        V->>V: decide: create / flag / dismiss / noop (compare-and-set)
        opt flag newly created or confirmed
            V--)N: emit content/flagged (race winner only)
            N->>N: email the author
        end
        V-->>RT: 200
    end
```

### Verdict decision logic

```mermaid
flowchart TD
    W[task verdict arrives] --> RT["record verdict:<br/>match itemType + itemId + mediaId + roundId"]
    RT -- no matching row --> N1["noop<br/>(forged / superseded round / resolved)"]
    RT -- aggregate --> AF{any task<br/>flagged?}
    AF -- yes --> OF{open flag<br/>for item?}
    OF -- none --> CR["create automated flag<br/>+ notify author"]
    OF -- "pending (user report)" --> MK["CAS pending→flagged<br/>+ notify author"]
    OF -- already flagged --> N2[noop — idempotent]
    AF -- no --> AR{all tasks resolved<br/>AND an open flag?}
    AR -- yes --> DM["CAS open→dismissed<br/>(incl. provider overturning its own flag)"]
    AR -- no --> N3["noop<br/>(tasks still pending, or nothing to clear)"]
```

A clean *task* never dismisses an item while another task is still pending or flagged. But a genuine all-clear aggregate dismisses any **open** flag — including one the provider previously flagged: an all-clear means the provider re-reviewed and cleared the content (a deliberate overturn, or a re-submitted edit that came back clean — editing flagged content always re-triggers review), and **the provider's latest review wins**. Local admin resolutions (`confirmed`/`dismissed`) are terminal — they aren't open — so they're never auto-overturned.

## Flag lifecycle

```mermaid
stateDiagram-v2
    [*] --> pending: user report
    [*] --> flagged: automated verdict (no prior flag)
    pending --> flagged: provider confirms (CAS)
    pending --> dismissed: every task clear (CAS)
    pending --> confirmed: admin upholds
    pending --> dismissed: admin dismisses
    flagged --> dismissed: provider overturns (all tasks clear)
    flagged --> confirmed: admin upholds
    flagged --> dismissed: admin dismisses
    confirmed --> [*]
    dismissed --> [*]
```

`resolveModeration` (admin) guards on the flag still being open and **clears the item's submission rows in the same transaction** — a redelivered provider webhook then finds no matching task and no-ops, so it can't resurrect a resolved flag or re-email the author. (`disputed` exists in the enum for the owner-dispute follow-up; nothing transitions into it yet.)

## Read filtering (enforcement)

The read side — hiding flagged items from general reads and the `isFlagged`/"Flagged" indicators — merged earlier in **#1296**. This PR only *produces* the flags (`flagged`/`confirmed`) that #1296 acts on. The proposal Report action + flagged UI are split into the stacked **#1314**.

## Round correlation (content refs)

Every ref sent to a provider is `itemType:itemId:roundId[:mediaId]`, e.g.

```
post:9f3c…d2:4a17…9b        ← the post's text task
post:9f3c…d2:4a17…9b:0      ← its first attachment (Hive fans out per media)
```

All three vendors echo the ref back on the webhook, so each verdict carries the exact round it answers — verdicts for a superseded round (content edited and re-submitted) or for tasks we never recorded are dropped. `itemId` and `roundId` are validated as UUIDs before any DB access.

## Webhook security model

| Gate | Applies to | Mechanism |
|---|---|---|
| Shared secret in callback URL | all vendors | timing-safe compare, fails closed when unconfigured |
| Vendor HMAC signature | Checkstep, Lasso | Checkstep: `x-auth-signature/date/nonce`, HMAC-SHA256 over SHA-256(`body.date.nonce`), key-rotation tolerant, ±10 min replay window. Lasso: `X-Lasso-Signature` = `sha256=` + base64 HMAC of raw body. Enabled via `MODERATION_WEBHOOK_SIGNING_KEY` |
| Round-matched, update-only recording | all vendors (Hive's only second factor — it documents no callback signing) | a verdict must hit a recorded `(item, task, round)` row; anything else no-ops |

## Concurrency & failure guarantees

| Scenario | Outcome | Mechanism |
|---|---|---|
| Webhook arrives before/while round is written | impossible — round is recorded before the provider knows the refs | write-first ordering |
| Provider submit fails mid-flow | flag + round rolled back, user retry starts clean | `flagItem` rollback |
| Inngest retry after partial submit | landed verdicts preserved | memoized roundId + `onConflictDoNothing` |
| Two users report the same item concurrently | one flag, one provider round | partial unique index + `created` flag from insert |
| Two task verdicts processed concurrently | serialized; final `clear` reliably dismisses | `SELECT … FOR UPDATE` per item |
| Webhook races admin resolution | admin decision stands, no duplicate email | CAS `WHERE status='pending'` |
| Webhook redelivered after resolution | noop | submissions cleared atomically on resolve |
| Forged payload with valid secret | cannot create/clear flags for unsubmitted content | round-matched update-only recording |

## Configuration

| Env var | Purpose |
|---|---|
| `MODERATION_PROVIDER` | `hive` \| `lasso` \| `checkstep` (feature off when unset) |
| `MODERATION_API_KEY` | the active vendor's API key |
| `MODERATION_WEBHOOK_SECRET` | shared secret embedded in the callback URL; async path is off without it |
| `MODERATION_WEBHOOK_SIGNING_KEY` | vendor webhook signing key (Checkstep platform key / Lasso dashboard secret); enables HMAC verification |

## Testing

Unit tests across the moderation services: the full verdict decision matrix (out-of-order, duplicate, and stale webhooks), the pure aggregation (`computeSubmissionAggregate`, extracted DB-free), round-before-submit ordering and rollback contracts, content-ref round-tripping, provider wire formats, HMAC verification (signed/tampered/rotated/replayed), and the flag-access gate (`assertModerationItemAccess`: HIDDEN proposals and already-flagged posts gated to owner/admin). The concurrency primitives (`FOR UPDATE`, the round-replace predicate) stay covered by reasoning + the unit-tested transition layer; the read-filtering SQL is exercised end-to-end by the DB-backed visibility test in #1296.

## Follow-ups (out of scope)

- **CSAM → immediate media removal:** add a dedicated `csam` `ModerationCategory`, mapped per provider (Hive's CSAM class, Checkstep/Lasso policy code); on a CSAM verdict, synchronously delete the storage objects **and** attachment rows (reuse `deleteResourceObject`) while still creating/keeping the flag and hiding the item, with the reason recorded.


packages/common/src/services/decision/submitProposal.ts
```diff
@@ -99,6 +100,17 @@ export const submitProposal = async ({
     );
   }
 
+  // Synchronous moderation gate: block disallowed proposal text on submit (the
+  // async pass below re-checks the text plus any attachments via the provider).
+  const proposalContent = existingProposal.proposalData as {
+    title?: string;
+    description?: string;
+  } | null;
+  const proposalText = [proposalContent?.title, proposalContent?.description]
```
scazan: in proposalData, we might have any number of text fields. We should parse all text-related fields in the template to examine

packages/common/src/services/moderation/contentRef.ts
```diff
@@ -0,0 +1,42 @@
...
+ * item without storing a pending DB row. `mediaId` distinguishes the per-
+ * attachment submissions of one item (text + N media each get their own id).
+ */
+export const encodeContentRef = (
+  itemType: ModerationItemType,
+  itemId: string,
+  mediaId?: string,
+): string =>
+  mediaId
+    ? `${itemType}${SEP}${itemId}${SEP}${mediaId}`
```
scazan: Couldn't we make this simpler with an array and a join(SEP)?

packages/common/src/services/decision/extractProposalText.ts
```diff
@@ -0,0 +1,35 @@
...
+ * as moderatable text — proposal templates can define any number of custom
+ * text fields, so an allowlist of known field names would silently skip them.
+ */
+const NON_TEXT_KEYS: ReadonlySet<string> = new Set([
+  'category',
+  'budget',
+  'attachmentIds',
+  'collaborationDocId',
+  'collaborationDocVersionId',
+]);
```
valentin0h, line 13: We shouldn't moderate `proposalData`; it should be purely meta info. If not, all content will always be in the yjs tiptap doc.

packages/common/src/services/decision/listAllProposals.ts
```diff
@@ -138,8 +180,13 @@ export const listAllProposals = async ({
...
       ),
       getSelectedProposalIds(processInstanceId),
-    ],
-  );
+      // Flagged items reach this point only for their creator or an admin —
+      // decorate them so the UI can render the "Flagged" indicator.
+      getActivelyFlaggedItemIds(
+        'proposal',
+        pageItems.map((proposal) => proposal.id),
+      ),
```
valentin0h, line 188: would be interesting to see how we can cache this at some point.

valentin0h: I'm thinking we don't need to be precise in what we fetch here, we can be more broad, so that we can fetch early and utilise caching better.

packages/common/src/services/decision/submitProposal.ts
```diff
@@ -91,6 +91,10 @@ export const submitProposal = async ({
   // created by the DB trigger links back to a concrete document revision.
   const parsed = parseProposalData(existingProposal.proposalData);
 
+  // Validate the submitted proposal against its template (throws on invalid).
+  // Content moderation is NOT a sync gate on submit: the `content/submitted`
+  // event emitted by the submit router drives async provider review, which
+  // hides the proposal if the verdict comes back disallowed.
```
valentin0h:
```suggestion
```

packages/common/src/services/decision/validateProposalAgainstTemplate.ts
```diff
@@ -13,14 +13,19 @@ import type { ProposalTemplateSchema } from './types';
...
+ *
  * @throws {ValidationError} when the proposal data does not satisfy the template schema
  * @throws {CommonError} when TipTap credentials are missing for a collab-doc proposal
  */
 export async function validateProposalAgainstTemplate(
   proposalTemplate: ProposalTemplateSchema,
   proposalData: unknown,
   title?: string,
-): Promise<void> {
+): Promise<Record<string, unknown>> {
```
valentin0h: We don't seem to be using the return value in this PR?

packages/common/src/services/decision/listAllProposals.ts
```diff
@@ -40,6 +62,10 @@ export const listAllProposals = async ({
     ? decodeCursor<{ value: string | Date }>(input.cursor)
     : undefined;
 
+  // The caller's own grants unioned with public grants — used by the
+  // moderation owner-exception subquery (proposal.profileId membership).
+  const accessUserIds = resolveAccessUserIds(user);
```
valentin0h, line 67: 👍

packages/common/src/services/moderation/providers/lasso.ts
```diff
@@ -1,15 +1,44 @@
...
+): boolean => {
+  const signature = headerValue(headers, 'x-lasso-signature');
+  if (!signature) {
+    return false;
+  }
+  const expected = `sha256=${createHmac('sha256', signingKey)
+    .update(rawBody, 'utf8')
+    .digest('base64')}`;
+  return timingSafeStringEqual(signature.trim(), expected);
+};
```
valentin0h, line 41: Nice, it's tested right?

packages/common/src/services/moderation/providers/lasso.ts
```diff
@@ -92,12 +189,63 @@ export const createLassoProvider = ({
...
-    return referenceFrom(result);
+    return { ...referenceFrom(result), submittedRefs: [ref] };
+  },
+
+  parseWebhook: ({ rawBody }: ModerationWebhookInput): ModerationVerdict[] => {
+    const payload = lassoWebhookSchema.parse(JSON.parse(rawBody));
+    const verdicts: ModerationVerdict[] = [];
+    for (const action of payload.actions) {
+      if (action.action_type !== 'ChangeStatus' || action.type !== 'content') {
+        continue;
```
valentin0h, line 223: Are we expecting something else? If not, we should log ideally.

packages/common/src/services/moderation/contentRef.ts
```diff
@@ -0,0 +1,68 @@
...
+  itemType: ModerationItemType;
+  itemId: string;
+  /**
+   * The submission round this ref belongs to. A re-submission (after an edit,
+   * or a fresh report) mints a new round id, so a delayed webhook for a
+   * superseded round can't be mistaken for a verdict on the current one.
+   */
+  roundId: string;
+  /** Set when the submission is a single attachment of a multi-part item. */
+  mediaId?: string;
```
valentin0h, line 30: Feels like it should be its own type at first look. Just noting but there might be good reason to have it here..

packages/common/src/services/moderation/resolveModerationMedia.ts
```diff
@@ -0,0 +1,162 @@
...
+ * prose fields. Users carry no moderatable text here.
+ *
+ * For collab-doc proposals the prose lives in the TipTap document, not in
+ * `proposalData` — reading only the stored record would submit an empty
+ * string to the provider. So the document's template fragments are fetched
+ * (as plain text) and merged over the stored record, mirroring how
+ * `validateProposalAgainstTemplate` assembles its validation data. A TipTap
+ * failure falls back to the stored record rather than blocking moderation.
+ */
+export const resolveModerationItemText = async (
```
valentin0h: Should this be in this file? It looks like media-specific.

packages/common/src/services/moderation/resolveModerationMedia.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (itemType === 'proposal') {
+    const proposal = await db.query.proposals.findFirst({
+      where: { id: itemId },
+      with: { processInstance: true },
+    });
+    if (!proposal) {
+      return '';
+    }
+
+    const stored = recordOf(proposal.proposalData);
```
valentin0h: See other comment.
```suggestion
```

packages/common/src/services/moderation/resolveModerationMedia.ts
```diff
@@ -0,0 +1,162 @@
...
+      // Legacy single-fragment docs (no template) store everything under
+      // `default`; template-driven docs have one fragment per field.
+      const fragmentNames = template
+        ? getProposalFragmentNames(template)
+        : ['default'];
+      const fragmentTexts = await getTipTapClient().getDocumentFragments(
+        parsed.collaborationDocId,
+        fragmentNames,
+        { format: 'text' },
+      );
```
valentin0h: Worth looking into if we're missing moderating links/images when using text. We might not.

valentin0h (approved):
Looking good 👍

packages/common/src/services/moderation/providers/lasso.ts
scazan: yep!

PR #1274, opened 2026-06-09
feat(access): profile-scoped permission rows on access roles

valentin0h:
Lets a single global role (e.g. a future Public role) carry different permissions per profile/decision instance, so per-process capabilities can be configured without minting a role per instance. A profile-scoped permission row overrides the role's global row for that profile only; everywhere else only global rows apply, so scoped grants can never leak across instances.

### DIFF
--- a/packages/common/src/services/access/getRoles.ts
+++ b/packages/common/src/services/access/getRoles.ts
@@ -6,6 +6,8 @@ import {
   desc,
   eq,
   gt,
+  inArray,
+  isNotNull,
   isNull,
   lt,
   or,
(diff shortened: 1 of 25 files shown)


services/db/migrations/20260609193913_acoustic_mulholland_black/migration.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE "access_role_permissions_on_access_zones" ADD COLUMN "profile_id" uuid;--> statement-breakpoint
+-- Dedupe before adding the unique constraint: no unique constraint existed on
+-- (access_role_id, access_zone_id) before, so duplicates are representable.
+-- Keep the most recently updated row per (role, zone), tie-broken by id.
+DELETE FROM "access_role_permissions_on_access_zones" a
```
scazan: This feels unnecesary as we don't have dupes currently anywhere.

packages/common/src/services/decision/duplicateInstance.ts
```diff
@@ -254,13 +254,16 @@ async function copyCustomRoles({
...
+    // Copy zone permissions. Rows scoped to the source profile follow the
+    // role to the target profile; global rows (NULL) stay global.
     if (role.zonePermissions.length > 0) {
       await db.insert(accessRolePermissionsOnAccessZones).values(
         role.zonePermissions.map((zp) => ({
           accessRoleId: newRole.id,
           accessZoneId: zp.accessZoneId,
           permission: zp.permission,
+          profileId:
+            zp.profileId === sourceProfileId ? targetProfileId : zp.profileId,
```
scazan: This is interesting.. I wonder why we would need to set zp.profileId on the new role if it isn't null and/or isn't === sourceProfileId. Feels like a potential error.

packages/common/src/services/access/getRoles.ts
```diff
@@ -104,6 +104,10 @@ export const getRoles = async (params?: {
               .from(accessZones)
               .where(eq(accessZones.name, zoneName)),
           ),
+          // Only the role's global permission row: per-profile override rows
+          // would multiply join rows (breaking limit-based pagination) and are
+          // read via the effective resolvers (e.g. getDecisionRole) instead.
+          isNull(accessRolePermissionsOnAccessZones.profileId),
```
scazan, line 141: The global profile rows are what are available to the invite user modal where we invite users to an organization with a specific role. One important thing to consider is that the global public role will end up in that list and won't make much sense since everyone gets the public role permissions by default. It would be a "invite user with public permissions" which technically will work but is odd. So we still need to treat the public role as a special case throughout (probably the only one or maybe out of 2 total special case roles I'd imagine, public and admin).

scazan (commented):
I don't want to block the migration, however the app logic will expose the public role here to more places then we want. Luckily, we are not currently exposing the global roles to the decision process role management, which means this won't have an effect on that. But the moment we do expose that, I want to flag that both the admin and member role that are both global will appear there suddenly, which we don't want. 
If we agree on the approach then I think the migration should merge and we can even merge the app logic if we address the global public role in a separate PR before we actually add one. As it is, this will open up a potential issue that definitely needs to be addressed before we make a move on the global public role.

services/db/migrations/20260609193913_acoustic_mulholland_black/migration.sql
valentin0h: yeah, was just meant to ensure the migration wont fail but I'll remove.

packages/common/src/services/access/getRoles.ts
```diff
@@ -31,12 +34,37 @@ interface Role {
...
+ * roles are granted only by their dedicated services (e.g. the
+ * public-participation flow), never by hand. Accepts the callback's table arg
+ * since db.query aliases tables.
+ */
+export const assignableRoleFilter = (accessRoleCols: typeof accessRoles): SQL =>
+  or(
+    isNotNull(accessRoleCols.profileId),
+    inArray(accessRoleCols.id, EXPOSABLE_GLOBAL_ROLE_IDS),
+  )!;
+
```
valentin0h, line 60: I'll revisit this.

packages/common/src/services/access/getRoles.ts
```diff
@@ -15,6 +17,7 @@ import {
   accessRoles,
   accessZones,
 } from '@op/db/schema';
+import { ROLES } from '@op/db/seedData/accessControl';
```
scazan: This is coming in from the seed data so won't work

scazan (approved):

PR #1287, opened 2026-06-11
feat(decisions): split process builder overview into Process Settings and Overview tabs

nourmalaeb:
Stacked on #1286.

- Sidebar gains a "Process Settings" section (the old Overview form, renamed) ahead of a new "Overview" tab
- Overview tab: headline + short description inputs above a divider, rich text body editor with bubble menu, autosaved like other sections
- New `overview` field ({ headline, description, content }) on `instanceData` through the encoder, `InstanceOverview` type, and `updateDecisionInstance` merge — JSONB, no migration
- Content-error guard: malformed stored body content is logged and shown as an error state instead of being silently replaced (and autosave is suppressed so it can't overwrite server data)
- Integration tests for overview round-trip, sibling preservation, and clearing; e2e specs updated for the new tab structure

---
  - https://app.asana.com/0/0/1215166793963399


apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -26,8 +52,200 @@ export default function OverviewSection(props: SectionProps) {
...
+  };
+
+  const extensions = useMemo(
+    () => [
+      StarterKit.configure({
+        link: { openOnClick: false },
+      }),
+      Placeholder.configure({
+        placeholder: t(
+          "Write what residents need to know about this process — its goals, timeline, who's running it, how to participate.",
```
scazan: Should we make this a bit more generic. It's gonna work for this process... BUT.. we will forget about it being here when we have self-serve which is coming shortly after Columbus. Could we either make it configurable or make it generic with the normal placeholder?

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -26,8 +52,200 @@ export default function OverviewSection(props: SectionProps) {
...
+      console.error('Failed to load overview content', { instanceId, error });
+      setContentError(true);
+    },
+    editorProps: {
+      attributes: {
+        class: cn(
+          baseEditorStyles,
+          'min-h-40',
+          // Placeholder decoration (Placeholder extension)
+          '[&_.is-empty]:before:pointer-events-none [&_.is-empty]:before:float-start [&_.is-empty]:before:h-0 [&_.is-empty]:before:text-neutral-gray3 [&_.is-empty]:before:content-[attr(data-placeholder)]',
```
scazan: Can we use the standard RichTextEditor for this that has Tiptap and the Common styles builtin already? If it's missing anything maybe we build more into that one. This way we can keep it very consistent style-wise as well

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -1,13 +1,39 @@
...
+import { useProcessBuilderStore } from '@/components/decisions/ProcessBuilder/stores/useProcessBuilderStore';
+
+/**
+ * TipTap JSON document holding the overview body. Headline and description
+ * are plain strings (separate inputs), so other surfaces can render them
+ * without parsing a TipTap document — only the body needs a rich viewer.
+ */
+interface OverviewBodyDoc {
+  type: 'doc';
+  content?: Record<string, unknown>[];
```
scazan: Could we type this as JSONContent[]?

packages/common/src/services/decision/schemas/instanceData.ts
```diff
@@ -43,12 +43,24 @@ export interface PhaseInstanceData {
...
 }
 
+/** Public-facing overview content (headline, short description, rich text body) */
+export interface InstanceOverview {
+  headline?: string;
+  description?: string;
+  /** Rich text body as a TipTap JSON document */
+  body?: {
+    type: 'doc';
+    content?: Record<string, unknown>[];
```
scazan: Can this also be JSONContent[]?

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -26,8 +52,200 @@ export default function OverviewSection(props: SectionProps) {
...
+              setDescription(e.target.value);
+              saveOverview({ description: e.target.value });
+            }}
+            placeholder={t(
+              'Add a short description — one or two lines that sit under the headline.',
+            )}
+            aria-label={t(
+              'Add a short description — one or two lines that sit under the headline.',
+            )}
+            className="w-full bg-transparent text-base text-neutral-black placeholder:text-neutral-gray3 focus:outline-none"
```
scazan: Should this be a component at this point that we can re-use?

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -26,8 +52,200 @@ export default function OverviewSection(props: SectionProps) {
...
+            type="text"
+            value={headline}
+            maxLength={HEADLINE_MAX_LENGTH}
+            onChange={(e) => {
+              setHeadline(e.target.value);
+              saveOverview({ headline: e.target.value });
+            }}
+            placeholder={t('Add a headline')}
+            aria-label={t('Add a headline')}
+            className="w-full bg-transparent font-serif text-title-lg text-neutral-black placeholder:text-neutral-gray3 focus:outline-none"
```
scazan: We should match text-neutral-charcoal like we have in the proposal template builder (instead of text-neutral-black here). We can try to just use the same components to avoid those issues.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
scazan: Another slight issue with this is that for each block we are getting the placeholder. You can see here that there are two texts added and the placeholder persists on every new block:

Image

Image

scazan (commented):
I think the main thing is we should pull out the components for the editor so we can keep them matched with the proposal template styles

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Using the shared `RichTextEditor` component means that this is now stored as an HTML string.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Made the text more generic, and only appear when the editor is empty.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Good suggestion, done. This also means that the data is stored as HTML, resolving the type issues we had.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Those are collaborative Tiptap components, we don't need the complexity here. Extracted these to a simple shared component.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Yes, done

packages/common/src/services/decision/schemas/instanceData.ts
nourmalaeb: Resolved by switching to shared `RichTextEditor` component which stores data as an HTML string

scazan: nice, thanks. We should probably address the HTML string soon as well (not in this PR). It will be better to store this as JSON content in the end.

scazan (approved):

PR #1288, opened 2026-06-11
feat(decisions): render overview page hero and about body

nourmalaeb:
Renders the decision Overview tab (`/decisions/[slug]/overview`), behind the server-side `decision_overview` flag.

- **Overview page**: gradient hero (headline, optional subhead, Browse/Submit CTAs), a placeholder slot for the phases timeline, and an "About the process" body. Submit CTA is gated on phase rules + submit access (same gate as `StandardDecisionPage`). Content is mocked (`decisionOverviewMock.ts`) shaped like a future `instanceData.overview`; falls back to `instance.name` / `instance.description`.
- **DecisionHeader refactor**: header is now header-only (bar + stepper) instead of wrapping the page body. The results-view gradient moved into `ResultsPage`'s own hero, and `DecisionTranslationProvider` moved up to the composition root of each route (the `(decision-view)` layout + the two legacy routes) so it still encloses the now-sibling header and body.
- **Mobile**: route-scoped top clearance on the current-phase tab so the floating Overview/Current toggle doesn't block content; the legacy decision route is untouched.

Stacked PR #1291 wires the Submit CTA to create a draft proposal.

### DIFF
--- a/apps/app/src/components/decisions/DecisionHeader.tsx
+++ b/apps/app/src/components/decisions/DecisionHeader.tsx
@@ -4,19 +4,15 @@ import { useTrackPageView } from '@/hooks/useTrackPageView';
 import { getDecisionCommonProperties } from '@op/analytics/client-utils';
 import { trpc } from '@op/api/client';
 import { type ProcessPhase } from '@op/api/encoders';
-import { isLastPhase } from '@op/common/client';
-import { cn } from '@op/ui/utils';
 import { type ReactNode } from 'react';
 
 import { useTranslations } from '@/lib/i18n';
 
 import { DecisionInstanceHeader } from '@/components/decisions/DecisionInstanceHeader';
 import { DecisionStepperBar } from '@/components/decisions/DecisionStepperBar';
-import { DecisionTranslationProvider } from '@/components/decisions/DecisionTranslationContext';
 
 interface DecisionHeaderProps {
   instanceId: string;
-  children?: ReactNode;
   /** Decision profile slug for building the edit link */
   decisionSlug?: string;
   /** Whether the current user has admin access to this decision */
@@ -35,6 +31,12 @@ interface DecisionHeaderProps {
   showStepper?: boolean;
 }
 
+/**
+ * Header bar + optional phase stepper for a decision. Render inside a
+ * DecisionTranslationProvider — the stepper relies on it for phase-name
+ * translations. Backgrounds are owned by the page (e.g. the results hero),
+ * not the header.
+ */
 export function DecisionHeader(props: DecisionHeaderProps) {
   const { instanceId } = props;
 
@@ -52,7 +54,6 @@ export function DecisionHeader(props: DecisionHeaderProps) {
 
 function DecisionHeaderContent({
   instanceId,
-  children,
   decisionSlug,
   isAdmin,
   canReadUpdates,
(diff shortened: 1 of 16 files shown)


apps/app/src/components/decisions/decisionOverviewMock.ts
nourmalaeb, line 1: Added mock data in a separate file for easy deletion later -- we'll throw away the data and the `DecisionOverviewData` type.

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -1,35 +1,192 @@
...
+  const headline = overview.headline ?? instance.name;
+
+  // Same gate as StandardDecisionPage: the phase must accept proposals and
+  // the viewer must have submit access.
+  const currentPhase = instance.instanceData?.phases?.find(
+    (p) => p.phaseId === instance.currentStateId,
+  );
+  const canSubmitProposal =
+    currentPhase?.rules?.proposals?.submit === true &&
+    instance.access?.submitProposals === true;
```
valentin0h, line 76: I think we should have this under `instance.access.submitProposals`. If not, maybe we need to swap

`trpc.decision.getInstance.useSuspenseQuery({ instanceId });` with `getInstanceBySlug`

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -1,35 +1,192 @@
...
+          </ButtonLink>
+          {canSubmitProposal ? (
+            <ButtonLink
+              color="primary"
+              href={currentPhaseHref}
+              className="w-auto"
+            >
+              {t('Submit a proposal')}
+            </ButtonLink>
+          ) : null}
```
valentin0h, line 156: same link?

apps/app/src/components/decisions/DecisionHeader.tsx
valentin0h, line 1: nice 💯

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -1,35 +1,192 @@
...
+
+  if (!html && !fallbackText) {
+    return null;
+  }
+
+  return (
+    <section className="flex flex-col gap-4">
+      <Header2 className="font-serif">{t('About the process')}</Header2>
+      {html ? (
+        <ProposalHtmlContent html={html} />
```
valentin0h, line 182: Nice, can you also make the other callers have this format like here, `html ? <ProposalHtmlContent .../> : null (or whatever)`

valentin0h: Maybe we rename `ProposalHtmlContent` to something more generic, like `HtmlContentRenderer`

valentin0h (approved):

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: will update the link in #1291

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: Yeah we should! I'll do that in a separate PR.

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: no swap needed — `getInstance` already returns `access.submitProposals` (same query `StandardDecisionPage` gates on); fwiw the slug lookup is `getDecisionBySlug`, there's no `getInstanceBySlug`.

PR #1289, opened 2026-06-11
feat(api): open account.getMyAccount to public and anonymous callers

valentin0h:
Lets public pages resolve the viewer without rejecting: no-session callers get null, while anonymous sign-ins resolve their real (email-less) account created by the signup trigger. Keeps a single useUser() hook with a nullable user instead of a separate useOptionalUser, so there is one way to read the viewer everywhere. Ships independently — anonymous sessions and public routes stay gated until the follow-up PR.

feat(api): convert account.getMyAccount to an openProcedure with nullable output
refactor(app): make useUser()'s user nullable and null-safe across call sites
test(api): gating matrix asserts admission only; resolution-by-tier covered separately

### DIFF
--- a/apps/app/src/utils/getUser.ts
+++ b/apps/app/src/utils/getUser.ts
@@ -1,10 +1,26 @@
 import { createClient } from '@op/api/serverClient';
+import { redirect } from 'next/navigation';
 import { cache } from 'react';
 
 /**
- * Cached user fetch for server components.
+ * Cached user fetch for server components. Resolves null for public
+ * (no-session) visitors.
  */
 export const getUser = cache(async () => {
   const client = await createClient();
   return client.account.getMyAccount();
 });
+
+/**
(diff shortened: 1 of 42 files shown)


apps/app/src/utils/getUser.ts
```diff
@@ -1,10 +1,26 @@
...
+ */
+export const getRequiredUser = async () => {
+  const user = await getUser();
+
+  if (!user) {
+    redirect('/login');
+  }
+
+  return user;
+};
```
valentin0h, line 26: To be looked at in the next PR where we make the changes in the proxy

services/api/src/routers/account/getMyAccount.ts
valentin0h: @scazan feels like this block should go. Wdyt? I'm thinking we have the trigger, is there some other case this is covering?

scazan (approved):

PR #1291, opened 2026-06-11
feat(decisions): overview submit CTA creates a draft proposal directly

nourmalaeb:
Stacked on #1288.

- extracts a shared `useCreateProposal` hook from `DecisionActionBar` (creates an empty draft via `decision.createProposal`, redirects to its edit page, pending state + error toast)
- the overview hero's "Submit a proposal" button now creates the draft and lands on the proposal editor instead of linking to the Current Phase view
- `DecisionActionBar` reuses the hook (behavior unchanged)

### DIFF
--- a/apps/app/src/components/decisions/useCreateProposal.ts
+++ b/apps/app/src/components/decisions/useCreateProposal.ts
@@ -5,69 +5,67 @@ import { useUser } from '@/utils/UserProvider';
 import { trpc } from '@op/api/client';
 import { createSBBrowserClient } from '@op/supabase/client';
 import { toast } from '@op/ui/Toast';
-import { useState } from 'react';
+import { useTransition } from 'react';
 
 import { useRouter, useTranslations } from '@/lib/i18n';
 
 /**
- * Creates an empty draft proposal for the instance and navigates to its edit
- * page. `isCreating` stays true through the redirect on success so buttons
- * can keep showing a pending state.
+ * Creates an empty draft proposal for the instance and navigates to the href
+ * the caller builds from the created proposal. `isCreating` stays true through
+ * the navigation so buttons can keep showing a pending state.
  *
  * Public (no-session) visitors are given an anonymous session first when the
  * `anonymous_signin` flag is on, so the draft has an account to attribute to.
  */
 export function useCreateProposal({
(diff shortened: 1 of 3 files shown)


apps/app/src/components/decisions/useCreateProposal.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Creates an empty draft proposal for the instance and navigates to its edit
+ * page. `isCreating` stays true through the redirect on success so buttons
+ * can keep showing a pending state.
+ *
+ * Public (no-session) visitors are given an anonymous session first when the
+ * `anonymous_signin` flag is on, so the draft has an account to attribute to.
+ */
+export function useCreateProposal({
+  instanceId,
+  decisionSlug,
```
valentin0h: I'd rather we pass a `navigateTo` or `navigateAfter` than construct the path in this hook.

apps/app/src/components/decisions/useCreateProposal.ts
```diff
@@ -0,0 +1,74 @@
...
+  const { user } = useUser();
+  const [isCreating, setIsCreating] = useState(false);
+  const supabase = createSBBrowserClient();
+  const utils = trpc.useUtils();
+  const anonymousSigninEnabled = useFeatureFlag('anonymous_signin');
+
+  const createProposalMutation = trpc.decision.createProposal.useMutation();
+
+  const createProposal = async () => {
+    setIsCreating(true);
```
valentin0h: Can we use `startTransition`?

valentin0h (approved):
Nice, I guess a second usage is comming?

nourmalaeb:
> Nice, I guess a second usage is comming?

We have two usages now, in `DecisionActionBar` and `DecisionOverview`

apps/app/src/components/decisions/useCreateProposal.ts
nourmalaeb: Good call, I made that change

apps/app/src/components/decisions/useCreateProposal.ts
nourmalaeb: Yes much better

PR #1293, opened 2026-06-11
fix(decisions): persist only dirty fields in process builder localStorage

nourmalaeb:
## Summary

- Visiting the process builder of a published process stamped a full server snapshot into localStorage, which shadowed other admins' saved changes on later visits (one admin's rubric criteria invisible to other admins)
- "Update Process" sent that stale snapshot back, silently reverting other admins' saved edits
- Store now persists only a per-decision dirty-field map; the merged instance view is in-memory and reseeded from the server on every load
- "Update Process" sends only the fields the user actually edited
- Persist version bump discards everyone's legacy full snapshots on deploy, so stale state self-heals

## Tests

- E2E: two-admin repro in `tests/e2e/tests/process-builder-multi-admin.spec.ts` — fails against pre-fix build, passes with fix
- Unit: store persistence contract in `apps/app` — **note: this adds the first vitest setup to the app workspace** (config + test script). Covers the legacy-snapshot migration discard and per-field dirty semantics that e2e can't reach. Open to dropping if we don't want apps/app as a unit-test surface.

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -13,6 +13,7 @@
     "format:check": "oxfmt --check",
     "start": "next start -p 3100",
     "start:e2e": "NEXT_PUBLIC_APP_PORT=4100 NEXT_PUBLIC_API_PORT=4300 next start -p 4100",
+    "test": "vitest run",
     "typecheck": "tsgo --noEmit"
   },
   "dependencies": {
--- a/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext.tsx
+++ b/apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { trpc } from '@op/api/client';
+import { ProcessStatus } from '@op/api/encoders';
 import { useDebouncedCallback } from '@op/hooks';
 import { toast } from '@op/ui/Toast';
 import {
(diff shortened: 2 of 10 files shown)


apps/app/src/components/decisions/ProcessBuilder/stores/useProcessBuilderStore.ts
```diff
@@ -244,26 +311,80 @@ export const useProcessBuilderStore = create<ProcessBuilderState>()(
...
+        }),
+
+      clearDirtyFields: (decisionId, fields) =>
+        set((state) => {
+          const existing = state.dirty[decisionId];
+          if (!existing) {
+            return state;
+          }
+
+          const { config: savedConfig, ...savedRest } = fields;
```
scazan, line 327: nit: usage of the variable name "rest" is really local to this spread.. beyond this line it's not really descriptive. We should give them better variable names. In this case, maybe savedFields or even savedFieldsWithoutConfig.

apps/app/src/components/decisions/ProcessBuilder/ProcessBuilderAutosaveContext.tsx
```diff
@@ -125,14 +137,24 @@ export function ProcessBuilderAutosaveProvider({
...
+      promise
+        .then(() => {
+          // Keep the persisted dirty map holding unsaved/failed edits only.
+          clearDirtyFields(decisionProfileId, payload);
+        })
+        .catch(() => {
+          // Error toasted by onError. Restore the payload so the next save
+          // retries it (newer edits win); no scheduled retry — a rejecting
+          // payload would loop.
+          dirtyFieldsRef.current = {
```
scazan, line 149: Should this go to the mutation's onError callback instead?

apps/app/package.json
```diff
@@ -13,6 +13,7 @@
     "format:check": "oxfmt --check",
     "start": "next start -p 3100",
     "start:e2e": "NEXT_PUBLIC_APP_PORT=4100 NEXT_PUBLIC_API_PORT=4300 next start -p 4100",
+    "test": "vitest run",
```
scazan, line 16: Is this and vitest.config meant to be in this PR? maybe a remnant of something?

scazan (approved):

PR #1294, opened 2026-06-12
feat(db): add decision_boundaries table for district tagging

bloudermilk:
## Stage 1 of 2 — migrations first

Adds the `decision_boundaries` table that backs automatic council-district tagging of proposals. **This should be merged and deployed before the feature PR**, which depends on the table existing.

### What's here
- `decision_boundaries` table: named MultiPolygon boundaries (WGS84), each optionally linked to a proposal-taxonomy category (`taxonomy_term_id`), GIST-indexed for `ST_Contains`, unique on `lower(name)`.
- A `multiPolygon` Drizzle custom type (built-in `geometry()` only supports points).
- v2 relation (boundary ↔ taxonomy term) + schema export.
- The migration was regenerated on top of the latest `dev` so its snapshot chain is current; the SQL was trimmed to only the `decision_boundaries` statements (a stale upstream snapshot caused `generate` to re-emit an already-applied `access_role_permissions.profile_id` change).

### Follow-up
The feature PR (auto-tagging, validation, import CLI, UI) builds on this branch and targets `dev` separately.

### DIFF
--- a/services/db/helpers/customTypes.ts
+++ b/services/db/helpers/customTypes.ts
@@ -6,6 +6,20 @@ export const tsvector = customType<{ data: string }>({
   },
 });
 
+/**
+ * PostGIS MultiPolygon geometry in WGS84 (SRID 4326).
+ *
+ * Drizzle's built-in `geometry()` only models `point`, so polygon boundaries
+ * use this custom type. The column is opaque to the ORM — it is written and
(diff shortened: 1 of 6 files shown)


scazan (approved):
nice. You might need to rebase off the latest dev just to be sure the migration snapshot is up to date.

bloudermilk:
@scazan are you able to authorize this so we can see CI pass?

PR #1295, opened 2026-06-12
refactor(decisions): drop localStorage persistence from process builder

nourmalaeb:
## Summary

Stacked on #1293. Removes localStorage persistence from the process builder — what you see is always server data or this session's edits, never a stale persisted snapshot.

Why: the dirty-field overlay made localStorage *safe* but still misleading for collections (categories, phases, rubric criteria) — an unsaved local array hides items other admins have since saved, which is the original multi-admin confusion in a milder form. Draft edits autosave within 1s, so persistence bought them almost nothing, and the hydration machinery was the largest complexity tax in the ProcessBuilder.

- Store is in-memory only: persist middleware, versioning, migration, and the whole hydration dance (`skipHydration` / `onFinishHydration` / `rehydrate`) deleted
- Initializer seeds synchronously during render — the five section hydration gates and their skeleton-while-hydrating states go away
- "Unsaved changes" indicator next to "Update Process" while published edits are pending; `beforeunload` warns on hard unloads
- Cleans up the legacy `process-builder` localStorage key on mount
- Dirty map and dirty-only save payloads from #1293 unchanged; unit + e2e suites adapted and green

Unsaved published edits now die with the page (refresh/close) — visible via the indicator and warned via `beforeunload`, instead of silently persisting a misleading local copy.


scazan (approved):
Works well! glad to reduce some of this complexity

PR #1296, opened 2026-06-14
feat(moderation): hide flagged content from reads + Flagged indicator

scazan:
Read-side of content moderation, split out of the async-flow PR so it can be reviewed on its own. Given an active moderation flag on an item, hide it from general reads while keeping it visible — with a **Flagged** indicator — to the people allowed to act on it (the author and admins). Depends only on the already-merged `moderation_flags` schema, and is inert until a flag reaches `flagged`/`confirmed` (which the stacked async pipeline produces).

When is a proposal hidden:
Screenshot 2026-06-15 at 14 41 19

Filtering is a correlated `NOT EXISTS` against `moderation_flags` **inside the paginated SQL** (never a post-fetch JS filter), via `noActiveModerationFlag()` in `moderationVisibility.ts`.

| Read | Who still sees a flagged item | Notes |
|---|---|---|
| `listProposals` / `listAllProposals` | creator; instance admins skip the filter | results carry `isFlagged` |
| `getProposal` | creator + collaborators; instance admins | otherwise 404 — non-leaking, same audience as `HIDDEN` |
| `listPosts` / `getPosts` / `listProfilePosts` | author; org/profile admins skip the filter | filtered on the paginated SQL stage |
| `listRelatedOrganizationPosts` | author | aggregate feeds span many orgs (no single governing profile to admin-check) |

Authors/admins see a Flagged indicator (`LuFlag` + `text-functional-red`) in `ProposalPreview` and `PostFeed`, driven by the new `isFlagged` field on the proposal/post payloads. A DB-backed visibility test (`services/api`) covers hidden-from-outsider and visible-to-author/admin-with-`isFlagged` for both posts and proposals.

Stacked under: the async moderation flow (#1269) builds on this.


apps/app/src/components/decisions/ProposalPreview.tsx
```diff
@@ -111,6 +112,15 @@ export function ProposalPreview({
           </div>
         )}
 
+        {/* Only the author (+ collaborators) and admins ever receive a flagged
+            proposal — everyone else has it filtered out server-side. */}
+        {proposal.isFlagged && (
+          <div className="flex items-center gap-1 text-sm text-functional-red">
```
scazan, line 118: Styles on this are not quite right, but I want to submit a follow-up here for style adjustments.

packages/common/src/services/posts/getPost.ts
```diff
@@ -96,9 +115,49 @@ export const getPost = async ({
     },
   });
 
+  // Resolve the flag state once (reused below for the gate and for the
+  // `isFlagged` decoration, so the moderation_flags table is hit a single time).
+  const flaggedItemIds = await getActivelyFlaggedItemIds('post', [post.id]);
```
scazan: Assumes a "shorter" comment trail... but let's address that separately as I'm not happy with our posts authorization at the moment.

services/api/src/encoders/posts.ts
```diff
@@ -28,6 +28,10 @@ const basePostsEncoder = createSelectSchema(posts)
       .optional(),
     userReaction: z.string().nullish(),
     commentCount: z.number(),
+    // True only on posts the read filters let through to their author or an
+    // admin while an active moderation flag hides them from everyone else;
+    // drives the "Flagged" indicator. Absent on legacy/unenriched payloads.
+    isFlagged: z.boolean().optional(),
```
valentin0h, line 34: Wonder if, because this is an owner-only property, it's worth surfacing more detail, like the status. 
But we can do that at a later stage.

packages/common/src/services/posts/listRelatedOrganizationPosts.ts
```diff
@@ -121,21 +138,29 @@ export const listRelatedOrganizationPosts = async (
   const orgIds = organizations?.map((org: any) => org.id) ?? [];
   orgIds.push(organizationId); // Add our own org so we see our own posts
 
+  const actorProfileId = await getCurrentProfileId(user.id);
```
valentin0h, line 141: Added an Asana ticket to make this a request state rather than a global state.

packages/common/src/services/access/index.ts
```diff
@@ -229,6 +229,43 @@ export const getProfileAccessUser = memoize(
...
+  user,
+  profileId,
+}: {
+  user?: AccessUser;
+  profileId: string;
+}): Promise<NormalizedRole[]> => {
+  const profileUser = await getProfileAccessUser({ user, profileId });
+  if (profileUser) {
+    return profileUser.roles;
+  }
```
valentin0h, line 255: Maybe not a realistic use case given which entities can currently be made public, but it looks like we might get the global user roles here and never fall back to the org.

valentin0h (approved):
lgtm

packages/common/src/services/access/index.ts
scazan: ah good point with the new changes for global roles.

scazan: I'll file a followup that to look at how we want to handle that.

PR #1297, opened 2026-06-15
feat(decisions): sign in anonymous visitors to start a proposal

valentin0h:
Lets public visitors start a proposal by giving them an anonymous session (behind the anonymous_signin flag) so the draft has an account to attribute.

### DIFF
--- a/packages/common/src/services/user/index.ts
+++ b/packages/common/src/services/user/index.ts
@@ -64,7 +64,9 @@ export const getUserByAuthId = async ({
   const user = await db._query.users.findFirst({
     where: (table, { eq }) => eq(table.authUserId, authUserId),
     with: {
-      authUser: true,
+      authUser: {
+        columns: { isAnonymous: true },
+      },
       avatarImage: true,
       organizationUsers: {
         with: {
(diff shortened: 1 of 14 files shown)


packages/common/src/services/user/index.ts
```diff
@@ -64,6 +64,7 @@ export const getUserByAuthId = async ({
   const user = await db._query.users.findFirst({
     where: (table, { eq }) => eq(table.authUserId, authUserId),
     with: {
+      authUser: true,
```
scazan: One reason for previously not adding this in prior is that it makes it way too easy to leak details related to auth to other users (including things that might be auth only like phone number). Any reason it was added here?

scazan (approved):
If we aren't explicitly making use of that authUser let's not include it in the getUserByAuthId. Otherwise we should probably do a select on it to just pull out some details.

packages/common/src/services/user/index.ts
valentin0h: We should fix this in the encoder/schema level I think. But I'll fix here as well. We need `is_anonymous`

PR #1299, opened 2026-06-15
feat(decisions): mark required fields in proposal editor and preview

nourmalaeb:
- Render a red asterisk next to required field labels in the proposal template preview (readonly text + dropdown fields).
- Mirror the same marking in the live proposal editor (collaborative text + dropdown fields).
- Thread `required` through `FieldHeader`, the readonly field components, and `CollaborativeTextField`.
- Wire `aria-required` on the editable inputs (rich text contenteditable + collaborative select) so the asterisk isn't the only required cue.

Supersedes #1298 (closed by a branch rename).

### DIFF
--- a/apps/app/src/components/collaboration/CollaborativeTextField.tsx
+++ b/apps/app/src/components/collaboration/CollaborativeTextField.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { RequiredAsterisk } from '@op/ui/RequiredAsterisk';
 import { cn } from '@op/ui/utils';
 import Placeholder from '@tiptap/extension-placeholder';
 import type { Editor } from '@tiptap/react';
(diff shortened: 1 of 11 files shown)


apps/app/src/components/collaboration/CollaborativeTextField.tsx
```diff
@@ -97,6 +100,12 @@ export function CollaborativeTextField({
           {title && (
             <span className="font-serif text-title-sm14 text-neutral-charcoal">
               {title}
+              {required && (
+                <span className="text-functional-red" aria-hidden="true">
```
scazan: We should componentize this since we use it multiple times here and elsewhere

scazan (approved):

apps/app/src/components/collaboration/CollaborativeTextField.tsx
nourmalaeb: will do!

PR #1300, opened 2026-06-15
feat: RTL-aware text direction handling for Arabic

nourmalaeb:
RTL/Arabic text-direction support across UI primitives and decision views.

- `@op/ui`: free-text inputs use `unicode-bidi: plaintext` (empty inherits locale dir, filled follows content); constrained types (email/url/number/tel/password) forced LTR with icon pinned left. `NumberField` normalizes Arabic-Indic/Persian numerals to ASCII. `Header1-4` accept `dir` (default `auto`). RichText components forward `dir` instead of hardcoding it.
- App: `dir="auto"` on content that can be either script (proposal title/body, prose renderer, resource cards, search, collaborative editors) so LTR content lays out correctly in an RTL locale (alignment, list markers, truncation side).
- Prose: per-block `unicode-bidi: plaintext` in `viewerProseStyles` so mixed LTR/RTL paragraphs each align correctly.
- Swapped remaining physical Tailwind classes (`pr-`/`right-`/`text-left`/`border-l`/masonry gutter) for logical equivalents.

### DIFF
--- a/packages/ui/src/components/Header.tsx
+++ b/packages/ui/src/components/Header.tsx
@@ -5,41 +5,65 @@ import { cn } from '../lib/utils';
 export const Header1 = ({
   children,
   className,
+  dir,
 }: {
   children: React.ReactNode;
   className?: string;
+  dir?: 'ltr' | 'rtl' | 'auto';
 }) => {
-  return <h1 className={cn(headingClasses.h1, className)}>{children}</h1>;
+  return (
+    <h1 dir={dir} className={cn(headingClasses.h1, className)}>
+      {children}
+    </h1>
+  );
 };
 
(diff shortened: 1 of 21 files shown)


packages/ui/src/components/Header.tsx
```diff
@@ -5,41 +5,65 @@ import { cn } from '../lib/utils';
...
 };
 
 export const Header4 = ({
   children,
   className,
+  dir = 'auto',
 }: {
   children: React.ReactNode;
   className?: string;
+  dir?: 'ltr' | 'rtl' | 'auto';
```
scazan, line 60: This should be a type we can re-use

scazan (approved):

PR #1303, opened 2026-06-16
feat(decisions): show a log in button for signed-out visitors in the decision header

valentin0h:
Signed-out visitors viewing a public decision saw nothing in the header where the account menu sits; this surfaces a log in CTA so they can sign in.


valentin0h:
Split into a stack to separate concerns: #1305 (CommonUser.isAnonymous / encodeUser) and #1306 (the decision-header log in button, stacked on #1305). Closing this in favor of those.

PR #1304, opened 2026-06-16
feat(decisions): infinite scroll on decision page proposal listing

op-bot-common:
Mirrors the posts feed pattern (`DecisionSidePanel.UpdatesFeed`): swap the single-page `useSuspenseQueries` fetch in `ProposalsList` for `useSuspenseInfiniteQuery` + `useInfiniteScroll` sentinel. The `listProposals` service gains an opaque pagination cursor that today encodes the next offset, keeping the offset-based query compatible with TanStack Query's infinite-pagination interface without touching the heavier `listAllProposals` keyset path (which already returns `next`).

The component splits into a thin `ProposalsList` shell that owns the filter / sort state and a `ProposalsListContent` that consumes pre-fetched pages via per-phase loader components (`CurrentPhaseProposalsLoader`, `ResultsPhaseProposalsLoader`). Each loader owns its own `useSuspenseInfiniteQuery` and `useInfiniteScroll` wiring, so React's rules-of-hooks stay clean when `phase` switches at the page boundary. Translation state moves into `useProposalsTranslation` and the filter UI into `ProposalsFilterBar`; `useProposalFilterItems` is hoisted into `useProposalFilters.ts` so `ManualSelectionToolbar` shares the same item set instead of inlining a near-duplicate copy.

E2E: `proposal-listing-infinite-scroll.spec.ts` seeds 55 proposals, asserts only the first 50 render initially, then `scrollIntoViewIfNeeded`s the last loaded card so the IntersectionObserver fires regardless of nested scroll containers.

Asana: https://app.asana.com/0/1213315744811204/1212017248821706


apps/app/src/components/decisions/ProposalsGrid.tsx
```diff
@@ -0,0 +1,475 @@
+'use client';
```
scazan, line 1: Moved verbatim out of `ProposalsList.tsx` (the only change is the rename `Proposals` → `ProposalsGrid`, to avoid confusion with `ProposalsList`). No logic changes

apps/app/src/components/decisions/ProposalsFilterBar.tsx
```diff
@@ -0,0 +1,150 @@
+'use client';
```
scazan, line 1: Extracted verbatim from `ProposalsList.tsx` in the split commit. No logic changes.

apps/app/src/components/decisions/useProposalsTranslation.ts
```diff
@@ -0,0 +1,138 @@
+'use client';
```
scazan, line 1: Extracted verbatim from `ProposalsList.tsx` in the split commit. No logic changes.

apps/app/src/components/decisions/ProposalListSkeleton.tsx
```diff
@@ -0,0 +1,67 @@
+import { Skeleton } from '@op/ui/Skeleton';
```
scazan, line 1: Extracted verbatim from `ProposalsList.tsx` in the split commit. No logic changes; the 5 prior importers of `ProposalListSkeleton` were repointed here.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -8,614 +8,293 @@ import {
...
+    total,
+    isFetchingNextPage,
+    shouldShowTrigger,
+    infiniteScrollRef: ref,
+  };
 };
 
-const NoProposalsFound = ({ hasFilter }: { hasFilter: boolean }) => {
-  const t = useTranslations();
+const CurrentPhaseProposalsLoader = ({
```
scazan, line 123: Why two loaders: the current phase uses `listProposals` and the results phase uses `listAllProposals`, which return different page shapes — and suspense queries can't be called conditionally. Each loader runs its own infinite query and feeds the shared `ProposalsListContent` via a render prop.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -8,614 +8,293 @@ import {
...
+  );
+  const [sortOrder, setSortOrder] = useQueryState(
+    'sort',
+    parseAsString.withDefault('newest'),
+  );
+  const [filterParam, setProposalFilter] = useQueryState(
+    'filter',
+    parseAsStringLiteral(PROPOSAL_FILTER_VALUES),
+  );
+  const requestedFilter =
```
scazan, line 203: Filter state lives in the URL (nuqs). `filter` is intentionally left without a default so an absent value falls back to the ballot-aware default here (→ "My ballot" once the user has voted). The next block downgrades the profile-only filters to ALL for anonymous/stale-link visitors so the header can't claim a filter the query didn't actually apply.

packages/common/src/services/decision/listAllProposals.ts
```diff
@@ -69,52 +87,107 @@ export const listAllProposals = async ({
...
+        profile: true,
       },
-      profile: true,
-    },
-    limit: limit + 1, // Fetch one extra to check whether there's a next page.
-    orderBy: (table, { asc, desc }) =>
-      dir === 'asc' ? asc(table[orderBy]) : desc(table[orderBy]),
-  });
+      limit: limit + 1, // Fetch one extra to check whether there's a next page.
+      // `id` tie-break: without it, rows sharing the sort value are skipped at page boundaries.
```
scazan, line 208: The `id` tie-breaker (order + cursor) was added here: paging on `createdAt`/`updatedAt` alone skips rows that share a boundary timestamp, which matters now that this endpoint drives results-phase infinite scroll. Mirrors `listProposals`. Covered by the new "does not skip rows that share a boundary timestamp" test.

scazan (commented):
A few inline notes to make review easier: the four new component files are verbatim extractions from `ProposalsList.tsx` (the split commit), and I've flagged a couple of spots that need context.

packages/common/src/services/decision/listProposals.ts
```diff
@@ -574,10 +605,24 @@ export const listProposals = async ({
     };
   });
 
+  // Keyset cursor off the last row's sort value + id tiebreak (votes can't
+  // keyset, so it stays a single page).
+  const lastItem = proposalsWithCounts[proposalsWithCounts.length - 1];
+  const cursorValue =
+    lastItem && orderBy !== 'votes' ? lastItem[orderBy] : null;
+  const next =
+    hasMore && lastItem && cursorValue
```
nourmalaeb: If we sort on rubric scores or something where `cursorValue` can be falsy (e.g the rubric score is `0`) this breaks. Could harden to:

```
hasMore && lastItem && cursorValue != null
```

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -8,614 +8,293 @@ import {
...
-  const t = useTranslations();
+const CurrentPhaseProposalsLoader = ({
+  queryParams,
+  children,
+}: {
+  queryParams: ProposalQueryParams;
+  children: (data: ProposalsLoaderRenderProps) => React.ReactNode;
+}) => {
+  const [paginatedData, query] =
+    trpc.decision.listProposals.useSuspenseInfiniteQuery(queryParams, {
```
nourmalaeb, line 131: Just noting that since the query is keyed on `queryParams` the skeleton flashes when the filters change. It's fine I think, but would be ideal to keep the filters themselves separate rather than have them also be part of the suspended UI. Not for this PR I reckon.

apps/app/src/components/decisions/ProposalsGrid.tsx
```diff
@@ -0,0 +1,475 @@
...
+
+  return (
+    <ViewProposalsList
+      {...props}
+      revisionRequestIdByProposalId={revisionRequestIdByProposalId}
+    />
+  );
+};
+
+const NoProposalsFound = ({ hasFilter }: { hasFilter: boolean }) => {
```
nourmalaeb, line 80: This `hasFilter` prop passed down from `ProposalsList.tsx` only acounts for the category filter, not the All/Mine/Shortlister filter. So if I filter by shortlisted and there are no shortlisted proposals I see the "No propsals yet" empty state instead of the "No proposal found" empty state.

Image

packages/common/src/services/moderation/providers/checkstep.ts
nourmalaeb, line 1: Was this (and related files) meant to be part of this PR? Seems unrelated to infinite scroll

nourmalaeb (approved):
No blockers but left some comments

packages/common/src/services/moderation/providers/checkstep.ts
scazan: This functionality was blocking our ability to work on the moderation stuff which had to to get done by Friday, so I just pushed to this branch so she could work from here. It's unrelated but it's piggybacking here.

packages/common/src/services/decision/listProposals.ts
scazan: Good catch — done in d33aebee6. Both `listProposals` and `listAllProposals` now gate the cursor on `cursorValue != null` instead of a truthy check, so a falsy-but-valid sort value (e.g. a rubric score of 0) still produces a cursor instead of silently ending pagination.

apps/app/src/components/decisions/ProposalsGrid.tsx
scazan: Fixed in d33aebee6 — `hasFilter` now also accounts for the All/Mine/Shortlisted filter (`proposalFilter !== ProposalFilter.ALL`), not just the category, so an empty shortlisted result shows "none match your filters" instead of "no proposals yet".

apps/app/src/components/decisions/ProposalsList.tsx
scazan: Agreed — leaving as-is for this PR. The filter controls being part of the suspended subtree is what causes the flash; pulling them out above the Suspense boundary is the right fix but it's a separate change.

PR #1305, opened 2026-06-16
feat(api): make CommonUser.isAnonymous a required field sourced from the session

valentin0h:
Adds an encodeUser helper that derives isAnonymous from the Supabase auth identity (ctx.user, or an admin getUserById result) instead of the authUser DB relation, and drops the now-unused authUser join in getUserByAuthId.


scazan (approved):
Super nice.

PR #1307, opened 2026-06-16
fix(api): open companion read procedures to anonymous callers

valentin0h:
Comments, like/follower data, and editor revision requests still required network auth, so they failed for anonymous participants on public decision/proposal pages. Opening these reads lets the existing service-layer authorization decide access and restores legitimate anonymous flows.

fix(api): open posts.getPosts to anonymous callers
fix(api): open profile.getRelationships to anonymous callers
fix(api): open account.getUserProfiles to anonymous callers
fix(api): open decision.listProposalRevisionRequests to anonymous callers

### DIFF
--- a/packages/common/src/services/profile/profileRelationships.ts
+++ b/packages/common/src/services/profile/profileRelationships.ts
@@ -8,7 +8,7 @@ import {
 } from '@op/db/schema';
 import { alias } from 'drizzle-orm/pg-core';
 
-import { ValidationError } from '../../utils/error';
+import { UnauthorizedError, ValidationError } from '../../utils/error';
 import { getCurrentProfileId } from '../access';
 
 export const addRelationship = async ({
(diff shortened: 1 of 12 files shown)


packages/common/src/services/profile/profileRelationships.ts
```diff
@@ -175,11 +175,10 @@ export const getRelationships = async ({
   if (sourceProfileId) {
     conditions.push(eq(profileRelationships.sourceProfileId, sourceProfileId));
   } else if (!targetProfileId) {
-    // Only resolve currentProfileId when needed as fallback (avoids extra DB query)
-    const currentProfileId = await getCurrentProfileId(authUserId);
-    if (!currentProfileId) {
-      throw new ValidationError('You must be logged in to view relationships');
+    if (!authUserId) {
+      return [];
```
scazan: not an auth error?

scazan (approved):

packages/common/src/services/profile/profileRelationships.ts
valentin0h: good point, missed that.

PR #1314, opened 2026-06-16
feat(moderation): proposal Report action + flagged UI

scazan:
Stacked on #1320 (which is stacked on #1269).

The proposal-side moderation UI. Adds a **Report** action in the proposal header that opens a confirmation dialog and sends the proposal for asynchronous review via the `moderation.flagItem` route (from #1320); a "Hidden from members after a moderation review" indicator on a flagged proposal's view; and a red **Flagged** badge on proposal list cards, alongside the existing Selected/Shortlisted badges. The Report action is gated behind the `moderation` feature flag. User-facing strings are translated across all 7 dictionaries.

Depends on the `moderation.flagItem` tRPC route in #1320 — won't typecheck until that lands.


apps/app/src/components/decisions/ReportProposalDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  // Gated behind the `moderation` flag until the feature ships. Checked after
+  // the hooks above so the rules of hooks hold.
+  if (!moderationEnabled) {
+    return null;
+  }
+
+  return (
+    <>
+      <Button
```
nourmalaeb, line 48: Is there supposed to be a different appearance for the button if it is reported or not? An outline or something?

nourmalaeb (approved):
I think we're missing a "reported" state for the Report button. Otherwise looks good!

PR #1316, opened 2026-06-16
feat(decisions): wire overview page to real instanceData.overview

nourmalaeb:
- Read overview content (headline/description/body) from `instanceData.overview` instead of the placeholder mock; delete `decisionOverviewMock`.
- Map hero subhead from `description`; body is an HTML string rendered via the existing `HtmlContentRenderer` path.
- Fall back to `instance.name` when the overview headline is empty or absent, and to `instance.description` when body is absent.


apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -99,7 +99,7 @@ function DecisionOverviewContent({
         </div>
         <div className="min-w-0 md:col-span-7 md:col-start-6">
           <OverviewAbout
-            html={overview.body}
+            html={overview?.body}
```
scazan: We should maybe just make this prop `content` and then it's type is HTML (or string). We want to move this type to json eventually I think.

scazan (approved):

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: sounds good

PR #1317, opened 2026-06-16
feat(ui): add PhaseTimeline component

nourmalaeb:
First of a 3-PR stack for the decision overview phase timeline.

- New presentational `PhaseTimeline` (@op/ui), vertical sibling to `PhaseStepper`.
- Renders the four Figma states: completed (green rail), current (filled teal card + "Now open!" tag + arrow), upcoming (gray rail), advanceable (off-white card + Advance button).
- Reuses the `Phase` type and `formatDateRange`; Storybook story covers all states.


packages/styles/shared-styles.css
```diff
@@ -190,6 +190,7 @@
    * ============================================ */
   --color-primary-50: var(--op-primary-50);
   --color-primary-100: var(--op-primary-100);
+  --color-primary-200: var(--op-primary-200);
```
nourmalaeb, line 193: Used by the new component

packages/ui/src/components/PhaseCard.tsx
```diff
@@ -0,0 +1,193 @@
...
+  state,
+  className,
+  isNowOpen,
+  nowOpenLabel = 'Now open!',
+  isAdvanceable,
+  advanceLabel = 'Advance',
+  onAdvance,
+  href,
+}: PhaseCardProps) {
+  if (state === 'completed') {
```
scazan: small thing, can we split these three into components so we don't have a big 'ol if statement in the function. this way we can easily see the branching and can isolate each component's dependencies (no variable scope bleeding into the individual components). keeps it modular.

even a good use-case for a match()!

scazan (approved):

packages/ui/src/components/PhaseCard.tsx
nourmalaeb: Done — split into a thin dispatcher plus one self-contained component per treatment (CompletedPhaseCard / CurrentPhaseCard / AdvanceablePhaseCard / UpcomingPhaseCard), each with its own narrowed props. Used match()!

PR #1318, opened 2026-06-16
refactor(decisions): extract useAdvancePhase hook

nourmalaeb:
Second of a 3-PR stack. Stacked on #1317.

- Extract the advance-to-next-phase flow (transitionFromPhase mutation, confirm Modal/Sheet, PostHog tracking, success toast + refresh) out of `DecisionProcessStepper` into a shared `useAdvancePhase` hook.
- `DecisionProcessStepper` now consumes the hook; no behavior change.
- Lets the overview phase timeline (next PR) reuse the same advance flow.

### DIFF
--- a/apps/app/src/components/decisions/useAdvancePhase.tsx
+++ b/apps/app/src/components/decisions/useAdvancePhase.tsx
@@ -15,7 +15,7 @@ import { useRouter, useTranslations } from '@/lib/i18n';
 interface UseAdvancePhaseArgs {
   instanceId?: string;
   /** Phase the process is currently in (the one being advanced out of). */
-  currentStateId: string;
+  currentPhaseId: string;
   /** Phase the process advances into. Undefined when there is no next phase. */
   nextPhaseId?: string;
   /** Translated names used in the confirmation copy. */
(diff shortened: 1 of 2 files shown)


apps/app/src/components/decisions/useAdvancePhase.tsx
```diff
@@ -0,0 +1,210 @@
...
+import { toast } from '@op/ui/Toast';
+import { usePostHog } from 'posthog-js/react';
+import { type ReactNode, useCallback, useRef, useState } from 'react';
+
+import { useRouter, useTranslations } from '@/lib/i18n';
+
+interface UseAdvancePhaseArgs {
+  instanceId?: string;
+  /** Phase the process is currently in (the one being advanced out of). */
+  currentStateId: string;
```
scazan: This comes in from the DB as currentState but it's confusing to have state and phase interchanged. maybe its a good opportunity to name this prop currentPhaseId instead.

scazan (approved):

apps/app/src/components/decisions/useAdvancePhase.tsx
nourmalaeb: Good call — renamed the hook arg to currentPhaseId. The DB-sourced currentStateId now maps to it at the two call sites (DecisionProcessStepper, DecisionPhaseTimeline).

PR #1320, opened 2026-06-16
feat(moderation): user-initiated content reports (flagItem)

scazan:
Stacked on #1269.

The user-initiated report path, split out of #1269 so the automated-moderation flow can be reviewed on its own. Adds the `flagItem` / `submitUserFlag` services, the `assertModerationItemAccess` read-gate (flagging ships an item's text + signed attachment URLs to the external provider, so it's gated exactly like reading the item — posts like `getPost`, proposals like `getProposal`, user profiles existence-only), and the rate-limited `moderation.flagItem` tRPC route. A report records a `pending` flag and submits a fresh round to the provider; the verdict (handled in #1269) later confirms or clears it.

Depends on the shared stores, provider integration, and verdict pipeline in #1269.

### DIFF
--- a/packages/common/src/services/moderation/assertModerationItemAccess.ts
+++ b/packages/common/src/services/moderation/assertModerationItemAccess.ts
@@ -1,11 +1,5 @@
-import { db, eq } from '@op/db/client';
-import {
-  EntityType,
-  Visibility,
-  posts,
-  postsToProfiles,
-  users,
-} from '@op/db/schema';
+import { db } from '@op/db/client';
+import { EntityType, Visibility } from '@op/db/schema';
 import type { User } from '@op/supabase/lib';
 import { checkPermission, permission } from 'access-zones';
 
@@ -20,13 +14,18 @@ import { hasActiveModerationFlag } from './moderationVisibility';
 import type { ModerationItemType } from './types';
 
 /**
- * Asserts the reporter can actually see the item they're flagging, and that it
+ * Asserts the caller can actually see the item they're flagging, and that it
  * exists. Flagging ships the item's text and signed attachment URLs to the
  * external moderation provider, so it must be gated like reading the item —
- * otherwise an authenticated session could exfiltrate restricted content (or
- * probe for ids) by flagging it. Mirrors the read-path checks: posts gate like
- * `getPost`, proposals like `getProposal`; user profiles are platform-visible,
- * so a user item only needs to exist.
+ * otherwise a caller could exfiltrate restricted content (or probe for ids) by
+ * flagging it. Mirrors the read-path checks: posts gate like `getPost`,
+ * proposals like `getProposal`; user profiles are platform-visible, so a user
+ * item only needs to exist.
+ *
+ * `user` is optional: a sessionless caller resolves to public grants only (via
+ * the `GLOBAL_USER_PUBLIC` fallback in the access helpers), so they can flag a
+ * publicly readable item but are denied anything restricted — exactly the read
+ * path's behavior for an anonymous visitor.
  *
  * Two intentional carve-outs from the read path:
  *  - DRAFT proposals are NOT gated here — drafts are reportable.
@@ -40,18 +39,17 @@ export const assertModerationItemAccess = async ({
 }: {
   itemType: ModerationItemType;
   itemId: string;
-  user: User;
+  user?: User;
 }): Promise<void> => {
   if (itemType === 'post') {
-    const [post] = await db
-      .select({
-        id: posts.id,
-        profileId: posts.profileId,
-        rootProfileId: posts.rootProfileId,
-      })
-      .from(posts)
-      .where(eq(posts.id, itemId))
-      .limit(1);
+    const post = await db.query.posts.findFirst({
+      where: { id: itemId },
+      columns: {
+        id: true,
+        profileId: true,
+        rootProfileId: true,
+      },
+    });
     if (!post) {
       throw new NotFoundError('Post', itemId);
     }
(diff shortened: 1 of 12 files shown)


services/api/src/routers/moderation/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  created: z.boolean(),
+});
+
+export const moderationRouter = router({
+  // A signed-in user reports an item. Records a pending flag and submits it for
+  // async provider review; the provider's webhook confirms or clears it.
+  // Tighter than the default rate limit: each call signs attachment URLs and
+  // makes outbound provider submissions, and nobody legitimately reports five
+  // items in a minute.
+  flagItem: authenticatedProcedure({
```
scazan: this should be an open procedure (using that router middleware ) instead

packages/common/src/services/moderation/flagItem.ts
```diff
@@ -0,0 +1,130 @@
...
+          callbackUrl: input.callbackUrl,
+        });
+      } catch (error) {
+        await deps.rollback?.(flag);
+        throw error;
+      }
+    }
+  }
+
+  return { flag, created: true };
```
scazan: do we need to pass "created: true" back? Is it useful?

services/api/src/routers/moderation/flagItem.test.ts
valentin0h, line 1: to keep consistency with other tests

- let's have only `flagItem.test.ts`
- move `describeAccessTierGating` to that file.

services/api/src/routers/moderation/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  flagId: z.uuid(),
+});
+
+export const moderationRouter = router({
+  // A signed-in user reports an item. Records a pending flag and submits it for
+  // async provider review; the provider's webhook confirms or clears it.
+  // Tighter than the default rate limit: each call signs attachment URLs and
+  // makes outbound provider submissions, and nobody legitimately reports five
+  // items in a minute.
+  flagItem: openProcedure({
```
valentin0h, line 17: Comment and code don't match here. I'm guessing we want either `authenticatedProcedure` or `authenticatedConfirmedProcedure`

services/api/src/routers/moderation/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  })
+    .input(flagItemInputSchema)
+    .output(flagItemOutputSchema)
+    .mutation(async ({ ctx, input }) => {
+      // openProcedure resolves an optional user; reporting records the reporter
+      // and ships the item's content to the provider, so fail closed without a
+      // session rather than letting an anonymous caller flag content.
+      if (!ctx.user) {
+        throw new UnauthorizedError('Must be signed in to report content');
+      }
```
valentin0h: will go if we change the procedure type
```suggestion
```

services/api/src/routers/moderation/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      const flaggedByProfileId = await getCurrentProfileId(ctx.user.id);
```
valentin0h: Better move into `submitUserFlag` so that we can run in a `Promise.all`

packages/common/src/services/moderation/submitUserFlag.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const provider = getModerationProvider();
+  const callbackUrl = getModerationCallbackUrl();
+  // Fresh round per report: encoded into the provider refs, so only this
+  // round's verdicts can land on the rows recorded below.
+  const roundId = crypto.randomUUID();
+  const [content, media] = await Promise.all([
+    resolveModerationItemText(input.itemType, input.itemId),
+    resolveModerationMedia(input.itemType, input.itemId),
+  ]);
```
valentin0h, line 78: Looks like it can all be run in `Promise.all`?

packages/common/src/services/moderation/assertModerationItemAccess.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (itemType === 'post') {
+    const [post] = await db
+      .select({
+        id: posts.id,
+        profileId: posts.profileId,
+        rootProfileId: posts.rootProfileId,
+      })
+      .from(posts)
+      .where(eq(posts.id, itemId))
+      .limit(1);
```
valentin0h: Curious if you prefer select over the relational queries. I prefer the api and syntax of `findFirst()`

packages/common/src/services/moderation/assertModerationItemAccess.ts
```diff
@@ -0,0 +1,158 @@
...
+      policies: {
+        [EntityType.DECISION]: { decisions: permission.READ },
+      },
+    });
+
+    // An already-flagged post is hidden from everyone but its author and the
+    // root-profile admins (see getPost's gate); the same audience may flag it,
+    // so a non-owner can't ship hidden content to the vendor by re-flagging.
+    if (await hasActiveModerationFlag('post', post.id)) {
+      const actorProfileId = await getCurrentProfileId(user.id);
```
valentin0h: We fetched this already upstream.

packages/common/src/services/moderation/assertModerationItemAccess.ts
```diff
@@ -0,0 +1,158 @@
...
+        user,
+        profileId: proposal.profileId,
+      });
+      const isInstanceAdmin = checkPermission(
+        { profile: permission.ADMIN },
+        instanceProfileUser?.roles ?? [],
+      );
+      if (!proposalProfileUser && !isInstanceAdmin) {
+        throw new NotFoundError('Proposal', itemId);
+      }
```
valentin0h, line 152: One more place where the profile user substitution could bite. I don't think it's an issue here, 'cause we're checking the proposal user. Just noting, looks good otherwise.

valentin0h (approved):

packages/common/src/services/moderation/assertModerationItemAccess.ts
scazan: I do tend to prefer it (always liked that drizzle was closer to SQL than Prisma or MongoDB). The newer Drizzle syntax is one I like even less unfortunately. Feels like it's moving into mongodb territory. But, whatever is clearest for the group. I reach for select when I need something specific like just a few columns. or if I have complex joins where I don't want to bring entire chunks of data over with me.

PR #1321, opened 2026-06-17
fix(posts): gate decision/proposal post reads through a fail-closed polymorphic reader

valentin0h:
Closes a fail-open leak: proposal posts were readable by any network member because assertProfileTypeAccess passes profile types absent from its policy map unchecked. Decision and proposal feeds now route through a polymorphic listProfilePosts that resolves the profile type server-side and gates each fail-closed (proposals on the parent decision).


packages/common/src/services/posts/access.ts
```diff
@@ -0,0 +1,70 @@
...
+    throw new NotFoundError('Profile', profileId);
+  }
+
+  switch (profile.type) {
+    // The READ grant lives on the decision profile itself.
+    case EntityType.DECISION: {
+      await assertProfileTypeAccess({
+        user,
+        profileIds: [profileId],
+        policies: { [EntityType.DECISION]: { decisions: permission.READ } },
```
scazan, line 38: This works, but It feels like there is some double duty here.. we have assertProfileTypeAccess for this purpose to fork out different policies per entity type and now we are doing an extra query to get the profile. Could we maybe just move the fallback logic for a proposal into the assertProfileTypeAccess instead?

scazan (approved):
Approved cause it works.. although it does feel like we are recreating assertProfileTypeAccess a bit. the boundary for the entity switch is intended to be inside of that one.

packages/common/src/services/posts/access.ts
valentin0h: Good call-out. I kept it outside `assertProfileTypeAccess` for now. that helper answers "what permission does this type need," but the proposal case is "what entity even holds the grant" (the parent decision), which is case-specific. Felt lower-risk to keep it separate.

That said, it's got me a bit suspicious of `assertProfileTypeAccess` in general. It's fail-open on unlisted types (what leaked here), and every caller just passes a single DECISION policy. Worth a closer look later.

PR #1323, opened 2026-06-17
feat(decisions): let anonymous visitors upgrade to a full account

valentin0h:
After an anonymous visitor submits an idea, offer to create a real account that keeps it — entering an email links that identity onto the existing anonymous user instead of starting a new one, so the proposal stays theirs. Promoted accounts skip org-joining during onboarding and return to their idea once done. Google identity linking is deferred to a follow-up; this round is email + OTP only.

## Demo
CleanShot 2026-06-25 at 11 54 55@2x

### DIFF
--- a/apps/app/src/app/[locale]/start/layout.tsx
+++ b/apps/app/src/app/[locale]/start/layout.tsx
@@ -1,5 +1,6 @@
 import { getUser } from '@/utils/getUser';
 import { assertWalledGardenAccess } from '@/utils/walledGarden';
+import { headers } from 'next/headers';
 
 import { Link } from '@/lib/i18n/routing';
 
@@ -9,8 +10,15 @@ import { TranslatedText } from '@/components/TranslatedText';
 const StartLayout = async ({ children }: { children: React.ReactNode }) => {
   const user = await getUser();
 
+  // Layouts don't receive searchParams, so read the query string the proxy
+  // exposes via x-search to detect the promote (anon-upgrade) onboarding. That
+  // flow is run by intentionally non-member accounts, so admit them past the
+  // membership gate while still redirecting any no-session/anonymous visitor.
+  const search = (await headers()).get('x-search') ?? '';
+  const isPromoteFlow = new URLSearchParams(search).get('promote') === '1';
+
   // Onboarding is inside the walled garden.
-  await assertWalledGardenAccess(user);
+  await assertWalledGardenAccess(user, { allowNonMembers: isPromoteFlow });
 
   return (
     <div className="relative flex h-svh w-full flex-col items-center justify-center font-sans">
--- a/apps/app/src/app/[locale]/start/layout.tsx
+++ b/apps/app/src/app/[locale]/start/layout.tsx
@@ -10,10 +10,8 @@ import { TranslatedText } from '@/components/TranslatedText';
 const StartLayout = async ({ children }: { children: React.ReactNode }) => {
   const user = await getUser();
 
-  // Layouts don't receive searchParams, so read the query string the proxy
-  // exposes via x-search to detect the promote (anon-upgrade) onboarding. That
-  // flow is run by intentionally non-member accounts, so admit them past the
-  // membership gate while still redirecting any no-session/anonymous visitor.
+  // Layouts don't get searchParams; read the query string the proxy exposes via
+  // x-search. The promote flow runs as non-members, so admit them past the gate.
   const search = (await headers()).get('x-search') ?? '';
   const isPromoteFlow = new URLSearchParams(search).get('promote') === '1';
 
--- a/apps/app/src/app/login/page.tsx
+++ b/apps/app/src/app/login/page.tsx
@@ -4,15 +4,26 @@ import { getSafeRedirectPath } from '@op/common/client';
 import { useAuthUser } from '@op/hooks';
 import { useSearchParams } from 'next/navigation';
 
+import { LinkAccountPanel } from '@/components/LinkAccountPanel';
 import { LoginPanel } from '@/components/LoginPanel';
 
 const LoginPage = () => {
   const user = useAuthUser();
   const searchParams = useSearchParams();
   const redirectParam = getSafeRedirectPath(searchParams.get('redirect'));
+  const isLinkMode = searchParams.get('link') === '1';
 
   if (!user || user.isFetching || user.isPending) {
     return null;
+  }
+
+  // Link mode: an anonymous visitor upgrading to a full account. They already
+  // have a session, so the walled-garden / signed-in paths below would bounce
+  // them. The dedicated panel links the new identity onto the anon user rather
+  // than creating a separate account. This must come before the walled-garden
+  // check below, which would otherwise send the upgrading visitor to LoginPanel.
+  if (isLinkMode && user.data?.user?.is_anonymous) {
+    return <LinkAccountPanel />;
   }
 
   // An anonymous session is not "signed in" for our purposes: the walled-garden
--- a/apps/app/src/app/login/page.tsx
+++ b/apps/app/src/app/login/page.tsx
@@ -17,11 +17,9 @@ const LoginPage = () => {
     return null;
   }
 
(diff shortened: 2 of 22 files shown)


apps/app/src/app/[locale]/start/layout.tsx
```diff
@@ -9,8 +10,13 @@ import { TranslatedText } from '@/components/TranslatedText';
 const StartLayout = async ({ children }: { children: React.ReactNode }) => {
   const user = await getUser();
 
+  // Layouts don't get searchParams; read the query string the proxy exposes via
+  // x-search. The promote flow runs as non-members, so admit them past the gate.
+  const search = (await headers()).get('x-search') ?? '';
+  const isPromoteFlow = new URLSearchParams(search).get('promote') === '1';
```
Comment (scazan) [inline apps/app/src/app/[locale]/start/layout.tsx:16]: Totally fine for now. Maybe we should just standardize to nuqs here though since they support server-side parsing as well.

apps/app/src/app/[locale]/start/layout.tsx
```diff
@@ -9,8 +10,13 @@ import { TranslatedText } from '@/components/TranslatedText';
...
+
   // Onboarding is inside the walled garden.
-  await assertWalledGardenAccess(user);
+  await assertWalledGardenAccess(user, { allowNonMembers: isPromoteFlow });
```
Comment (scazan) [inline apps/app/src/app/[locale]/start/layout.tsx:19]: What is a "member"? Usually that terms is reserved for users who belong to an organization.

apps/app/src/app/login/page.tsx
```diff
@@ -4,17 +4,26 @@ import { getSafeRedirectPath } from '@op/common/client';
...
 import { useSearchParams } from 'next/navigation';
 
+import { LinkAccountPanel } from '@/components/LinkAccountPanel';
 import { LoginPanel } from '@/components/LoginPanel';
 
 const LoginPage = () => {
   const user = useAuthUser();
   const searchParams = useSearchParams();
   const redirectParam = getSafeRedirectPath(searchParams.get('redirect'));
+  const isLinkMode = searchParams.get('link') === '1';
```
scazan, line 14: Also can use nuqs instead

apps/app/src/components/decisions/PromoteAccountModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const { user } = useUser();
+  const [promote, setPromote] = useQueryState('promote');
+  // The just-submitted proposal's profileId, so we can return the visitor to
+  // their own idea (the proposal view) after they finish creating an account.
+  const [proposalId] = useQueryState('proposal');
+
+  // Only anonymous sign-ins can be promoted; a full account has nothing to
+  // upgrade. (`isAnonymous` is session-derived, same as onboarding.ts.)
+  const isAnonymous = Boolean(user?.isAnonymous);
+  const isOpen = isAnonymous && promote === '1';
```
scazan, line 34: Does this need to be a string 1? it seems like it's a boolean masquerading as a number masquerading as a string :)

apps/app/src/components/decisions/PromoteAccountModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const close = () => {
+    void setPromote(null);
+  };
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onOpenChange={(open) => (open ? null : close())}
+      className="sm:max-w-[29rem]"
```
scazan, line 44: We should probably just fit this to a proper tailwind sizing. I would recommend one rem off here for max-w-112

apps/app/src/components/decisions/PromoteAccountModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const goToLogin = () => {
+    const base = window.location.pathname;
+    const redirect = proposalId ? `${base}/proposal/${proposalId}` : base;
+    window.location.assign(
+      `/login?link=1&redirect=${encodeURIComponent(redirect)}`,
+    );
+  };
+
+  return (
+    <div className="flex flex-col gap-6 p-8 sm:px-12 sm:pt-12">
```
scazan, line 80: We are definitely beyond DRYness here for this modal style. We should pull it out as I think we have it copy pasted in a few places for the same design now. This one uses slightly different padding than the other ones so we are seeing drift already.

apps/app/src/components/decisions/PromoteAccountModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+      {/* TODO(anon-upgrade): restore "Already have an account? Log in" once we
+          support linking an email that already belongs to a full account. */}
+    </div>
+  );
+};
+
+// Opens a policy page in a new tab without toggling the consent checkbox it
+// sits inside (stop the press from reaching the surrounding React Aria
+// Checkbox).
+const PolicyLink = ({
```
scazan, line 165: This isn't so much a PolicyLink as it is an ExternalLink that we could use in all kinds of places. Good one to pull out (and re-use in onboarding and in the Avatar menu for instance)

apps/app/src/components/MultiStepForm/index.tsx
```diff
@@ -45,9 +45,10 @@ export const MultiStepForm: React.FC<{
   const goToStep = React.useCallback(
     (targetStep: number) => {
       setStep(targetStep);
-      const query = { step: targetStep.toString() };
 
-      const params = new URLSearchParams(query);
+      // Preserve existing query params (e.g. promote flag + redirect) across steps.
+      const params = new URLSearchParams(window.location.search);
```
scazan, line 50: use nuqs here for sure (as this one is a complicated beast of a form)

apps/app/src/components/Onboarding/index.tsx
```diff
@@ -54,6 +55,13 @@ const ProgressInPortal = (props: ProgressComponentProps) => (
   </Portal>
 );
 
+// Prefetch for the org-search step. Only on the org-search branch: the promote
+// flow skips it, and prefetching there fires a network-tier 403 for the new user.
+const PrefetchDomainOrganizations = () => {
```
scazan, line 60: I think we should probably consider whether we need to do this in the end because if a user who belongs to an organization exists already, we might want to just onboard them into common entirely ( for instance, maybe they match the email domain of a current organization. ). Low priority and for the future perhaps

apps/app/src/components/Onboarding/index.tsx
```diff
@@ -282,6 +294,10 @@ export const OnboardingFlow = () => {
     return <LoadingSpinner />;
   }
 
+  if (isPromoteFlow) {
+    return <PromoteOnboardingFlow hasHydrated={hasHydrated} />;
```
scazan, line 298: I'm a little worried that trying to bolt this on this onboarding form is complicating things since this form is already quite complicated as it is. We might consider an entirely new onboarding flow for this particular use case, but with a shared set of components that we can explicitly include or omit.

apps/app/src/components/LinkAccountPanel.tsx
```diff
@@ -0,0 +1,277 @@
...
+  // onto the anon user, so that path drops link mode and goes to normal login.
+  const regularLoginHref = isSafeRedirectPath(redirectParam)
+    ? `/login?redirect=${encodeURIComponent(redirectParam)}`
+    : '/login';
+
+  // After linking, route through onboarding with the page to return to.
+  // `redirectParam` carries the locale prefix the locale-less /login route lacks.
+  const goAfterLink = useCallback(() => {
+    const dest = isSafeRedirectPath(redirectParam) ? redirectParam : '/';
+    const locale = dest.split('/')[1] || 'en';
```
scazan, line 69: Can we just do a useLocale here?

scazan (approved):
Looks good. I do worry that we are building on top of the already complex onboarding flow instead of just composing a new one.. I think that would be a lot cleaner and less prone to issues. wdyt? But not a blocker, let's get it out and we also need to take a fresh look at the whole onboarding form in the near term anyway.

PR #1324, opened 2026-06-17
feat(supabase): enable and style email-change confirmation

valentin0h:
Require email confirmation and add a template so attaching an email to an anonymous user sends a verification code. Independent of the upgrade-flow stack.

### DIFF
--- a/supabase/templates/email_change.html
+++ b/supabase/templates/email_change.html
@@ -1,0 +1,465 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html dir="ltr" lang="en">
+  <head>
+    <link
+      rel="preload"
+      as="image"
+      href="https://common.oneproject.org/email-logo.png"
+    />
+    <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
+    <meta name="x-apple-disable-message-reformatting" />
+    <meta name="color-scheme" content="light" />
+    <meta name="supported-color-schemes" content="light" />
+    <!--$-->
+    <style>
+      @font-face {
+        font-family: 'Roboto';
+        font-style: normal;
+        font-weight: 400;
+        mso-font-alt: 'Helvetica';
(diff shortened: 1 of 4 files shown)


supabase/templates/email_change.html
valentin0h, line 1: Almost identical to the `magiclink` one but with slightly different wording. I think it's better to keep them separate.

supabase/templates/email_change.html
```diff
@@ -0,0 +1,465 @@
...
+                        text-align: center;
+                        font-size: 12px;
+                        font-weight: 700;
+                        line-height: 12px;
+                        color: rgb(163, 163, 163);
+                        margin: 16px 0;
+                      "
+                    >
+                      One Project Org,<!-- -->
+                      <!-- -->2025
```
scazan: 2026?

supabase/templates/email_change.html
```diff
@@ -0,0 +1,465 @@
...
+                                        margin: 16px 0;
+                                      "
+                                    >
+                                      This code will only be valid for the next
+                                      10 minutes.<!-- -->
+                                      <br />
+                                      <!-- -->If you’re having problems, send us
+                                      an<!-- -->
+                                      <a
+                                        href="mailto:connect@oneproject.org"
```
scazan: maybe support@oneproject.org

supabase/templates/email_change.html
```diff
@@ -0,0 +1,465 @@
...
+                                                  Liberation Mono,
+                                                  Courier New,
+                                                  monospace;
+                                                font-size: 10px;
+                                                line-height: 14px;
+                                                color: rgb(82, 82, 82);
+                                                margin: 16px 0;
+                                              "
+                                            >
+                                              Co-creating Alternatives.<br />Global
```
scazan, line 250: This doesn't sound familiar. worth checking if we use this elsewhere

scazan (approved):

supabase/templates/email_change.html
valentin0h: Confirmed it's an established tagline — already used in both magiclink.html and confirm.html, so keeping it for consistency.

supabase/templates/email_change.html
valentin0h: Good catch — updated to 2026 across all three templates.

supabase/templates/email_change.html
valentin0h: Switched to support@oneproject.org across all three templates.

PR #1328, opened 2026-06-18
refactor(access): un-export getProfileAccessUser from read paths

valentin0h:
Read callers only ever consumed the caller's effective roles, yet were fetching the full profile-access user — whose identity fields can fall back to the GLOBAL_USER_PUBLIC sentinel and thus describe someone other than the requester. Route reads through a roles-only getter so that substituted identity never leaves the access module; getProfileAccessUser remains for the asserts and cache invalidators.


scazan (approved):
very nice

PR #1332, opened 2026-06-18
feat(decisions): store overview body as JSON, render via RSC

nourmalaeb:
Stacked on #1331. Combines the planned storage + render-wire steps — storing JSON requires the JSON-capable renderer to ship at the same time, or the overview page can't display newly-saved bodies.

- Overview editor persists `editor.getJSON()` (read via an editor ref to dodge stale-closure capture) instead of `getHTML()`.
- `InstanceOverview.body` + the API encoders accept `string | JSONContent`, so legacy HTML rows and new JSON rows both read/write during rollout; the input encoder keeps a size cap.
- The overview page renders the body on the **server (RSC)** via `RichTextRenderer` and threads it into `DecisionOverview` as an `aboutSlot` — prose ships as server HTML with no client JS; only embed leaves stay client islands.
- `@op/ui` `RichTextEditor` `content` widened to `Content` so it seeds from a JSON doc.
- `JSONContent` re-exported from `@op/common/client` for typing stored bodies (no new runtime dep).
- Overview round-trip tests updated: JSON body is valid, legacy HTML string still round-trips, genuinely-malformed body degrades to undefined.

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
@@ -1,8 +1,11 @@
+import { createClient } from '@op/api/serverClient';
 import type { Metadata } from 'next';
 import { getTranslations } from 'next-intl/server';
+import type { ReactNode } from 'react';
 import { Suspense } from 'react';
 
 import { DecisionOverviewSuspense } from '@/components/decisions/DecisionOverview';
+import { RichTextRenderer } from '@/components/decisions/RichTextRenderer';
 import { DecisionContentSkeleton } from '@/components/skeletons/DecisionSkeleton';
 
 import { loadDecision } from '../loadDecision';
@@ -46,9 +49,29 @@ const DecisionOverviewPage = async ({
   const { slug } = await params;
   const { instanceId } = await loadDecision(slug);
 
+  // Render the overview body on the server (RSC) from its TipTap JSON, so the
+  // prose ships as HTML with no client JS (only embed leaves are client
+  // islands). Best-effort: on failure the slot is null and the client query +
+  // error boundary in DecisionOverviewSuspense still drive the page.
+  let aboutSlot: ReactNode = null;
+  try {
+    const client = await createClient();
+    const instance = await client.decision.getInstance({ instanceId });
+    const body = instance.instanceData?.overview?.body;
+    if (body) {
+      aboutSlot = <RichTextRenderer content={body} />;
+    }
+  } catch {
+    aboutSlot = null;
+  }
+
   return (
     <Suspense fallback={<DecisionContentSkeleton />}>
-      <DecisionOverviewSuspense instanceId={instanceId} decisionSlug={slug} />
+      <DecisionOverviewSuspense
+        instanceId={instanceId}
+        decisionSlug={slug}
+        aboutSlot={aboutSlot}
+      />
     </Suspense>
   );
 };
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
@@ -1,4 +1,5 @@
 import { createClient } from '@op/api/serverClient';
+import { logger } from '@op/logging';
 import type { Metadata } from 'next';
 import { getTranslations } from 'next-intl/server';
 import type { ReactNode } from 'react';
@@ -61,7 +62,14 @@ const DecisionOverviewPage = async ({
     if (body) {
       aboutSlot = <RichTextRenderer content={body} />;
     }
-  } catch {
+  } catch (error) {
+    // Best-effort: log for observability, then fall back to null. The client
+    // suspense query + APIErrorBoundary in DecisionOverviewSuspense remain the
+    // safety net, so a server-side fetch hiccup degrades rather than 500s.
+    logger.warn('Failed to server-render decision overview body', {
+      instanceId,
+      error: error instanceof Error ? error.message : String(error),
+    });
     aboutSlot = null;
   }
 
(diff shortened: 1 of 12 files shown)


packages/ui/package.json
```diff
@@ -53,6 +53,7 @@
     "./RequiredAsterisk": "./src/components/RequiredAsterisk.tsx",
     "./StatusDot": "./src/components/StatusDot.tsx",
     "./RichTextEditor": "./src/components/RichTextEditor/index.ts",
+    "./RichTextEditor/editorConfig": "./src/components/RichTextEditor/editorConfig.ts",
```
nourmalaeb, line 59: Export `editorConfig` separately so that we can pull the config into server components without any of the client side code.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx
```diff
@@ -45,11 +53,37 @@ const DecisionOverviewPage = async ({
...
   const { instanceId } = await loadDecision(slug);
+  const { utils, queryClient } = await createServerUtils();
+
+  // One server fetch renders the body as RSC and seeds the cache the client
+  // useSuspenseQuery hydrates from (no second fetch, no divergence). Body is
+  // best-effort: on failure the cache stays empty, so the client refetches and
+  // its APIErrorBoundary drives the error UX.
+  let aboutSlot: ReactNode = null;
+  try {
+    const instance = await utils.decision.getInstance.fetch({ instanceId });
```
Comment (nourmalaeb) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/overview/page.tsx:64]: Fetching getInstance here looks redundant with the client useSuspenseQuery in DecisionOverviewContent, but it's one fetch, not two: `utils…fetch()` renders the body as RSC **and** seeds the cache the client query hydrates from. Single fetch, no server/client divergence.

services/api/src/encoders/decision.ts
```diff
@@ -193,6 +194,47 @@ export const instancePhaseDataEncoder = z.object({
...
+ */
+const MAX_OVERVIEW_BODY_SIZE = 5_000_000;
+
+/**
+ * A stored TipTap JSON doc — a `doc`-rooted object (what `editor.getJSON()`
+ * always produces). Requiring `type === 'doc'` rejects arbitrary objects at the
+ * write boundary, so a misrouted value can't silently store and then render
+ * blank; on read, a non-doc object degrades via the body's `.catch(undefined)`.
+ */
+const isRichTextDoc = (val: unknown): val is JSONContent =>
```
nourmalaeb, line 212: `isRichTextDoc` requires `type === 'doc'` (what `getJSON()` always emits) so a misrouted non-doc object is rejected at write instead of stored-then-rendered-blank. The `.catch(undefined)` is scoped to `body` (not the whole `overview`), so a malformed `body` degrades alone without dropping `headline`/`description`.

packages/common/src/services/decision/schemas/instanceData.ts
```diff
@@ -47,8 +48,12 @@ export interface PhaseInstanceData {
...
   headline?: string;
   description?: string;
-  /** Rich text body as an HTML string (TipTap getHTML output) */
-  body?: string;
+  /**
+   * Rich text body. New content is a TipTap JSON doc (`editor.getJSON()`),
+   * which the static renderer consumes directly. Legacy rows hold an HTML
+   * string (`editor.getHTML()`) until backfilled; both shapes are read.
+   */
+  body?: string | JSONContent;
```
nourmalaeb, line 56: Union, not just JSON: new content is a TipTap JSON doc, but legacy rows are HTML strings and must keep rendering (RichTextRenderer dual-reads). No backfill — the renderer's HTML fallback covers un-migrated rows.

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -6,15 +6,20 @@ import { ButtonLink } from '@op/ui/Button';
...
-
 interface DecisionOverviewProps {
   instanceId: string;
   decisionSlug: string;
+  /**
+   * The "About" body, pre-rendered on the server (RSC) from the overview's
+   * TipTap JSON via RichTextRenderer, so the prose ships as server HTML with no
+   * client JS. Null when there's no body (falls back to the plain description).
+   */
+  aboutSlot?: ReactNode;
```
nourmalaeb, line 25: `aboutSlot` is the body pre-rendered on the **server** (RSC, in page.tsx) and passed as a slot into this client component, so the prose ships as HTML with zero client JS — only LinkPreview embeds stay client islands. That's why it's a ReactNode prop, not `content`.

packages/ui/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -21,10 +22,13 @@ export const RichTextEditor = forwardRef<
...
   {
     extensions?: Extensions;
-    content?: string;
+    /** HTML string or a TipTap JSON doc (the editor seeds from either). */
+    content?: Content;
     placeholder?: string;
     onUpdate?: (content: string) => void;
     onChange?: (content: string) => void;
+    /** Emits the editor's JSON doc on every update (for JSON-stored content). */
+    onChangeJSON?: (content: JSONContent) => void;
```
nourmalaeb, line 31: Additive: `onChangeJSON` emits editor.getJSON() from the same tiptap onUpdate that already fires onChange(html). Existing HTML consumers are unaffected; only the overview editor opts in (needs JSON for the static renderer).

apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -161,25 +169,25 @@ const OverviewHero = ({
...
   }
 
   return (
     <section className="flex flex-col gap-4">
       <Header2 className="font-serif">{t('About the process')}</Header2>
-      {content ? (
-        <ProposalHtmlContent html={content} />
+      {bodySlot ? (
+        bodySlot
       ) : fallbackText ? (
```
scazan: Not usually a fan of these nesting ternaries. we should try to avoid them generally. Just making a note.

scazan (approved):

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: Good call — refactored in dd49e925. Picks the body (server slot, else plain-text fallback) via if/else into one variable, then a single guard + return. Also dropped the dead `: null` arm (the early return already guarantees one is set).

PR #1333, opened 2026-06-18
Use git worktrees

bloudermilk:
Basic setup to use Claude worktrees. Let me know if you all are already using another method here. I noticed `.trees` in the `.gitignore` already but wasn't sure how that was set up.


scazan (approved):
Great. Yeah, we use our worktrees locally under .trees.. but this is a worthy add I think

PR #1350, opened 2026-06-22
feat(decisions): add pinned resources to decision overview sidebar

nourmalaeb:
Adds a read-only **Pinned Resources** list to the decision overview sidebar, below the phase timeline.

- Surfaces the decision profile's existing resource collection(s) — no new schema or pinned flag; same data the side-panel Resources manager edits.
- Compact cards: type icon (doc/link, play for video links), title, added date.
- Server-rendered: `page.tsx` seeds the collection + resource queries into the dehydrated cache so the sidebar hydrates from server HTML (no client waterfall).
- Anonymous viewers of public decisions can read (open procedures, public READ grant).

preview of pinned resources


apps/app/src/components/decisions/DecisionOverview.tsx
```diff
@@ -111,17 +117,30 @@ function DecisionOverviewContent({
...
+              instanceId={instanceId}
+              isAdmin={instance.access?.admin}
+              decisionSlug={decisionSlug}
+            />
+          </div>
+          {/* Pinned resources sit below the phase timeline. They're a
+              non-critical sidebar extra, so a load failure quietly renders
+              nothing rather than tripping the page-level error boundary. */}
+          {profileId ? (
+            <APIErrorBoundary fallbacks={{ default: () => null }}>
```
scazan: Are we expecting this to error normally? If we are surfacing errors generally we might want to just surface something here instead. There really shouldn't be any errors from that endpoint and if there is the user might want to be given the change to retry it

apps/app/src/components/decisions/OverviewPinnedResources.tsx
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const OverviewPinnedResourcesSuspense = ({
+  profileId,
+}: {
+  profileId: string;
+}) => {
+  const t = useTranslations();
+  const [collections] = trpc.resources.collections.list.useSuspenseQuery(
+    { profileId },
+    { staleTime: STALE_TIME },
```
scazan: Any reason for a specific staleTime here on this one?

apps/app/src/components/Resources/utils.ts
```diff
@@ -15,6 +15,23 @@ export const hostnameForDisplay = (
   }
 };
 
+// Hosts whose links are a playable video — drive the Play icon on resource
+// cards. Matched against the bare hostname (with `www.` stripped) and its
+// subdomains, so `m.youtube.com` and `player.vimeo.com` both count.
+const VIDEO_HOSTS = ['youtube.com', 'youtu.be', 'vimeo.com', 'loom.com'];
```
scazan, line 21: This is a bit shaky since we'll be using iframely and video sources can come from anywhere.. as a stopgap it seems... ok.. but I don't think we should rely on a finite list of domains for determining video hosts. Can we pull it from something else in the data?

scazan (approved):

apps/app/src/components/decisions/DecisionOverview.tsx
nourmalaeb: good call — switched it to surface a small "couldn't load pinned resources" message in the section instead of silently rendering nothing. scoped to its own boundary so a failure here doesn't take down the whole overview.

apps/app/src/components/decisions/OverviewPinnedResources.tsx
nourmalaeb: no real reason — copy-pasted from ResourcesTabContent. dropped it; getInstance doesn't set one either and it's fine.

apps/app/src/components/Resources/utils.ts
nourmalaeb: yeah, agreed it's shaky. problem is the data doesn't carry a video signal today — getLinkPreview only pulls thumbnail_url and drops iframely's oembed type. left the host list as a stopgap for now and made a note to plumb that type through properly as a follow-up.

PR #1353, opened 2026-06-22
feat(decisions): hide overview CTAs and view toggle until first phase begins

nourmalaeb:
Hide the hero proposal CTAs and the Overview/Current Phase toggle when the first phase hasn't begun (today < first phase start date).

- New `hasFirstPhaseStarted` helper; missing/unparseable start date fails open (stays visible).
- `OverviewHero` CTAs gated via `showCtas`.
- `DecisionViewToggle` returns null until the first phase starts.

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/layout.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/layout.tsx
@@ -12,6 +12,7 @@ import { DecisionHeader } from '@/components/decisions/DecisionHeader';
 import { DecisionSidePanel } from '@/components/decisions/DecisionSidePanel';
 import { DecisionTranslationProvider } from '@/components/decisions/DecisionTranslationContext';
 import { DecisionViewToggle } from '@/components/decisions/DecisionViewToggle';
+import { hasFirstPhaseStarted } from '@/components/decisions/hasFirstPhaseStarted';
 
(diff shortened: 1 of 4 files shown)


apps/app/src/components/decisions/DecisionViewToggle.tsx
```diff
@@ -18,11 +23,21 @@ interface DecisionViewToggleProps {
...
   const segment = useSelectedLayoutSegment();
   const activeView = segment === 'current' ? 'current' : 'overview';
 
+  // Warm from the layout's prefetch, so this read resolves without suspending.
+  const [instance] = trpc.decision.getInstance.useSuspenseQuery({ instanceId });
+
+  if (!hasFirstPhaseStarted(instance.instanceData?.phases)) {
+    return null;
+  }
+
```
valentin0h: Better do the check in https://github.com/oneprojectorg/common/blob/dev/apps/app/src/app/%5Blocale%5D/(no-header)/decisions/%5Bslug%5D/(decision-view)/layout.tsx#L46

and pass `isActive` prop to `DecisionViewToggle`

valentin0h (approved):
Ideally, we'd drive this from the backend. Either API or RSC.

apps/app/src/components/decisions/DecisionViewToggle.tsx
nourmalaeb: Good point, done in ab0fa10 and 4cecf2d. Cleaner and simpler now.

PR #1354, opened 2026-06-23
fix(decisions): wrap overview headers and add headline character count

nourmalaeb:
- Overview headline/description inputs are now auto-growing textareas, so long text wraps instead of getting cut off
- Enter is blocked on both fields and moves focus to the next field
- Headline shows a character counter that fades in once typing
- Headline cap lowered to 50 (client + server encoder)


nourmalaeb:
Split into two stacked PRs: #1364 (wrap header fields) and #1365 (headline character count, stacked on #1364).

PR #1360, opened 2026-06-23
feat(auth): gate the walled garden at the (main) layout

valentin0h:
Stop non-members from entering the app shell and failing piecemeal on network-gated queries; gate the closed-network boundary at a single RSC chokepoint and drop the proxy login-redirect in favor of it.


nourmalaeb (approved):

apps/app/src/components/screens/ForbiddenScreen.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useTranslations } from '@/lib/i18n';
+
+/**
+ * Walled-garden 403 screen, rendered by the `forbidden.tsx` boundary.
+ */
+export const ForbiddenScreen = () => {
+  const t = useTranslations();
+
+  return (
+    <div className="flex size-full flex-col items-center justify-center gap-8">
```
scazan: This should be a component as we are duplicating it now and these styles are going to get out of sync as we update.

scazan (approved):
Looks good to me. no blockers.

apps/app/src/components/screens/ForbiddenScreen.tsx
valentin0h: Good call — extracted the shared layout into a `StatusScreen` component (8c30360d1) that both `PageError` and `ForbiddenScreen` now use, so the styles can't drift.

PR #1387, opened 2026-06-24
feat(decisions): vanity URL /[locale]/:slug for decision pages

op-bot-common:
Adds vanity URLs for decision pages: `common.oneproject.org/en/columbus` (and bare `common.oneproject.org/columbus`) now resolve to the same content as the existing `/en/decisions/columbus`. The longer canonical URL keeps working and stays canonical for now — link builders, emails, OG metadata are unchanged.

A simpler solution in advance of a more robust solution.

Two changes:

- `next.config.mjs` rewrite maps `/:locale/:slug` → `/:locale/decisions/:slug`, with a negative-lookahead that excludes the real top-level routes (`admin`, `decisions`, `org`, `profile`, `search`, `start`, `api`, `info`, `login`) so they still resolve normally. Locale alternation mirrors `SUPPORTED_LOCALES` in `@op/common/locales` — kept in sync by hand for now.
- `proxy.ts` widens the existing next-intl locale-missing redirect to run for anonymous visitors too. The branch was previously gated on `user &&`, so logged-out users at `/columbus` skipped locale detection entirely and 404ed; they now get redirected to `/en/columbus` (or the detected locale) before the rewrite fires.

slug collisions with route names are being managed manually for now (a columbus slug must exist)


apps/app/next.config.mjs
```diff
@@ -135,6 +135,16 @@ const config = {
...
     return [
       ...previewApiRewrites,
+      // Vanity URL for decision processes: `/en/columbus` resolves to the same
+      // page as `/en/decisions/columbus`. The negative lookahead excludes
+      // top-level route names so `/en/admin`, `/en/decisions`, etc. continue to
+      // hit their real pages. Locale alternation mirrors SUPPORTED_LOCALES in
+      // @op/common/locales — keep them in sync.
+      {
+        source:
+          '/:locale(en|es|fr|pt|bn|so|ar)/:slug((?!admin|decisions|org|profile|search|start|api|info|login)[^/]+)',
```
scazan: We should probably create this from our locale list instead rather than hardcoding. Also do we need to hardcode all the possible slugs here? Let's just whitelist `columbus` instead for now.

packages/common/locales.mjs
valentin0h, line 1: suggestion: can we make this the source of truth, have `locale.ts` import `SUPPORTED_LOCALES`

valentin0h (approved):
lgtm; I would add an e2e to confirm the happy path.

packages/common/locales.mjs
scazan: oh yes, thought I had done that.. updating. thanks.

PR #1388, opened 2026-06-24
feat(api): gate posts.createPost service layer and lower tier to confirmed

op-bot-common:
Lets confirmed non-network accounts comment on public decisions, by lowering `posts.createPost` from `networkAuthenticatedProcedure` to `authenticatedConfirmedProcedure` and pushing the write gate into the service layer. The original investigation (see Asana task linked below) showed a bare tier swap would widen the existing lenient pass-through on `posts.createPost` from "any network user can write on any org/individual profile" to "any confirmed account can"; this PR closes that gap by replacing the lenient `assertProfileTypeAccess` call with a new `assertPostWriteAccess` that dispatches deny-by-default on the server-resolved root profile type.

## Behavior

| Path | Gate |
|---|---|
| DECISION announcement | `{ decisions: ADMIN }` on the decision profile |
| DECISION comment / reply | `{ decisions: SUBMIT_PROPOSALS }` — public-decision global grant admits any confirmed caller |
| ORG announcement (top-level on org profile) | `{ profile: ADMIN }` with org-admin fallback (org roles live on `organizationUsers`, not `profileUsers`) — defense-in-depth; production uses `organization.createPost` |
| ORG comment / reply | `getNetworkMembership(email)` — walled-garden membership (network email domain *or* allow-list entry). Anyone in the walled garden can comment on any org post |
| Legacy `postsToOrganizations` comment | `getNetworkMembership(email)` (same as ORG comment) |
| INDIVIDUAL / USER (write) | deny — not a supported posting surface yet |

`posts.getPost` **stays on `networkAuthenticatedProcedure`** — the walled garden is the only thing that distinguishes "in-network read" from "any confirmed read", and there's no network-scope grant to fall back on at the service layer. The only service-layer change there is to fail-close `INDIVIDUAL` / `USER` profile reads so an individual-profile post can't be fetched by id even from inside the network.

Asana: https://app.asana.com/0/1213315744811204/1215960325435100


packages/common/src/services/posts/access.ts
```diff
@@ -68,3 +72,154 @@ export const assertPostReadAccess = async ({
...
+  targetProfileId,
+  rootProfileId,
+}: {
+  targetProfileId: string | null | undefined;
+  rootProfileId: string;
+}): AccessZonePermission => {
+  if (targetProfileId && targetProfileId === rootProfileId) {
+    return { decisions: permission.ADMIN };
+  }
+  return { decisions: decisionPermission.SUBMIT_PROPOSALS };
```
scazan: Worth reminding us that this was a conscious decision to only allow users with submission rights to comment. Maybe worth a refresh on whether we still think that is a good approach

packages/common/src/services/posts/access.ts
```diff
@@ -68,3 +72,123 @@ export const assertPostReadAccess = async ({
...
+          profilePermissions: { profile: permission.ADMIN },
+          orgFallbackPermissions: { profile: permission.ADMIN },
+        });
+        return;
+      }
+      await assertOnWalledGarden(user);
+      return;
+
+    // INDIVIDUAL / USER profiles aren't a supported posting surface yet.
+    default:
```
valentin0h, line 191: Are we missing an `EntityType?Proposal`  case?

valentin0h (approved):

PR #1392, opened 2026-06-24
perf(decisions): cache instance and categories in Redis

op-bot-common:
`getInstance` and `getProcessCategories` take a lot of hits per pageload, so cache the viewer-independent payload of each under the existing `decision` cache type and route every existing mutation through a new `invalidateDecisionInstance` helper that drops the instance snapshot, categories list, and submitter face-pile together. The READ gate stays outside the cache (same pattern as `listProposalSubmitters`), so a hit can never bypass authorization — user-specific access bits are recomputed per call from the cached row.

Editor reads (ProcessBuilder autosave, plus any other `forEdit` caller) skip the cache via a new `skipCache` service option exposed on `getInstanceInputSchema` / `getCategoriesInputSchema` as `forEdit`, so a writer always sees its own most recent mutation even before cross-node memcache TTLs out.

Invalidation is wired into every mutation that touches the cached row: instance updates / deletes / phase transitions / manual selections, the cron-driven `transitionMonitor`, and the proposal mutations that move the draft↔submitted count (submit / delete / update). The existing inline `invalidate({ type: 'decision', params: [id, 'submitters'] })` calls are replaced with `invalidateDecisionInstance` so the three keys can't drift.

Closes [ONE-94](mention://issue/0a5a02ef-7e84-434f-885b-266ed2bf4f17).


packages/common/src/services/decision/transitionMonitor.ts
```diff
@@ -195,6 +196,11 @@ async function advanceInstanceTransitions({
 
       result.processed++;
 
+      // Cron-driven advance: no router runs here, so invalidate the cached
+      // instance snapshot directly. The loop below re-fetches the row for the
+      // next iteration; the cache invalidation keeps external readers fresh.
+      await invalidateDecisionInstance(processInstanceId);
```
scazan, line 202: Noting the synchronous invalidation which is important for realtime channel invalidations

nourmalaeb (approved):

PR #1403, opened 2026-06-25
feat(app): expose legal pages at /tos and /privacy

op-bot-common:
Move the public Terms of Use and Privacy Policy pages from `/info/tos` and `/info/privacy` to bare `/tos` and `/privacy` so they are easy to link to externally. The pages still mount the existing `ToSContent` and `PrivacyPolicyContent` components — they live in a `(legal)` route group that shares the previous `FullScreenSplitLayout`.

Updates the proxy matcher and safe-redirect whitelist to skip the new paths, the in-app `PolicyLink` hrefs (`PromoteAccountModal`, `ToSContentShort`, `PrivacyPolicyContentShort`), and the a11y-baseline route entries.

Asana: https://app.asana.com/0/1213315744811204/1215937085805103/f


scazan:
This was unnecessary and surfaced that we already have these

PR #1405, opened 2026-06-25
feat(decisions): SEO metadata + dynamic OG image for public decision pages

nourmalaeb:
- Canonical decision page now emits description, `robots: index`, openGraph + twitter metadata (owner-aware title).
- Colocated `opengraph-image` route renders a dynamic card (decision name, owner, proposal/participant stats) over the avatar gradient or header image; logs + retries on asset-fetch failure.
- Current-phase tab overrides openGraph so the root LinkPreview image doesn't double up with the dynamic og:image.
- `metadataBase` added to the root layout so relative image URLs resolve.
- `truncateDescription` helper (code-point aware) + unit test.

Note: re-applied onto dev's restructured decision routes. Local typecheck/tests not run (no node_modules in worktree) — CI gates them.


apps/app/scripts/og-preview.mjs
```diff
@@ -0,0 +1,121 @@
+import { ImageResponse } from 'next/og.js';
+// Standalone render of the decision OG card, one PNG per avatar gradient.
+// Reads fonts/logo from disk (no server). Run from apps/app:
+//   node scripts/og-preview.mjs <outDir>
+import { readFileSync, writeFileSync, mkdirSync } from 'node:fs';
+import { join } from 'node:path';
+import { createElement as h } from 'react';
```
scazan: we don't need to make this harder to read and reason about by calling it h() :P The audience editing this code isn't likely to remember h() so much as a shortcut.. but it was also never a good idea to not just be explicit. We always prefer longer names unless it is REALLY a common shortening (like `t()` or `i` in a loop)

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts
```diff
@@ -3,6 +3,23 @@ import { CommonError } from '@op/common';
...
+/**
+ * Display name for a decision's attribution ("by X"). Prefers the steward and
+ * falls back to the owner — same rule as DecisionListItem (steward is
+ * nullable, owner is not).
+ */
+export function getDecisionAttributionName(
+  instance:
+    | {
+        steward?: { name?: string | null } | null;
+        owner?: { name?: string | null } | null;
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts]: we should just show the steward I think.. if there isn't one then we should skip the owner. If someone has explicitly removed a steward then we'd probably want to respect that. I think we are setting the steward to the owner by default when we create the process

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/metaDescription.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Collapse whitespace and clamp to a meta-description-friendly length.
+ * Code-point aware (iterates with the spread operator) so a multi-byte
+ * character straddling the limit isn't cut mid-surrogate.
+ */
+export function truncateDescription(text: string, max = 155): string {
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/metaDescription.ts]: I'm almost certain we have this function already.. can we pull it in? We had issues previously where this was defined in two different places and we had two different methods for how we truncate it. So ideally we consolidate it to use the same one that the proposal card's use.

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/page.tsx
```diff
@@ -19,11 +21,34 @@ export async function generateMetadata({
...
+    // layout, as does a private decision (loadDecision throws into the catch
+    // below). The colocated opengraph-image route supplies og:image for this
+    // page; /current re-exports it for its own segment.
     return {
-      title: decisionProfile.name
-        ? `${label} | ${decisionProfile.name}`
-        : label,
+      title: steward ? `${name} | ${steward}` : name,
+      description,
+      ...(OPURLConfig('APP').IS_PRODUCTION
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/page.tsx:40]: nice catch here

scazan (approved):
This is very cool!

apps/app/scripts/og-preview.mjs
nourmalaeb: yeah good catch. something to add to our common-agent-toolkit lessons :)

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/metaDescription.ts
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/metaDescription.ts]: ah found it! also simplifies things 👍

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/loadDecision.ts]: sounds good! simplified things

PR #1417, opened 2026-06-25
perf(decisions): single-fetch /overview (drop redundant getInstance)

nourmalaeb:
The /overview route fetched the decision instance twice — `getDecisionBySlug` (via `loadDecision`) then `getInstance`, both deep-fetching the same row. The router's `getDecisionBySlug` already returns encoded `instanceData` + per-user `access`, so `getInstance` is redundant for /overview.

## Data flow

**Before — two fetches of the same instance row:**

```mermaid
flowchart TD
  REQ["GET /overview"]
  REQ --> LD["loadDecision → getDecisionBySlug"]
  REQ --> GI["getInstance.fetch (layout + page, deduped)"]
  LD --> DB1[("DB: instance row")]
  GI --> DB2[("DB: same row, fetched again")]
  LD --> CHROME["layout header: name + access"]
  GI -. seeds cache .-> CQ["client useSuspenseQuery(getInstance)"]
  CQ --> CONTENT["DecisionOverviewContent + DecisionHeader phases"]
```

**After — one fetch, props all the way down (no getInstance):**

```mermaid
flowchart TD
  REQ["GET /overview"]
  REQ --> LD["loadDecision → getDecisionBySlug<br/>(name + access + encoded instanceData)"]
  LD --> DB1[("DB: instance row — ONE fetch")]
  LD -- props --> ALL["layout + DecisionHeader + DecisionOverviewContent"]
```

## Changes

- /overview page + `DecisionOverview` render from `loadDecision` props — no `getInstance` (server or client).
- layout computes `isActive` from `loadDecision`'s phases; drops its `getInstance.fetch`.
- /current re-seeds `getInstance` in its own page (it was relying on the layout seed).
- `DecisionHeader` takes an optional `processInstance` prop (renders from props on the decision-view route; the canonical `/decisions/[slug]` page keeps the query fallback).
- access-parity test pins `getDecisionBySlug.access == getInstance.access` for the default Admin role.

**Net:** /overview does one instance fetch instead of two — ~175ms off time-to-content and ~half the per-request instance DB work (throughput headroom for anonymous traffic on public processes).

**Trade-off:** /overview content is now server-rendered-static — it won't live-update from the realtime channel until reload. Fine for rarely-changing overview content; flag if not.

**Tests:** access-parity (passes against local DB). Overview-render/CTA component tests skipped — the gating logic is unchanged (data source moved query→prop, typecheck-guaranteed); the real risk was access parity, which the DB test covers.

### DIFF
--- a/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
+++ b/apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
@@ -1,3 +1,9 @@
+import {
+  HydrationBoundary,
+  createServerUtils,
+  dehydrate,
+} from '@op/api/server';
+import { logger } from '@op/logging';
 import type { Metadata } from 'next';
 import { getTranslations } from 'next-intl/server';
 import { Suspense } from 'react';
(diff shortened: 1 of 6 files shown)


apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
```diff
@@ -51,18 +57,33 @@ const CurrentPhasePage = async ({
   const { slug } = await params;
   const { decisionProfile, instanceId, ownerSlug } = await loadDecision(slug);
 
+  // Seed the getInstance cache the current-phase content (DecisionStateRouter)
+  // hydrates from. The (decision-view) layout no longer fetches getInstance
+  // (the overview route is single-fetch), so /current seeds its own. Best
+  // effort: on failure the client refetches under its own boundary.
+  const { utils, queryClient } = await createServerUtils();
+  await utils.decision.getInstance.fetch({ instanceId }).catch((error) => {
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx]: Let's use try/catch here

scazan (approved):

apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx
Comment (nourmalaeb) [reply apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/current/page.tsx]: good call, switched to try/catch — d9f203f4e

PR #1450, opened 2026-06-27
refactor(decisions): extract ProposalsList sub-components

nourmalaeb:
**Stack** (merge bottom → top):
- **#1450 — extract ProposalsList sub-components ← this PR**
- #1470 — fixed header + sticky bar + floating toggle
- #1449 — full-width horizontal-scrolling filter row

---

Behavior-neutral cleanup. Splits the ~250-line `ProposalsListContent` into focused pieces ahead of the sticky-filter-bar feature work.

- New `ProposalsStickyFilterBar`, `MobileViewSwitch`, `TranslationNotice` — code moved verbatim (sticky bar keeps its existing sentinel-based stuck detection).
- Memoize the revision-request lookup map.

No behavior change. Feature work stacks on top.

### DIFF
--- a/apps/app/src/components/decisions/ProposalsList.tsx
+++ b/apps/app/src/components/decisions/ProposalsList.tsx
@@ -15,27 +15,20 @@ import {
   getLocationFieldMapView,
   templateCollectsLocation,
 } from '@op/common/client';
-import { useInfiniteScroll, useIntersectionObserver } from '@op/hooks';
-import { Button } from '@op/ui/Button';
-import { Link } from '@op/ui/Link';
+import { useInfiniteScroll } from '@op/hooks';
 import { cn } from '@op/ui/utils';
 import { parseAsString, parseAsStringLiteral, useQueryState } from 'nuqs';
 import { type RefObject, useCallback, useMemo } from 'react';
-import { LuLayoutGrid, LuMap } from 'react-icons/lu';
 
-import { useTranslations } from '@/lib/i18n';
-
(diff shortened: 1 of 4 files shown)


apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -427,82 +411,35 @@ const ProposalsListContent = ({
...
       )}
     >
-      {/* Sentinel at the filter bar's pre-pin top — drives the JS "stuck"
-          detection that toggles data-stuck on the bar below. */}
-      <div
-        ref={filterSentinelRef}
-        aria-hidden
-        className="pointer-events-none absolute inset-x-0 top-0 h-px"
+      <ProposalsStickyFilterBar
+        isMapMode={isMapMode}
```
scazan, line 415: yikes.. that is a lot of props :) It feels like a sign that we need to compose out of a few components rather than a single component for all. That will also enable better re-use of the Filter bar if we want to re-use this.

scazan (approved):
That huge prop list feels like a code smell. But it would be better to allow users to compose that together (imagine if we want a non-sticky ProposalsFilterBar as well). But non-blocking for release.

apps/app/src/components/decisions/ProposalsList.tsx
nourmalaeb: Yikes is right! Good call, we should separate the filters out from the sticky behaviour. Soon!

PR #1455, opened 2026-06-28
feat(decisions): use masonry layout for proposal card grids

nourmalaeb:
- Shared `ProposalMasonry` wrapper (react-masonry-css) replaces the fixed 3-col grid across the proposals grid, results, ballot, and review-assignment lists (+ skeletons).
- Variable-height cards pack tightly while keeping the server's left-to-right sort order. Breakpoint columns derived from `@op/styles` `screens`.
- Removed proposal-title truncation and the `min-w-80` floor so cards shrink to fit the column near breakpoints.

### DIFF
--- a/apps/app/src/components/decisions/ProposalListSkeleton.tsx
+++ b/apps/app/src/components/decisions/ProposalListSkeleton.tsx
@@ -1,12 +1,14 @@
 import { Skeleton } from '@op/ui/Skeleton';
 import { Surface } from '@op/ui/Surface';
 
+import { ProposalMasonry } from './ProposalMasonry';
+
 export const ProposalListSkeletonGrid = () => (
-  <div className="grid grid-cols-1 gap-6 md:grid-cols-2 lg:grid-cols-3">
+  <ProposalMasonry>
     {Array.from({ length: 6 }).map((_, index) => (
       <ProposalCardSkeleton key={index} />
     ))}
-  </div>
+  </ProposalMasonry>
 );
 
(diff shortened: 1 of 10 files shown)


apps/app/src/components/decisions/ProposalListSkeleton.tsx
```diff
@@ -1,12 +1,14 @@
 import { Skeleton } from '@op/ui/Skeleton';
 import { Surface } from '@op/ui/Surface';
 
+import { ProposalMasonry } from './ProposalMasonry';
+
 export const ProposalListSkeletonGrid = () => (
-  <div className="grid grid-cols-1 gap-6 md:grid-cols-2 lg:grid-cols-3">
+  <ProposalMasonry>
```
scazan: Do we really need to do the masonry here? Let's just use CSS grid for this so we can SSR it.

scazan (approved):
Let's fix the Skeleton as we want that to appear as soon as possible rather than after we have loaded the masonry library on the client.

apps/app/src/components/decisions/ProposalListSkeleton.tsx
scazan: Maybe something like this: `className="columns-1 md:columns-2 lg:columns-3 gap-6 [&>*]:mb-6 [&>*]:break-inside-avoid"
`

nourmalaeb: ah right, good call

PR #1470, opened 2026-06-29
refactor(decisions): fixed header + scrolling content, sticky filter bar + floating toggle

nourmalaeb:
**Stack** (merge bottom → top):
- #1450 — extract ProposalsList sub-components
- **#1470 — fixed header + sticky bar + floating toggle ← this PR**
- #1449 — full-width horizontal-scrolling filter row

---

Builds on #1450. Addresses https://app.asana.com/1/1108348036672214/project/1213315744811204/task/1216032874355632?focus=true 

- Wrap the decision-view header + content in a grid (header fixed, content row scrolls) so the scrollbar stays below the header.
- The proposals filter bar pins inside that scroll container via a zero-height sentinel + IntersectionObserver. `pinOffset` is a prop (default 0); the decision-view passes the floating-toggle clearance, wired through every DecisionStateRouter branch.
- The floating Overview/Current toggle overhangs the header; the phase header scrolls behind it. Once the bar pins, a token white band fills behind the toggle so content stops showing through.


apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/layout.tsx
```diff
@@ -38,22 +38,30 @@ const DecisionViewLayout = async ({
...
-      />
-      {children}
+      {/* Fixed header + scrolling content: the grid keeps the header in place
+          and confines the scrollbar to the content row. The content row is the
+          scroll container the proposals filter bar pins inside. */}
+      <div className="grid h-dvh grid-rows-[auto_1fr]">
+        <DecisionHeader
+          instanceId={instanceId}
+          decisionSlug={slug}
+          isAdmin={access?.admin}
```
Comment (scazan) [inline apps/app/src/app/[locale]/(no-header)/decisions/[slug]/(decision-view)/layout.tsx:48]: This isn't due to your PR but this actually should have just been passed permissions originally. As soon as we start to have multiples of these permission props we should probably just pass permission objects.

scazan (approved):

PR #1476, opened 2026-06-30
fix(decisions): rewrite vanity /current sub-page to decision route

nourmalaeb:
Vanity decision URLs now work for all sub-pages, not just the base slug.

A single catch-all rewrite maps `/:locale/:slug/:path*` → `/:locale/decisions/:slug/:path*`, so `/en/columbus`, `/en/columbus/current`, and any future sub-page resolve to the decision route. (`:path*` matches zero-or-more segments, covering the bare slug too.)


scazan (approved):
Perfect

PR #1480, opened 2026-06-30
feat(decisions): admin-uploadable overview hero background image

nourmalaeb:
Admins can upload a hero background image for a decision's Overview page.

- `backgroundImage` added to `instanceData.overview` (JSONB, no migration) + overview encoders
- Admin-gated `uploadOverviewBackgroundImage` mutation (asserts admin → uploads to `assets` → persists path); 3MB cap on client + server to stay under Vercel's request-body limit
- New `@op/ui/BannerImageField` (empty + filled states, Choose file, trash, filename · size); shared `useOverviewBackgroundImage` hook
- Desktop "Edit banner" button on the hero; mobile shows an admin bar in the shared header that opens a bottom sheet (Edit banner / Process settings)
- `OverviewHero` renders a blurred `<Image fill>` + dark scrim with white text; gradient fallback when no image
- Access-tier gating + MIME/size/data-URL rejection tests; new strings translated into all locales (ar/bn/so best-effort)

Note: scrim/blur are first-pass design values; ar/bn/so translations worth a native-speaker pass.


apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
```diff
@@ -107,6 +129,24 @@ function OverviewSectionContent({
           />
         </div>
 
+        <BannerImageField
+          label={t('Banner image')}
+          value={bannerUrl}
+          fileName={fileName}
+          fileSizeLabel={fileSizeLabel}
+          title={t('Upload banner image')}
+          description={t('PNG or JPG · recommended 2400×800px · max 3MB')}
```
scazan: Is this all we support? We have a list of supported image types that we can make re-use of here. Why not gif or webp for instance?

services/api/src/routers/decision/instances/uploadOverviewBackgroundImage.ts
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Uploads a hero background image for a decision overview. Admin-only: the
+ * caller must have decisions ADMIN on the instance's profile. The image bytes
+ * land in the `assets` bucket and the resulting storage path is persisted into
+ * `instanceData.overview.backgroundImage` (cleared by sending an empty string
+ * to updateDecisionInstance instead).
+ */
+export const uploadOverviewBackgroundImageRouter = router({
+  uploadOverviewBackgroundImage: authenticatedConfirmedProcedure()
+    .use(withDB)
```
scazan: We never pull in the DB to the router level (or at least never should)

services/api/src/routers/decision/instances/uploadOverviewBackgroundImage.ts
```diff
@@ -0,0 +1,144 @@
...
+    .output(
+      z.object({
+        url: z.string(),
+        path: z.string(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      const { instanceId, file, fileName, mimeType } = input;
+      const { user } = ctx;
+      const { db } = ctx.database;
```
scazan: Everything below this should be in a service instead.

packages/ui/src/components/BannerImageField.tsx
```diff
@@ -0,0 +1,159 @@
...
+  };
+
+  return (
+    <div className={cn('flex w-full flex-col gap-2', className)}>
+      {label ? <p className="text-sm text-neutral-black">{label}</p> : null}
+
+      {value ? (
+        <div className="flex flex-col gap-2">
+          <div className="relative aspect-[3/1] w-full overflow-hidden rounded-lg">
+            <img
```
scazan: Can this be a NextImage so that we get the benefits of that? or we simply pass in the image which is what we do with the Avatar.

services/api/src/encoders/decision.ts
```diff
@@ -249,6 +249,7 @@ const instanceOverviewEncoder = z.object({
   // Scope the read-path degradation to `body` alone: a malformed stored body
   // becomes undefined without taking headline/description down with it.
   body: overviewBodyEncoder.optional().catch(undefined),
+  backgroundImage: z.string().optional(),
```
scazan: "backgroundImage" is an odd naming since it's not the background of the decision and will be displayed in many more places than a background. Maybe heroImage or headerImage is a better bet. I can imagine this would also want to be used in something like an OG image.

packages/common/src/services/decision/overviewImage.ts
```diff
@@ -0,0 +1,17 @@
...
+// The image rides through the tRPC body as base64. Vercel caps the serverless
+// request body at ~4.5MB and base64 inflates ~33%, so anything over ~3.3MB raw
+// is rejected by the platform (a 413 the tRPC client can't parse as JSON)
+// before our handler runs. Cap well under that.
+export const MAX_BACKGROUND_IMAGE_SIZE = 3 * 1024 * 1024;
+
+export const ALLOWED_BACKGROUND_IMAGE_MIME_TYPES: readonly string[] = [
+  'image/png',
+  'image/jpeg',
+  'image/webp',
```
scazan: You are also listing webp and gif here but our upload messaging says we only accept png or jpeg

apps/app/src/lib/i18n/dictionaries/en.json
```diff
@@ -32,6 +32,11 @@
   "Communities Served": "Communities Served",
   "Strategies/Tactics": "Strategies/Tactics",
   "Back": "Back",
+  "Banner image": "Banner image",
```
scazan: a nit here.. it's nice to keep the languages in sync in terms of order of keys so they can be easily compared.

apps/app/src/lib/i18n/dictionaries/bn.json
```diff
@@ -1367,5 +1367,15 @@
   "With an account": "অ্যাকাউন্ট দিয়ে",
   "Edit your idea before review begins, get notified when it moves to the next phase, and like, comment, and follow other ideas.": "পর্যালোচনা শুরু হওয়ার আগে আপনার আইডিয়া সম্পাদনা করুন, এটি পরবর্তী ধাপে গেলে বিজ্ঞপ্তি পান, এবং অন্যান্য আইডিয়া পছন্দ করুন, মন্তব্য করুন ও অনুসরণ করুন।",
   "Create account": "অ্যাকাউন্ট তৈরি করুন",
-  "Collapsible": "সংকোচনযোগ্য"
+  "Collapsible": "সংকোচনযোগ্য",
+  "Banner image": "ব্যানার ছবি",
+  "Upload banner image": "ব্যানার ছবি আপলোড করুন",
+  "PNG or JPG · recommended 2400×800px · max 3MB": "PNG বা JPG · প্রস্তাবিত 2400×800px · সর্বোচ্চ 3MB",
```
scazan: Let's use a constant here (it exists in the codebase already) on maximum filesize. That is displayed in the Attachment upload component as well.

apps/app/src/components/decisions/useOverviewBackgroundImage.ts
```diff
@@ -0,0 +1,135 @@
...
+};
+
+/**
+ * Shared upload/remove logic for the decision overview hero background image.
+ * Used by the Process Builder Overview tab and the live overview's "Edit
+ * banner" modal. The upload mutation persists the storage path into
+ * `instanceData.overview.backgroundImage`; `onChange` lets the live page
+ * refresh its RSC-fed hero after a change.
+ */
+export function useOverviewBackgroundImage({
```
scazan: Couldn't this be any particular image upload? We can keep it generic and do something like pass in the uploadMutation as a prop. Particularly with image uploads which tend to be a bit complex to duplicate the logic of

apps/app/src/components/decisions/useOverviewBackgroundImage.ts
```diff
@@ -0,0 +1,135 @@
...
+// Recover the original display name (basename minus the timestamp prefix).
+const fileNameFromPath = (path?: string): string | undefined => {
+  if (!path) {
+    return undefined;
+  }
+  const base = path.split('/').pop() ?? path;
+  return base.replace(/^\d+_/, '');
+};
+
+const formatFileSize = (bytes: number): string => {
```
scazan: We must already have this in the attachments upload component

apps/app/src/components/decisions/EditBannerModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <Button
+        color="secondary"
+        size="small"
+        className="absolute top-4 right-4 z-10 hidden md:flex"
+        onPress={() => setIsOpen(true)}
+      >
+        <LuImage className="size-4" aria-hidden="true" />
+        {t('Edit banner')}
+      </Button>
+      <BannerUploadModal
```
scazan, line 37: We are mixing terms a bit here.. Banner vs OverviewImage vs backgroundImage. My suggestion below is that the backend property is something like heroImage, headerImage, or something like mainImage or something closer to it's semantic meaning rather than where it is used. I'm fine with this being BannerUploadModal though as banner makes sense.

apps/app/src/components/decisions/AdminOverviewBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+        date: new Date(phaseEndDate).toLocaleDateString(locale, {
+          month: 'long',
+          day: 'numeric',
+        }),
+      })
+    : null;
+
+  return (
+    <>
+      <button
```
scazan: This looks like maybe it's meant to just be a variant of our Button component. Maybe just use that? I think we have styles already for this and if not we should think about whether we want another variant.

scazan (changes requested):
I added a few comments in that I think we should work through a bit before merge.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/OverviewSection.tsx
nourmalaeb: Good call — now accepts png/jpeg/webp/gif via the shared `ALLOWED_UPLOAD_MIME_TYPES` allowlist (image subset), and the picker copy advertises all four so they can't drift.

services/api/src/routers/decision/instances/uploadOverviewBackgroundImage.ts
nourmalaeb: Removed — the base64 router is gone. Replaced with thin `signOverviewHeroImageUploadUrl` / `updateOverviewHeroImage` routers that validate input and delegate to `@op/common` services; no DB access at the router level.

services/api/src/routers/decision/instances/uploadOverviewBackgroundImage.ts
nourmalaeb: Done — all logic now lives in the `signOverviewHeroImageUploadUrl` and `updateOverviewHeroImage` services in `@op/common`.

packages/ui/src/components/BannerImageField.tsx
nourmalaeb: Added a `renderPreview` prop so the app injects `next/image` while keeping `@op/ui` free of a `next` dep (same reason `AvatarUploader` uses a plain ``). The app wrapper `OverviewHeroImageField` passes `<Image fill>`. The transient `blob:` optimistic frame falls back to a plain `` internally (next/image can't optimize blob URLs), so the stable public URL still gets optimized.

services/api/src/encoders/decision.ts
nourmalaeb: Renamed to `heroImage` across the schema, encoders, services, and hooks.

packages/common/src/services/decision/overviewImage.ts
nourmalaeb: Resolved — `overviewImage.ts` is deleted; we now use the single shared `ALLOWED_UPLOAD_MIME_TYPES` allowlist and the copy advertises all four types, so message and code can't diverge.

apps/app/src/lib/i18n/dictionaries/en.json
nourmalaeb: Fixed — the new keys now sit in one contiguous block in the same position across every locale.

apps/app/src/lib/i18n/dictionaries/bn.json
nourmalaeb: Now enforced via a shared `IMAGE_UPLOAD_SIZE_LIMIT` (5MB) in `utils/storage.ts`, interpolated into the copy instead of a hardcoded number and used by both the server record step and the client pre-check. Set below the 25MB document default since a banner doesn't need it, and generic so other image uploads (avatars, phase banners) can reuse it. The signed-URL migration also dropped the old 3MB Vercel-body cap.

apps/app/src/components/decisions/useOverviewBackgroundImage.ts
nourmalaeb: The signed-URL migration made this moot: the complex bits (MIME allowlist, size cap, storage trust-boundary check) now live in shared `@op/common` utils reused by both proposal attachments and the hero image, so the hook is thin with no duplicated upload logic. The UI is also now a reusable `@op/ui/BannerImageField` primitive + app `OverviewHeroImageField` wrapper for the upcoming phase-banner feature.

apps/app/src/components/decisions/useOverviewBackgroundImage.ts
nourmalaeb: Gone — no more client-side base64. The hero image uses the same signed-URL flow as proposal attachments (`signStorageUploadUrl` → PUT binary → record), so there's no FileReader/base64 to duplicate.

apps/app/src/components/decisions/EditBannerModal.tsx
nourmalaeb: Cleaned up: the backend property is `heroImage`; component names stay `Banner*` per your note. Added a shared `@op/ui/BannerImageField` primitive + app `OverviewHeroImageField` wrapper on top.

apps/app/src/components/decisions/AdminOverviewBar.tsx
nourmalaeb: Switched to `<Button unstyled>` so it goes through the component (RAC press/focus handling). No existing variant matched the full-width tinted bar, so `unstyled` + custom classes rather than adding a one-off variant — happy to add a variant if you'd prefer.

apps/app/src/components/decisions/OverviewHeroImageField.tsx
```diff
@@ -0,0 +1,76 @@
...
+        ),
+        chooseFile: t('Choose file'),
+        remove: t('Remove image'),
+      }}
+      onSelectFile={upload}
+      onRemove={remove}
+      uploading={isUploading || isRemoving}
+      error={uploadError || undefined}
+      renderPreview={({ src, className }) => (
+        <Image
```
nourmalaeb, line 66: We can't import the Next `<Image />` component into our UI library, so we pass it in as a prop.

apps/app/src/components/decisions/OverviewHeroImageField.tsx
nourmalaeb, line 1: Wraps the design system component so that:
- Consumers have a much simpler API due to translations and labels all being centralized here
- We can pass an optimized Next `<Image/>` for the image preview, which can't be imported into our UI package

scazan (approved):

PR #1490, opened 2026-07-01
chore(emails): upgrade react-email to v6, restyle, and extract shared components

nourmalaeb:
Upgrade `@op/emails` to react-email v6, restyle the templates, and factor out the shared building blocks. Entirely within `services/emails/` — public API (`OPNodemailer`, `OPBatchSend`, template exports) unchanged, so callers don't move.

## react-email v4 → v6
- Consolidate `@react-email/components`, `@react-email/render`, `@react-email/tailwind` into the single `react-email` package; add `@react-email/ui` (dev). Preview server now bundles Next 16, fixing the broken local `email dev`.
- `email start` → `email dev` script.

## Tailwind config (v4)
- Add `pixelBasedPreset` so utilities emit px, not rem (email clients drop rem).
- Drop the hand-maintained `spacing` map (the preset's px scale is identical, and dynamic values like `max-w-160` now resolve).
- Tighten the type scale (xs–xl now 10/12/14/16/18px); darker `primary.teal` (#387582).

## Shared components (dedup)
- Extract `Header`, `Footnote`, and `CtaButton` — each owns its base classes and merges any caller `className`. Replaces the copy-pasted markup across all 11 templates.
- Flatten `Header/index.tsx` → `Header.tsx` for consistency.

## Dev-server rendering
- Add `PreviewProps` to every template so all 11 render with realistic sample data in `email dev`.
- Remove props-defensiveness (`?.`) from the templates — prop types are non-nullable, so the compiler already guarantees callers pass real values.

## Fixes
- Double padding: `md:p-12` stacked on the base `p-8` (base padding inlines onto the `<td>`, the `md:` class stays on the `<table>`, so they added up at ≥768px). Card is now a uniform `p-8`.
- Invitation copy; logo link/url.


services/emails/tailwind.config.ts
```diff
@@ -1,7 +1,10 @@
 import { commonColors } from '@op/core';
-import type { TailwindConfig } from '@react-email/tailwind';
+import { pixelBasedPreset, type TailwindConfig } from 'react-email';
```
nourmalaeb, line 2: this is a new helper in `react-email` v6 that has all the `spacing` presets predefined for us

services/emails/components/CtaButton.tsx
nourmalaeb, line 1: this component replaces:

```javascript
<Section className="pb-0">
  <Button
    href={disputeUrl}
    className="rounded-lg bg-primary-teal px-4 py-3 text-white no-underline hover:bg-primary-teal/90"
    style={{
      fontSize: '0.875rem',
      textAlign: 'center',
      textDecoration: 'none',
    }}
  >
    {children}
  </Button>
</Section>
```

scazan (approved):
Great

PR #1497, opened 2026-07-02
chore(decisions): remove decision_overview feature flag

nourmalaeb:
- Removes the last `decision_overview` flag gate (Overview editor tab in `useNavigationConfig`) — the tab is now always visible
- `getServerFeatureFlag` helper kept (currently unused) for future server-side flags
- Flag still needs archiving in PostHog (dashboard-side)


scazan (approved):
Let's keep the getServerFeatureFlag though!

PR #1536, opened 2026-07-06
fix(hooks): eliminate hydration mismatches in useMediaQuery and useRelativeTime (React #418)

nourmalaeb:
Fixes React minified error `#418` hydration mismatches caused by client-only values being read during the first render.

- `useMediaQuery`: reimplemented with `useSyncExternalStore` — the server snapshot is always `false` so SSR HTML matches the hydration render, while components mounted after hydration read `matchMedia` synchronously (no false-frame flash on client-side navigations)
- `useRelativeTime`: relative strings depend on the wall clock, which differs between server render and client hydration; render a deterministic compact absolute date until mounted, then swap in the live relative string

Date-formatting hydration fixes (UTC display of phase dates) are split into a separate PR for discussion.

### DIFF
--- a/packages/hooks/src/useMediaQuery.tsx
+++ b/packages/hooks/src/useMediaQuery.tsx
@@ -1,47 +1,32 @@
 'use client';
 
-import { useEffect, useState } from 'react';
+import { useCallback, useSyncExternalStore } from 'react';
 
 /**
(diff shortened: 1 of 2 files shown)


scazan (approved):
Great to squash these ones!

PR #1553, opened 2026-07-09
fix(decisions): show all proposal locations on the map

op-bot-common:
The proposals map view only plotted the pages the infinite-scroll list query had already loaded, and the scroll sentinel isn't rendered in map mode — so the map was hard-capped at the first list page. On a location-enabled process (where map is the default view) with more proposals than one page, users silently saw only a subset of pins with no way to load more.

This adds a dedicated `decision.listProposalLocations` endpoint that returns **every** located proposal in the current scope — no pagination, and none of the heavy per-proposal enrichment (documents, relationship counts) the pins/hovercards don't use. The map now drives its markers, hovercards, and marker-click navigation from that full set, while the desktop list column keeps using the paginated list data it already gets.

To guarantee the map's pins apply the *exact* same access / phase / visibility / moderation filtering as the list (so a pin can never leak a proposal the viewer isn't allowed to see), the filter-building logic is extracted from `listProposals` into a shared `resolveProposalListScope` helper that both call.

## Testing
- New `listProposalLocations.test.ts`: returns all located proposals + drops those without coordinates, does not surface a hidden proposal to a non-admin member, and access-tier gating coverage.
- Existing `listProposals` suite (49 tests) still passes after the extraction.
- `pnpm w:api typecheck` and `pnpm w:app typecheck` clean.

ONE-456


packages/common/src/services/decision/resolveProposalListScope.ts
```diff
@@ -0,0 +1,419 @@
...
+  return and(...conditions)!;
+};
+
+/**
+ * Resolves the shared access/phase/visibility/moderation scope for a proposal
+ * list query. Both `listProposals` (paginated, enriched) and
+ * `listProposalLocations` (all rows, slim) use this so their filtering — and
+ * therefore what a viewer is allowed to see — can never diverge.
+ */
+export const resolveProposalListScope = async ({
```
op-bot-common, line 169: **Start here.** The whole access / phase / visibility / moderation filter builder was lifted out of `listProposals` into this shared helper. It returns `{ instance, currentProfileId, canManageProposals, profileRoles, isEmpty, buildWhereClause }`. Both the paginated list and the new map endpoint call it, so what a viewer may see can't drift between the two. The extraction is verbatim — the 49 existing `listProposals` tests pass unchanged, which is the main evidence it's behavior-preserving.

packages/common/src/services/decision/resolveProposalListScope.ts
```diff
@@ -0,0 +1,419 @@
...
+  // Unreached phases on non-legacy instances surface zero rows from both the
+  // non-draft and draft scopes; an empty category likewise can't match. Either
+  // makes the whole scope empty, so callers should short-circuit.
+  const isEmpty = phaseScope.isEmpty || categoryIsEmpty;
+
+  // Assemble the full WHERE clause. Parameterized on the table reference so
+  // the same builder can be used for the v2 relational findMany (where Drizzle
+  // passes an aliased `table`) and the plain count query (which passes the
+  // schema table). See `buildBaseConditions` above for the same pattern.
+  const buildWhereClause = (proposalsTable: typeof proposals): SQL => {
```
op-bot-common, line 318: `buildWhereClause` is parameterized on the table ref so the same predicate works for both the relational `findMany` (aliased table) and the plain count query — same trick the original used. `isEmpty` (just above) folds the two short-circuit cases (unreached phase, empty category) into one flag.

packages/common/src/services/decision/listProposals.ts
```diff
@@ -188,127 +62,17 @@ export const listProposals = async ({
...
+  // Access/phase/visibility/moderation scope is resolved by the shared helper
+  // so this paginated read and `listProposalLocations` filter identically.
+  const {
+    instance,
+    currentProfileId,
+    canManageProposals,
+    profileRoles,
+    isEmpty,
+    buildWhereClause,
+  } = await resolveProposalListScope({ input, user });
```
op-bot-common, line 74: `listProposals` now delegates scope resolution to the shared helper and keeps only what's unique to the paginated list: cursor pagination, vote-count sorting, and the heavy per-proposal enrichment (documents, relationship counts, flags).

packages/common/src/services/decision/listProposals.ts
```diff
@@ -321,7 +85,7 @@ export const listProposals = async ({
     next: null,
   };
 
-  if (phaseScope.isEmpty && skipAccessCheck) {
+  if (isEmpty && skipAccessCheck) {
```
op-bot-common, line 88: The empty-result check is intentionally split around the cursor decode (two `if (isEmpty ...)` guards) to preserve prior error semantics: a malformed cursor still throws for authenticated callers even when the set is empty. Worth a glance to confirm the ordering matches the old code.

packages/common/src/services/decision/listProposalLocations.ts
```diff
@@ -0,0 +1,107 @@
...
+ * returns **every** matching proposal that carries a location — no pagination.
+ *
+ * The map plots one pin per located proposal, so it can't be capped by the
+ * list's page size. This reuses `resolveProposalListScope` (the exact
+ * access/phase/visibility/moderation filter `listProposals` applies) so pins
+ * never leak a proposal the viewer isn't allowed to see, and skips the heavy
+ * per-proposal enrichment (documents, relationship counts) the pins/hovercards
+ * don't use.
+ */
+export const listProposalLocations = async ({
```
op-bot-common, line 19: **New endpoint — the map's pin source.** Two deliberate choices: (a) no pagination, so it returns every located proposal and the map can't be capped by page size; (b) slim — only the columns pins + hovercards need (proposalData, submittedBy, profile), skipping the document fetches / relationship counts / vote counts that make `listProposals` expensive. It reuses `resolveProposalListScope`, so filtering is identical to the list.

packages/common/src/services/decision/listProposalLocations.ts
```diff
@@ -0,0 +1,107 @@
...
+            columns: {},
+            with: { authUser: { columns: { isAnonymous: true } } },
+          },
+        },
+      },
+      profile: true,
+    },
+  });
+
+  const proposals = rows.flatMap((proposal) => {
```
op-bot-common, line 64: `flatMap` drops rows without coordinates (drafts, un-geocoded), so only real pins come back. `proposalData` is parsed once per row and reused.

services/api/src/routers/decision/proposals/list.ts
```diff
@@ -43,4 +52,24 @@ export const listProposalsRouter = router({
 
       return allProposalsListSchema.parse(result);
     }),
+  /**
+   * Every located proposal in the instance's current scope — the map's pin
+   * source, so the map isn't capped by the list's page size.
+   */
+  listProposalLocations: openProcedure()
```
op-bot-common: Standard `openProcedure` wiring, and it registers the same `decisionProposals` realtime channel as `listProposals` — so pins refresh on the same invalidations the list already listens to (no manual invalidation needed).

packages/common/src/services/decision/schemas/proposal.ts
```diff
@@ -154,6 +154,19 @@ export const proposalListSchema = z.object({
...
 export type ProposalList = z.infer<typeof proposalListSchema>;
 
+/**
+ * Response from `decision.listProposalLocations`. Every located proposal in the
+ * instance (not just the loaded list page) so the map can plot all pins. Reuses
+ * the full `proposalSchema` shape — the heavy fields (documents, counts) are
+ * simply left unset — so the map's marker + hovercard render the same
+ * `Proposal` type the paginated list produces.
+ */
+export const proposalLocationsSchema = z.object({
```
op-bot-common, line 164: Output reuses the full `proposalSchema` (the heavy fields are simply left unset) so the map renders the same `Proposal` type the list already produces — no separate client-side shape to keep in sync with the hovercard/list-item components.

services/api/src/encoders/decision.ts
```diff
@@ -668,6 +668,21 @@ export const proposalFilterSchema = z.object({
   limit: z.number().min(1).max(100).prefault(20),
 });
 
+/**
+ * Input for `decision.listProposalLocations` — the map's pin source. Mirrors
+ * `proposalFilterSchema` minus pagination (the endpoint returns every located
+ * proposal in scope) so the map applies the same filters as the list.
+ */
+export const proposalLocationsFilterSchema = z.object({
```
op-bot-common, line 676: Input mirrors `proposalFilterSchema` minus the pagination fields (`cursor`/`limit`), so the map applies the same category / status / ballot filters the list does.

apps/app/src/components/decisions/ProposalsMapView.tsx
```diff
@@ -14,8 +15,23 @@ import { ProposalMapListItem } from './ProposalMapListItem';
...
+  status?: ProposalStatus;
+  phase?: 'results';
+}
+
 interface ProposalsMapViewProps {
+  /** Loaded list pages — drives the desktop list column (stays paginated). */
   proposals: Proposal[];
+  /** Marker source. Every located proposal the pins should plot, which can be
+   * a wider set than `proposals` (the loaded list pages). */
+  pinProposals: Proposal[];
```
op-bot-common, line 34: The map is now presentational: `pinProposals` is the marker source (can be a wider set than the loaded list pages), while `proposals` still drives the paginated desktop list column. Splitting these two is what lets the map show all pins without changing the list.

apps/app/src/components/decisions/ProposalsMapView.tsx
```diff
@@ -181,3 +200,27 @@ export function ProposalsMapView({
...
 }
+
+/**
+ * Loads every located proposal in scope from `listProposalLocations` and feeds
+ * them to the map as pins, so the map isn't capped by the list's page size.
+ * Used for the phase-scoped browse map; the results phase renders
+ * `ProposalsMapView` directly with its `listAllProposals` data so the pins
+ * match that list's (phase-agnostic) scope.
+ */
+export function ProposalsMapWithLocations({
```
op-bot-common, line 211: `ProposalsMapWithLocations` is the thin suspense loader: it fetches all locations and feeds them in as `pinProposals`. Pulled out as its own component so the results phase can render `ProposalsMapView` directly and bypass it (see the next comment in ProposalsList).

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -465,14 +476,52 @@ const ProposalsListContent = ({
...
         {isMapMode ? (
-          <ProposalsMapView
-            proposals={allProposals}
-            instanceId={instanceId}
-            slug={slug}
-            decisionSlug={decisionSlug}
-            permissions={permissions}
-            mapView={mapView}
-          />
+          phase === 'results' ? (
```
op-bot-common, line 479: **The one behavioral subtlety worth scrutiny.** The results phase lists proposals via `listAllProposals` (phase-agnostic, excludes rejected/duplicate), so here it renders `ProposalsMapView` directly with that same loaded data as pins. If it went through the current-phase locations endpoint instead, the results map would plot pins the results list doesn't show. Non-results phases use the all-locations loader below.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -465,14 +476,52 @@ const ProposalsListContent = ({
...
+              instanceId={instanceId}
+              slug={slug}
+              decisionSlug={decisionSlug}
+              permissions={permissions}
+              mapView={mapView}
+            />
+          ) : (
+            // Local boundaries keep the pin query from suspending / erroring
+            // the whole list subtree (filter bar + view toggle stay mounted).
+            <APIErrorBoundary
```
op-bot-common, line 494: Local `Suspense` + `APIErrorBoundary` around the pin query so toggling into map mode doesn't suspend or blank the whole list subtree — the filter bar and view toggle stay mounted, and a map fetch failure degrades only the map region instead of the page-level error fallback.

op-bot-common (commented):
### Reviewer's guide

The bug: the proposals map only plotted the list pages that had loaded, so a location-enabled process with more proposals than one page silently showed a subset of pins. This PR gives the map its own all-locations data source.

Suggested reading order (I've left inline notes on each):

1. `resolveProposalListScope.ts` — the filtering extracted out of `listProposals` (start here).
2. `listProposals.ts` — now delegates to that helper (behavior-preserving; 49 existing tests pass).
3. `listProposalLocations.ts` — the new slim, unpaginated endpoint.
4. `list.ts` / `schemas/proposal.ts` / `encoders/decision.ts` — the tRPC wiring.
5. `ProposalsMapView.tsx` / `ProposalsList.tsx` — the frontend, including the results-phase subtlety.
6. `listProposalLocations.test.ts` — coverage, including the no-leak test.

The key invariant to check: the map's pins apply the *exact* same access/phase/visibility/moderation filter as the list, so no pin can leak a proposal the viewer isn't allowed to see.

services/api/src/routers/decision/proposals/listProposalLocations.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const ids = result.proposals.map((p) => p.id).sort();
+    expect(ids).toEqual([located1.id, located2.id].sort());
+    // The pin needs coordinates + author, so both must survive the slim map.
+    const first = result.proposals.find((p) => p.id === located1.id);
+    expect(first?.proposalData.location).toEqual({ lat: 40.7, lng: -74 });
+    expect(first?.submittedBy).toBeDefined();
+  });
+
+  it('does not surface a hidden proposal to a non-admin member', async ({
```
op-bot-common, line 91: The security-critical test: a HIDDEN proposal that *has* coordinates must not leak a pin to a non-admin member. Together with the all-locations / drop-unlocated test above and the access-tier gating block below, this is the coverage that guards the no-leak invariant.

nourmalaeb (approved):

PR #1556, opened 2026-07-09
feat(decisions): add join/claim-account flow for public decision processes

nourmalaeb:
Adds a "Join" account-claim flow for public decision processes (viewer resolves `submitProposals` without an account via the Public role — today: `columbus`). Three surfaces, one modal:

- **Decision header** and **proposal view header** swap "Log in" for "Join" for logged-out and anonymous visitors (`JoinOrUserMenu`).
- **Proposal comments** show a "Join Common to comment on this proposal" block where the composer would be.
- `JoinAccountModal` (`?join=1`): email → OTP claim inline, then the existing `/start?promote=1` promote onboarding, returning to the page the visitor left. No-session visitors get an anonymous user minted on submit; an existing anon session (e.g. a proposal author) is upgraded in place, so their data follows the claim.

The claim calls `supabase.auth.updateUser` on the anon session and never touches `account.login` — deliberately bypassing the invite-only allowList for public processes. Claim logic is extracted from `LinkAccountPanel` into a shared `useClaimAccount` hook (the panel is refactored onto it).

Also rides along (found dogfooding): `updateUserProfile` never invalidated the cached `getMyAccount` user, so freshly claimed accounts kept their old anonymous name in "Logged in as …" and the comment-box placeholder. Now invalidated, pinned by a test.

Tests: e2e for the decision-header and proposal-page join paths + full-account denial; vitest for the cache invalidation. Copy is placeholder pending a copy pass.

Note: liking/following as a claimed account also needs #1580 (relationship mutations were closed-network-gated).

Follow-up (separate PR): move the promote onboarding steps into the modal and converge `PromoteAccountModal` on the same in-modal flow.

### DIFF
--- a/apps/app/src/components/decisions/DecisionHeader.tsx
+++ b/apps/app/src/components/decisions/DecisionHeader.tsx
@@ -30,6 +30,12 @@ interface StandardDecisionHeaderProps extends DecisionHeaderBaseProps {
   useLegacy?: false;
   /** Whether the current user can read decision updates */
   canReadUpdates?: boolean;
+  /**
+   * Public process (viewer can submit proposals without an account): the
+   * header offers "Join" (account claim) instead of "Log in" to logged-out
+   * and anonymous visitors.
+   */
+  canJoin?: boolean;
   /** Center-column content, e.g. the Overview / Current Phase toggle */
   centerSlot?: ReactNode;
   /** Whether to render the phase stepper below the header bar (default true) */
@@ -108,6 +114,7 @@ function DecisionHeaderView({
   decisionSlug,
   isAdmin,
   canReadUpdates,
+  canJoin,
   centerSlot,
   showStepper = true,
   title,
@@ -135,6 +142,7 @@ function DecisionHeaderView({
         decisionSlug={decisionSlug}
         isAdmin={isAdmin}
         canReadUpdates={canReadUpdates}
+        canJoin={canJoin}
         centerSlot={centerSlot}
         mobileAdminBar={
           showAdminControls && decisionSlug ? (
@@ -170,6 +178,7 @@ function DecisionHeaderContent(props: StandardDecisionHeaderProps) {
   return (
     <DecisionHeaderView
       {...props}
+      canJoin={props.canJoin ?? instance.access?.submitProposals === true}
       title={
         props.profileName ||
         instance.name ||
--- a/apps/app/src/components/decisions/DecisionHeader.tsx
+++ b/apps/app/src/components/decisions/DecisionHeader.tsx
@@ -30,12 +30,6 @@ interface StandardDecisionHeaderProps extends DecisionHeaderBaseProps {
   useLegacy?: false;
   /** Whether the current user can read decision updates */
   canReadUpdates?: boolean;
-  /**
-   * Public process (viewer can submit proposals without an account): the
-   * header offers "Join" (account claim) instead of "Log in" to logged-out
-   * and anonymous visitors.
-   */
-  canJoin?: boolean;
   /** Center-column content, e.g. the Overview / Current Phase toggle */
   centerSlot?: ReactNode;
   /** Whether to render the phase stepper below the header bar (default true) */
@@ -125,6 +119,12 @@ function DecisionHeaderView({
   title: string;
   phases: ProcessPhase[];
   currentStateId: string;
+  /**
+   * Public process (viewer can submit proposals without an account): the
+   * header offers "Join" (account claim) instead of "Log in" to logged-out
+   * and anonymous visitors. Derived from the instance by both variants.
+   */
+  canJoin: boolean;
   /** Stored overview hero image path, for the admin Edit-banner controls. */
   heroImagePath?: string;
 }) {
@@ -178,7 +178,7 @@ function DecisionHeaderContent(props: StandardDecisionHeaderProps) {
   return (
     <DecisionHeaderView
       {...props}
-      canJoin={props.canJoin ?? instance.access?.submitProposals === true}
+      canJoin={instance.access?.submitProposals === true}
       title={
         props.profileName ||
         instance.name ||
@@ -203,6 +203,7 @@ function DecisionHeaderFromProps(
   return (
     <DecisionHeaderView
       {...props}
+      canJoin={instance.access?.submitProposals === true}
       title={
         props.profileName ||
         instance.name ||
--- a/apps/app/src/components/decisions/JoinAccountModal.tsx
+++ b/apps/app/src/components/decisions/JoinAccountModal.tsx
@@ -56,7 +56,7 @@ export const JoinAccountModal = () => {
       onOpenChange={(open) => (open ? null : close())}
       isDismissable
       surface="flat"
-      className="sm:max-w-[29rem]"
+      className="sm:max-w-128"
     >
       <JoinAccountModalContent />
     </Modal>
--- a/apps/app/src/components/decisions/JoinAccountModal.tsx
+++ b/apps/app/src/components/decisions/JoinAccountModal.tsx
@@ -58,7 +58,7 @@ export const JoinAccountModal = () => {
 
 /** Header button that opens the modal. Reads/writes `?join` via nuqs, so any
  * mount point must sit under a Suspense boundary (useSearchParams). */
-export const JoinDecisionButton = () => {
+export const JoinDecisionButton = ({ className }: { className?: string }) => {
   const t = useTranslations();
   const [, setJoin] = useQueryState('join');
 
@@ -66,6 +66,7 @@ export const JoinDecisionButton = () => {
     <Button
       color="primary"
       size="small"
+      className={className}
       onPress={() => {
         void setJoin('1');
       }}
--- a/apps/app/src/components/decisions/JoinAccountModal.tsx
+++ b/apps/app/src/components/decisions/JoinAccountModal.tsx
@@ -58,7 +58,7 @@ export const JoinAccountModal = () => {
 
 /** Header button that opens the modal. Reads/writes `?join` via nuqs, so any
  * mount point must sit under a Suspense boundary (useSearchParams). */
-export const JoinDecisionButton = ({ className }: { className?: string }) => {
+export const JoinDecisionButton = () => {
   const t = useTranslations();
   const [, setJoin] = useQueryState('join');
 
@@ -66,7 +66,6 @@ export const JoinDecisionButton = ({ className }: { className?: string }) => {
     <Button
       color="primary"
       size="small"
-      className={className}
       onPress={() => {
         void setJoin('1');
       }}
--- a/apps/app/src/components/decisions/JoinAccountModal.tsx
+++ b/apps/app/src/components/decisions/JoinAccountModal.tsx
@@ -6,6 +6,7 @@ import {
   useClaimAccount,
 } from '@/hooks/useClaimAccount';
 import { useUser } from '@/utils/UserProvider';
+import type { CommonUser } from '@op/api/encoders';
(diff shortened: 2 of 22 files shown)


apps/app/src/hooks/useClaimAccount.ts
```diff
@@ -0,0 +1,154 @@
...
+   * `mintAnonSession` signs in anonymously first when there is no session
+   * (the Join modal's public-visitor path). LinkAccountPanel leaves it off:
+   * its route requires an anon session at render, and if that session died
+   * before submit the claim must fail loudly rather than silently linking the
+   * email onto a fresh empty account (orphaning the visitor's proposal).
+   */
+  const requestEmailCode = useCallback(
+    async (
+      email: string,
+      { mintAnonSession = false }: { mintAnonSession?: boolean } = {},
```
nourmalaeb, line 96: **Start here — the core mechanic.** The claim is `updateUser({ email })` + `verifyOtp(type: email_change)` on the visitor's *existing* anonymous session, so everything the anon user owns (their draft proposal, comments) survives the upgrade — there is no ownership transfer anywhere because the user row never changes. `mintAnonSession` is opt-in and only the Join modal passes it: LinkAccountPanel's route guarantees an anon session at render, and if that session dies before submit the claim must fail loudly — silently minting a fresh user would link the email onto an empty account and orphan the visitor's proposal. Also deliberate: this path never touches `account.login`, so the invite-only allowList does not apply — that is the feature for public processes.

apps/app/src/hooks/useClaimAccount.ts
```diff
@@ -0,0 +1,154 @@
...
+    ): Promise<ClaimEmailResult> => {
+      const { data: sessionData } = await supabase.auth.getSession();
+      // A full account has nothing to claim — updateUser here would start a
+      // real email change on it. Reachable via a stale-cache Join button
+      // (cross-tab sign-in within getMyAccount's staleTime).
+      if (sessionData.session && !sessionData.session.user.is_anonymous) {
+        return { ok: false, alreadySignedIn: true };
```
nourmalaeb, line 103: Guards a stale-tab race: sign in as a full account in tab A, and tab B can keep showing Join for up to getMyAccount's 30s staleTime. Without this check, submitting an email there would start a real email *change* on the full account (confirmation mail to both addresses). The session read is authoritative where the cached user is not.

apps/app/src/components/decisions/JoinAccountModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+ * anonymous visitors; a full account never sees the Join button and the modal
+ * won't open for one.
+ */
+
+/**
+ * Who may claim: logged-out visitors and anonymous accounts. NOT the same as
+ * `!userCanInteract` — a full account without a currentProfile must see the
+ * user menu, not Join.
+ */
+export const isJoinEligible = (user: CommonUser | null | undefined): boolean =>
```
nourmalaeb, line 40: Deliberately not `!userCanInteract`: that would also be true for a full account with no `currentProfile`, which would surface a Join button that dead-ends in the already-signed-in guard above. `userCanInteract` answers "may this user write content"; this answers "is there no real account to claim from". ProposalComments uses both, for those two different decisions.

apps/app/src/components/decisions/ProposalView.tsx
```diff
@@ -268,6 +268,11 @@ export function ProposalView({
       isLoading={isLoading}
       editHref={editHref}
       canEdit={canEdit}
+      // Same viewer-access bit the comments prompt reads (getProposal mirrors
+      // the decision profile's SUBMIT_PROPOSALS grant onto proposal.access),
+      // so the Join button, the modal mount, and the prompt can't diverge —
+      // on any route that renders a proposal, including the legacy one.
+      canJoin={currentProposal.access?.submitProposals === true}
```
nourmalaeb, line 275: `canJoin` is derived from the same `proposal.access` bit the comments prompt reads (getProposal mirrors the decision profile's SUBMIT_PROPOSALS grant), rather than threading `instance.access` from the route. That keeps the header button, the modal mount, and the comments prompt provably in sync on every route that renders a proposal — an earlier draft keyed them to two different access objects, which left the legacy /profile route with a prompt whose Join button opened nothing. For non-public (incl. all legacy) processes the bit is false for visitors, so nothing changes there.

services/api/src/routers/account/updateUserProfile.ts
```diff
@@ -18,6 +19,14 @@ const updateUserProfile = router({
         user,
       });
 
+      // getMyAccount serves the cached user (with profile/currentProfile
+      // names); without this, "Logged in as" and the comment-box placeholder
+      // keep the old name until the cache expires.
+      await invalidate({
```
nourmalaeb, line 25: Ride-along fix with one subtlety: the `await` matters. UpdateProfileForm calls `utils.account.getMyAccount.invalidate()` the moment the mutation resolves, so the client refetch races this server-side purge — fire-and-forget would re-serve the stale cached user. Matches the awaited pattern in completeOnboarding and switchProfile; pinned by the warm-then-mutate test in updateUserProfile.test.ts.

nourmalaeb (commented):
Reviewer guide — one comment per decision worth scrutiny, starting in `useClaimAccount.ts`.

apps/app/src/components/decisions/DecisionHeader.tsx
```diff
@@ -170,6 +178,7 @@ function DecisionHeaderContent(props: StandardDecisionHeaderProps) {
   return (
     <DecisionHeaderView
       {...props}
+      canJoin={props.canJoin ?? instance.access?.submitProposals === true}
```
scazan: Do we need the prop if we are already checking here?

apps/app/src/components/decisions/JoinAccountModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+  const close = () => {
+    void setJoin(null);
+  };
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onOpenChange={(open) => (open ? null : close())}
+      isDismissable
+      className="sm:max-w-[29rem]"
```
scazan: not for this pr but we shouldn't need these specific rem values anymore with tailwind

apps/app/src/components/decisions/JoinAccountModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+  const submitToken = async () => {
+    if (!token || isSubmitting) {
+      return;
+    }
+
+    setIsSubmitting(true);
+    setError(undefined);
+
+    try {
+      const result = await verifyEmailCode({ email, token });
```
scazan, line 191: I'm curious how much this differs from the other OTP flows and wondering if we can extract this into a shared utility.

apps/app/src/components/decisions/JoinAccountModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+  return (
+    <div className="relative flex flex-col gap-6 p-8 text-center sm:px-12 sm:py-12">
+      <IconButton
+        size="small"
+        aria-label={t('Close')}
+        onPress={onClose}
+        className="absolute end-4 top-4"
+      >
+        <LuX className="size-5" aria-hidden />
```
scazan: Isn't this just recreating the type of model that we use elsewhere with the Close X built in?

apps/app/src/components/decisions/JoinAccountModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+      {otpSent ? (
+        <div className="flex flex-col gap-3 text-start">
+          <AuthCodeField
+            value={token}
+            isDisabled={isSubmitting}
+            onChange={setToken}
+            onSubmit={submitToken}
+          />
+          <Button
+            className="flex w-full items-center justify-center"
```
scazan: This is also a bit odd as I think that justify center and item center is likely already built into the button.

apps/app/src/components/decisions/ProposalComments.tsx
```diff
@@ -71,6 +80,21 @@ export function ProposalComments({
           </div>
         )}
 
+        {showJoinPrompt && (
+          <div className="mb-8">
+            {/* sm:p-4 matches the composer Surface this slot-replaces. */}
+            <Surface className="flex flex-col items-center gap-3 border-0 p-0 text-center sm:border sm:p-4">
```
scazan: I think we have this component elsewhere. We probably don't need to reinvent the class names here. I'm thinking we probably used the same type of style on the empty state and likely we use it for various other things. We should either create a variant on the surface or we should create a component I think.

apps/app/src/components/decisions/ProposalViewLayout.tsx
```diff
@@ -142,14 +149,23 @@ export function ProposalViewLayout({
               <Tooltip>{revisionRequestLabel}</Tooltip>
             </TooltipTrigger>
           )}
-          <div className="hidden gap-4 sm:flex">
+          <div className="hidden sm:flex">
             <LocaleChooser />
```
scazan, line 153: Without checking visually. This looks like it's changing the structure a little bit. I'm wondering if maybe we want to either remove the wrapping div or that the new component should live in here as well. Just something to check on both the public proposal view as well as the logged in proposal view.

apps/app/src/hooks/useClaimAccount.ts
```diff
@@ -0,0 +1,154 @@
...
+          return { ok: false, message: error.message };
+        }
+        // The new session isn't reflected in the cached account query.
+        // Fire-and-forget: the refetch only matters if the modal is abandoned,
+        // and both success paths end in a full navigation anyway.
+        void utils.account.getMyAccount.invalidate();
+      }
+
+      // TODO(anon-upgrade): updateUser fails if this email already belongs to
+      // another account; we surface the raw Supabase error for now.
```
scazan: Let's surface a friendlier error (that can be localized) instead.

scazan (approved):
Wondering if there was some copy/paste from the other login screens here and thinking we should pull those out into components. Some comments left as well on the errors when a user already has an account.

apps/app/src/components/decisions/JoinAccountModal.tsx
nourmalaeb: good catch

apps/app/src/components/decisions/ProposalViewLayout.tsx
nourmalaeb: yeah this is a little clunky but we want to hide the locale chooser on mobile (according to designs in figma) so this does that.

apps/app/src/components/decisions/ProposalComments.tsx
nourmalaeb: Good call — you meant EmptyState. Switched the join prompt to it in 6859cb8a3: standard icon circle + centered copy instead of the hand-rolled Surface classnames.

apps/app/src/components/decisions/JoinAccountModal.tsx
nourmalaeb: Differs in kind — verifyOtp({ type: 'email_change' }) on an existing anonymous session vs the sign-in flows, with different copy and terminal states. The code-field/submit/back shell is genuinely similar though; AuthCodeField/AuthEmailField are already shared, and extracting the full OTP panel is a good follow-up.

apps/app/src/hooks/useClaimAccount.ts
nourmalaeb: Done in ae7714db8 — email_exists now maps to localized copy, with a localized generic fallback otherwise (raw error still goes to the console for debugging).

PR #1558, opened 2026-07-10
fix(decisions): enable infinite scroll for the map view proposal list

nourmalaeb:
The list column next to the map never loaded past its first page — the infinite-scroll sentinel was gated behind `!isMapMode`, and even when rendered there, the observer never attached because the map subtree mounts behind a Suspense boundary.

- `useIntersectionObserver` now holds the observed node in state via a callback ref, so the observer attaches whenever the element actually mounts (late-mounting sentinels included) and detaches on unmount. It also resets `isIntersecting` on disconnect — a frozen `true` previously re-fired `fetchNextPage` on re-enable — and reads the newest batched observer entry instead of the oldest.
- `ProposalsList` defines one shared scroll sentinel for both views: grid mode renders it below the grid (unchanged), map mode passes it into `ProposalsMapView`'s list column via a new `listFooter` slot, so loading more grows the list under the sticky map. Mobile map (no list) never renders it.
- The hook's ref is now typed `RefCallback<T>`, deleting the `as React.RefObject<HTMLDivElement>` casts at all six call sites.
- New e2e test covers map-mode infinite scroll end to end (location-field template, Suspense-mounted sentinel, 51 → 55 cards).

Inline comments below walk through the decisions worth reviewer attention.

### DIFF
--- a/apps/app/src/components/Profile/ProfileFeed/index.tsx
+++ b/apps/app/src/components/Profile/ProfileFeed/index.tsx
@@ -12,7 +12,7 @@ import { useInfiniteScroll } from '@op/hooks';
 import { HorizontalList, HorizontalListItem } from '@op/ui/HorizontalList';
 import { SkeletonLine } from '@op/ui/Skeleton';
 import { cn } from '@op/ui/utils';
-import { Fragment, type RefObject, useCallback } from 'react';
+import { Fragment, type RefCallback, useCallback } from 'react';
 
 import {
   DiscussionModalContainer,
@@ -32,7 +32,7 @@ export type ProfileFeedRenderProps = {
   posts: PostToOrganization[];
   isEmpty: boolean;
   user: CommonUser;
-  infiniteScrollRef: RefObject<HTMLElement | null>;
+  infiniteScrollRef: RefCallback<HTMLElement>;
   shouldShowTrigger: boolean;
   isFetchingNextPage: boolean;
   handleReactionClick: (postId: string, emoji: string) => void;
@@ -154,7 +154,7 @@ export const ProfileFeedCards = ({
         )}
         {shouldShowTrigger && (
           <HorizontalListItem>
-            <div ref={infiniteScrollRef as React.RefObject<HTMLDivElement>}>
+            <div ref={infiniteScrollRef}>
               {isFetchingNextPage ? (
                 <div className="text-sm text-neutral-gray4">
                   <SkeletonLine lines={2} />
@@ -211,10 +211,7 @@ export const ProfileFeedList = ({
         />
       </PostFeed>
       {shouldShowTrigger && (
-        <div
-          ref={infiniteScrollRef as React.RefObject<HTMLDivElement>}
-          className="flex justify-center py-4"
-        >
+        <div ref={infiniteScrollRef} className="flex justify-center py-4">
           {isFetchingNextPage ? (
             <div className="text-sm text-neutral-gray4">
               <SkeletonLine lines={2} />
--- a/apps/app/src/components/decisions/ProposalsList.tsx
+++ b/apps/app/src/components/decisions/ProposalsList.tsx
@@ -435,6 +435,20 @@ const ProposalsListContent = ({
 
   const hideFilters = !!proposalsHidden && !canManageProposals;
 
+  // One sentinel definition for both views — map mode renders it inside the
+  // list column (via listFooter), grid mode below the grid. Only the loading
+  // skeleton differs.
+  const renderScrollSentinel = (skeleton: React.ReactNode) =>
+    shouldShowTrigger ? (
+      <div
+        ref={infiniteScrollRef}
+        className="py-4"
+        data-testid="proposals-infinite-scroll-sentinel"
+      >
+        {isFetchingNextPage ? skeleton : null}
+      </div>
+    ) : null;
+
   return (
     <div
       className={cn(
@@ -490,17 +504,7 @@ const ProposalsListContent = ({
               decisionSlug={decisionSlug}
               permissions={permissions}
               mapView={mapView}
-              listFooter={
-                shouldShowTrigger ? (
-                  <div
-                    ref={infiniteScrollRef}
-                    className="py-4"
-                    data-testid="proposals-infinite-scroll-sentinel"
-                  >
-                    {isFetchingNextPage ? <ProposalCardSkeleton /> : null}
-                  </div>
-                ) : null
-              }
+              listFooter={renderScrollSentinel(<ProposalCardSkeleton />)}
             />
           ) : (
             // Local boundaries keep the pin query from suspending / erroring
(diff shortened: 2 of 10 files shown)


packages/hooks/src/useIntersectionObserver.tsx
```diff
@@ -23,13 +23,19 @@ export const useIntersectionObserver = <T extends HTMLElement = HTMLElement>(
     initialIsIntersecting = false,
   } = options;
   const [isIntersecting, setIsIntersecting] = useState(initialIsIntersecting);
-  const ref = useRef<T>(null);
+  // The observed node is state (set via callback ref), not a ref, so the
+  // observer re-attaches when the element mounts later than this hook's
+  // effect — e.g. a sentinel inside a Suspense boundary that resolves after
+  // the first commit.
+  const [node, setNode] = useState<T | null>(null);
```
nourmalaeb, line 30: **Start here — the root cause.** The old implementation attached the observer in an effect that read `ref.current`, with deps `[threshold, rootMargin, enabled]`. Any element that mounts *after* that effect runs — like the map view's scroll sentinel, which sits behind the Suspense boundary around the pin query — was never observed, so the map list column was permanently stuck at page 1. Holding the node in state via a callback ref makes the effect re-run when the element actually mounts (and detach when it unmounts). This is the standard callback-ref pattern; the identity-stable `setNode` means React never re-invokes it on ordinary renders, so there's no observer churn.

packages/hooks/src/useIntersectionObserver.tsx
```diff
@@ -23,13 +23,19 @@ export const useIntersectionObserver = <T extends HTMLElement = HTMLElement>(
...
 
   useEffect(() => {
-    if (!enabled || !ref.current) return;
+    if (!enabled || !node) return;
 
     const observer = new IntersectionObserver(
-      ([entry]) => {
+      (entries) => {
+        // Entries are chronological; only the newest reflects current state.
+        const entry = entries[entries.length - 1];
```
nourmalaeb, line 38: Observer callbacks can deliver several batched entries (chronological). The old `([entry])` destructure read the *oldest* — a layout shift that pushes the sentinel in-then-out within one batch would strand `isIntersecting` on a stale value. Newest entry reflects current state.

nourmalaeb (commented):
Reviewer guide below — one comment per decision worth scrutiny, starting in `useIntersectionObserver.tsx`.

packages/hooks/src/useIntersectionObserver.tsx
```diff
@@ -40,12 +46,20 @@ export const useIntersectionObserver = <T extends HTMLElement = HTMLElement>(
...
 
-    observer.observe(ref.current);
+    observer.observe(node);
 
     return () => {
       observer.disconnect();
+      // Without a live observer nothing can correct a frozen `true`, and a
+      // stale intersecting state re-triggers consumers (e.g. a spurious
+      // fetchNextPage) the moment they re-enable.
+      setIsIntersecting(initialIsIntersecting);
```
nourmalaeb, line 56: **The one behavioral subtlety worth scrutiny.** Once the observer disconnects (sentinel unmounts, or `enabled` flips false when `hasNextPage` exhausts), nothing can ever correct a frozen `isIntersecting = true`. Two concrete failure modes without this reset: (a) filter change re-enables the query → `useInfiniteScroll`'s effect fires on the stale `true` and fetches page 2 while the user sits at the top of a fresh list; (b) user is at the bottom of the grid with a fetch in flight and switches to mobile map (which renders no sentinel) → every `isFetchingNextPage` flip re-triggers on the stale `true`, silently draining every remaining page. Resetting in the effect cleanup covers unmount, disable, and node-swap in one place.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -432,6 +435,20 @@ const ProposalsListContent = ({
 
   const hideFilters = !!proposalsHidden && !canManageProposals;
 
+  // One sentinel definition for both views — map mode renders it inside the
+  // list column (via listFooter), grid mode below the grid. Only the loading
+  // skeleton differs.
+  const renderScrollSentinel = (skeleton: React.ReactNode) =>
```
nourmalaeb, line 441: One sentinel definition shared by both views — same ref wiring, padding, and testid, so grid and map can't drift. Placement is the point: grid mode renders it below the grid (unchanged), map mode passes it into the list column via `listFooter`. Inside the column, loading more grows the list while the map stays sticky; the previous below-the-grid placement (gated behind `!isMapMode`) is why map mode had no infinite scroll at all.

apps/app/src/components/decisions/ProposalsMapView.tsx
```diff
@@ -42,6 +42,10 @@ interface ProposalsMapViewProps {
   /** Fallback camera — the process's default view, used only when no proposal
    * has a location to fit. */
   mapView: MapDefaultView;
+  /** Rendered after the last list item on desktop — hosts the infinite-scroll
+   * sentinel inside the list column so loading more never adds space below
+   * the sticky map. Mobile (map only, no list) never renders it. */
+  listFooter?: React.ReactNode;
```
nourmalaeb, line 48: Plain `ReactNode` slot keeps this component presentational — it doesn't know about pagination, it just renders whatever the list owner hands it after the last item. The mobile branch early-returns before the list and never renders the slot, so there's no off-screen sentinel fetching pages nobody can see (mobile pins already come from the all-locations query and don't need the list to paginate).

apps/app/src/components/Profile/ProfileFeed/index.tsx
```diff
@@ -32,7 +32,7 @@ export type ProfileFeedRenderProps = {
   posts: PostToOrganization[];
   isEmpty: boolean;
   user: CommonUser;
-  infiniteScrollRef: RefObject<HTMLElement | null>;
+  infiniteScrollRef: RefCallback<HTMLElement>;
```
nourmalaeb, line 35: Fallout from the hook's ref type change, here and in 5 sibling files (AllOrganizations, AllDecisions, ProfileUsersAccess, LandingScreen/Feed, ProposalsList): the hook now returns a properly typed `RefCallback<T>`, so every `as React.RefObject<HTMLDivElement>` cast at the call sites is deleted and the ref goes straight onto the element. No behavior change — these were type-level casts only.

tests/e2e/tests/proposal-listing-infinite-scroll.spec.ts
```diff
@@ -96,4 +96,105 @@ test.describe('Proposal Listing — Infinite Scroll', () => {
...
   });
+
+  /**
+   * Map browse mode renders the sentinel inside the list column, and that
+   * column mounts late — behind the Suspense boundary around the map's pin
+   * query — so this guards the callback-ref observer attach in
+   * `useIntersectionObserver` (a ref-object observer set up before the
+   * sentinel mounts never fires; see the map-view infinite scroll fix).
+   */
+  test('map view list column loads remaining proposals when scrolling', async ({
```
nourmalaeb, line 107: Regression test for the exact bug this PR fixes. The grid-mode test above passed both before and after the fix — grid sentinels mount in the first commit, so they never hit the late-mount path. This variant uses a location-field template, which (a) makes map browse the default view and (b) puts the sentinel behind the pin query's Suspense boundary, forcing the late-mount attach. Asserts page 1 lands at exactly PAGE_LIMIT, then all 55 after scrolling the list column.

scazan (approved):

PR #1578, opened 2026-07-14
fix(onboarding): accept scheme-less website URLs on org create form

scazan:
## Problem

On the organization create form, entering a scheme-less website like `venuecms.com` silently blocked advancing to the next step — no error was shown. Prefixing the value with `https://` worked around it.

## Root cause

Our `zodUrl` validator is intentionally lenient: it accepts a bare domain and auto-prefixes `https://`. The website field, however, renders as `<input type="url">` via a React Aria `TextField`, and React Aria Components default to `validationBehavior="native"`.

With native behavior, the browser's built-in URL constraint applies. A scheme-less `venuecms.com` fails that constraint (`typeMismatch` — the same reason `new URL('venuecms.com')` throws), so form submission was blocked **before** the value ever reached our zod schema. Because the app surfaces errors through TanStack/zod (which never ran), no error message appeared. Typing `https://venuecms.com` satisfies native validation, so it went through.

## Fix

Add `validationBehavior="aria"` to the website field so our zod schema is the single source of truth and native URL validation no longer blocks submission. `type="url"` is kept for the mobile-keyboard/semantic benefit. This matches the existing pattern already used in `ProcessSurveyModal` and `CustomFormModal`.

Applied to all three org forms with the same `type="url"` website field:

- `Onboarding/shared/OrganizationFormFields.tsx` (the onboarding create-org flow)
- `Profile/ProfileDetails/CreateOrganizationForm.tsx`
- `Profile/ProfileDetails/UpdateOrganizationForm.tsx`

## Tests

Added `packages/common/src/utils/validation.test.ts` documenting the contract (schema accepts bare domains and prefixes `https://`) and the root-cause mismatch (native URL validation rejects what the schema accepts). All tests pass; `pnpm w:app typecheck` is clean.

### DIFF
--- a/apps/app/src/components/Onboarding/PersonalDetailsForm.tsx
+++ b/apps/app/src/components/Onboarding/PersonalDetailsForm.tsx
@@ -372,7 +372,12 @@ export const PersonalDetailsForm = ({
               errorMessage={getFieldErrorMessage(field)}
               inputProps={{
                 placeholder: t('Enter your website URL'),
-                type: 'url',
+                // Not `type="url"`: our zodUrl validation accepts a bare domain
+                // (e.g. "venuecms.com") and auto-prefixes `https://`, but the
+                // browser's native URL validation rejects the scheme-less value
+                // and silently blocks form submission. `inputMode` keeps the
+                // URL-optimized keyboard without that native constraint.
+                inputMode: 'url',
               }}
             />
           )}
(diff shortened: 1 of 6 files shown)


nourmalaeb (changes requested):
Also needed on the `apps/app/src/components/Onboarding/PersonalDetailsForm.tsx` url field, not sure if there are others as well.

nourmalaeb (approved):

PR #1580, opened 2026-07-14
fix(profile): open like/follow to confirmed out-of-network accounts

nourmalaeb:
`profile.addRelationship` / `removeRelationship` (like/follow) used `networkAuthenticatedProcedure` (closed-network: `@oneproject.org` or allowList), so accounts claimed from public decision processes — which deliberately bypass the allowList — got a 403 on every like/follow while comments (lower tier) worked.

Per review, the generic profile endpoints are left untouched (still closed-network — org/individual relationships stay in-network only). Instead, proposal engagement gets its own endpoints:

- `decision.addProposalRelationship` / `decision.removeProposalRelationship`, on `authenticatedConfirmedProcedure`: confirmed non-anonymous accounts may like/follow proposals; anonymous and no-JWT callers still fail the gate.
- They accept **only** proposal targets (`NotFoundError` otherwise) and gate on `SUBMIT_PROPOSALS` on the parent decision (`assertProposalEngagementAccess`) — the same bar commenting uses. Whatever a claimed account can comment on, it can also like — and nothing more.
- The proposal UI (`useRelationshipMutations`, used by proposal view/cards) now calls these; org/individual follow UI keeps using the network-tier profile endpoints.

Gating tests cover both new endpoints (userJwt admits); integration tests assert a submit-capable caller may like a proposal, an unrelated confirmed account may not, and a non-proposal target is rejected.

Found investigating QA on #1556.


services/api/src/routers/profile/relationships.ts
```diff
@@ -85,21 +85,34 @@ const getRelationshipsInputSchema = z.object({
...
+const relationshipProcedure = authenticatedConfirmedProcedure({
   rateLimit: { windowSize: 10, maxRequests: 20 },
 });
 
 export const profileRelationshipRouter = router({
   addRelationship: relationshipProcedure
     .input(relationshipInputSchema)
     .mutation(async ({ input, ctx }) => {
-      const { targetProfileId, relationshipType, pending } = input;
+      const { targetProfileId, relationshipType } = input;
```
nourmalaeb: Gate runs before the insert; no-op for any non-proposal target. Proposal targets require SUBMIT_PROPOSALS on the parent decision — the same bar commenting already uses (assertPostWriteAccess walks a proposal comment up to its decision the same way) — so whatever an account can comment on, it can also like, and nothing more.

nourmalaeb (commented):
Follow-up from review: the tier demotion needed a replacement target guard. Two comments on the new gate.

nourmalaeb (commented):
Update: reworked the target guard after review discussion. The earlier version added a per-proposal draft/hidden/flagged visibility check (and extracted getProposal's gate to share it), which made likes *stricter* than comments — comments only require SUBMIT_PROPOSALS on the parent decision, with no per-proposal check. That inconsistency wasn't justified: liking is a lighter action than commenting, so it shouldn't be held to a higher bar. The guard now matches the comment path exactly (SUBMIT_PROPOSALS on the decision, via assertProposalEngagementAccess), and getProposal is left untouched. If we later decide engagement should require per-proposal visibility, that's a uniform change across comments and likes, not a like-only special case.

services/api/src/routers/profile/relationships.ts
```diff
@@ -95,6 +100,13 @@ export const profileRelationshipRouter = router({
     .mutation(async ({ input, ctx }) => {
       const { targetProfileId, relationshipType, pending } = input;
 
+      // Proposal targets are gated like proposal comments (SUBMIT_PROPOSALS on
+      // the parent decision); no-op for any non-proposal target.
+      await assertProposalEngagementAccess({
+        user: ctx.user,
```
scazan: This opens up the addRelation for non-proposal profiles to allow anyone to add relationships between orgs and individuals. In-network this is open for anyone to do but we don't want to expose this to non-network users. At the very least we should gate this on like/follow explicitly. I'd suggest a new endpoint for likes and follows so we can treat them with different permissions.

scazan (changes requested):
This opens up a security issue. Let's separate this one out into it's own endpoint so we can handle that differently

services/api/src/routers/profile/relationships.ts
nourmalaeb: Done in 1e9339595 — went with the new-endpoint approach. `profile.addRelationship`/`removeRelationship` are reverted to `networkAuthenticatedProcedure` untouched (org/individual relationships stay in-network only), and proposal like/follow moved to new `decision.addProposalRelationship`/`removeProposalRelationship` on the confirmed tier. Those accept only proposal targets (NotFound for any other profile) and gate on SUBMIT_PROPOSALS on the parent decision, same as commenting. The proposal UI now calls the new endpoints; the org follow UI keeps the network-tier ones.

services/api/src/routers/decision/proposals/addRelationship.ts
```diff
@@ -0,0 +1,66 @@
+import {
```
scazan, line 1: We should either name the file after the function or name the function after the file.

services/api/src/routers/decision/proposals/addRelationship.ts
```diff
@@ -0,0 +1,66 @@
...
+   */
+  addProposalRelationship: authenticatedConfirmedProcedure({
+    rateLimit: { windowSize: 10, maxRequests: 20 },
+  })
+    .input(proposalRelationshipInputSchema)
+    .mutation(async ({ input, ctx }) => {
+      const { targetProfileId, relationshipType } = input;
+
+      const { proposalId, processInstanceId } =
+        await assertProposalEngagementAccess({
```
scazan, line 34: This needs to live inside of the service layer as opposed to the router layer. This is to ensure that the service library maintains the permissions when used in other contexts.

services/api/src/routers/decision/proposals/removeRelationship.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
+  removeProposalRelationship: authenticatedConfirmedProcedure({
+    rateLimit: { windowSize: 10, maxRequests: 20 },
+  })
+    .input(proposalRelationshipInputSchema)
+    .mutation(async ({ input, ctx }) => {
+      const { targetProfileId, relationshipType } = input;
+
+      const { proposalId, processInstanceId } =
+        await assertProposalEngagementAccess({
```
scazan, line 23: This has the same issue where we're asserting permission inside of the router. What I would do instead is move it to the service layer into a new function on the service that wraps or calls the add profile relationship.

services/api/src/routers/decision/proposals/removeRelationship.ts
```diff
@@ -0,0 +1,38 @@
...
+import { proposalRelationshipInputSchema } from '../../../encoders/decision';
+import { authenticatedConfirmedProcedure, router } from '../../../trpcFactory';
+
+export const removeProposalRelationshipRouter = router({
+  /**
+   * Unlike/unfollow a proposal. Counterpart of `addProposalRelationship` —
+   * see that procedure for why this is confirmed-tier and proposal-only.
+   */
+  removeProposalRelationship: authenticatedConfirmedProcedure({
+    rateLimit: { windowSize: 10, maxRequests: 20 },
```
scazan, line 16: We don't need a custom rate limit here, I think.

services/api/src/routers/decision/proposals/addRelationship.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Like/follow a proposal. Separate from the generic (closed-network)
+   * `profile.addRelationship`: proposal engagement is open to confirmed
+   * out-of-network accounts — e.g. accounts claimed from a public decision
+   * process — and is gated on the parent decision instead: SUBMIT_PROPOSALS,
+   * the same permission commenting requires. Only proposal profiles are
+   * accepted.
+   */
+  addProposalRelationship: authenticatedConfirmedProcedure({
+    rateLimit: { windowSize: 10, maxRequests: 20 },
```
scazan, line 27: I don't think we need the rate limit here. We can just use the default one.

packages/common/src/services/decision/assertProposalEngagementAccess.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Proposal profiles carry no permissions of their own, so the grant is
+ * resolved on the process instance — the same pattern `assertPostReadAccess`
+ * uses for proposal posts (the comment write path reaches the parent decision
+ * via `resolvePostRoots`).
+ *
+ * Throws NotFoundError when the profile is not a proposal's — the proposal
+ * engagement endpoints accept only proposal targets. Returns the resolved
+ * proposal/process ids so callers don't need a second lookup.
+ */
+export async function assertProposalEngagementAccess({
```
scazan, line 26: I'm thinking we could probably just use `assertProfileTypeAccess` here.

scazan (approved):
I think, actually, there are still some changes to do but I will approve so we can merge the fix but if we do that then let's have a follow up that cleans up the code a bit.

services/api/src/routers/decision/proposals/addRelationship.ts
nourmalaeb: Done in follow-up #1598.

services/api/src/routers/decision/proposals/addRelationship.ts
nourmalaeb: Done in follow-up #1598.

services/api/src/routers/decision/proposals/removeRelationship.ts
nourmalaeb: Done in follow-up #1598.

services/api/src/routers/decision/proposals/removeRelationship.ts
nourmalaeb: Done in follow-up #1598.

services/api/src/routers/decision/proposals/addRelationship.ts
nourmalaeb: Done in follow-up #1598.

packages/common/src/services/decision/assertProposalEngagementAccess.ts
nourmalaeb: Done in follow-up #1598.

PR #1584, opened 2026-07-15
feat(ui): first-class Storybook stories for all @op/sense primitives

nourmalaeb:
- One story file per @op/sense primitive in `packages/ui/stories-sense/`, separate from the `stories/` (@op/ui) directory so it can move wholesale when Storybook leaves @op/ui
- Shared `withSense` decorator scopes stories in the `.sense` theme class; portaled content (dialogs, menus, toasts) re-scopes itself with `className="sense"`
- Stories surface under a `Sense/` section, sorted to the top of the sidebar
- Chart story deferred: recharts 3.8 pulls @reduxjs/toolkit 2.12, which needs a newer immer than the lockfile's 10.1.1 — add it back after an immer bump


nourmalaeb:
Split into the family PRs instead: infra + form-input stories → #1561, and one story commit each on #1572–#1577. Chart story still deferred pending the immer bump.

PR #1585, opened 2026-07-15
feat(terms): Re-accept updated policies modal

op-bot-common:
Adds a non-dismissable, app-wide modal that prompts already-onboarded users to review and re-accept the updated Terms of Use, Privacy Policy, and Code of Conduct. It shows for any logged-in (non-anonymous) user whose `onboardedAt` is on or before July 12, 2026 — the instant the updated policies go live. New users signing up afterward accept the current versions in the normal onboarding flow, so they never see it.

Clicking a policy link opens that document inside the same modal with a back button; "Agree and continue" reuses the existing `account.completeOnboarding` mutation, which stamps a fresh `onboardedAt` and records tos/privacy acceptance. That moves the user past the cutoff, so invalidating `getMyAccount` recomputes eligibility and closes the gate.

Asana: https://app.asana.com/0/1213315744811204/1215335788530623

Screenshot 2026-07-16 at 12 38 21


apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+      });
+    }
+  };
+
+  return (
+    <Modal
+      isOpen
+      isDismissable={false}
+      isKeyboardDismissDisabled
+      className="h-screen max-h-none w-screen max-w-none overflow-y-auto sm:h-auto sm:max-h-[80vh] sm:w-[34rem] sm:max-w-[34rem]"
```
scazan: We probably don't need any of this and can rely on the defaults for our modal

apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+  return (
+    <ModalBody className="gap-6 p-8 sm:px-10 sm:py-10">
+      <div className="flex flex-col gap-2 text-center">
+        <Header1>{t("We've updated our policies.")}</Header1>
+        <p className="text-base text-neutral-charcoal">
+          {t('Review the changes and accept to keep using Common.')}
+        </p>
+      </div>
+
+      <section className="flex flex-col gap-2 rounded-xl border border-neutral-gray1 bg-neutral-offWhite p-4">
```
scazan: Can we use a Surface component for this? We'll still need this background color though

apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+          {t("What's changed")}
+        </span>
+        <p className="text-base text-neutral-charcoal">
+          {t(
+            'Common now works with a third-party service that automatically reviews content posted on the platform to keep the community safe and uphold our Code of Conduct. Our Terms of Use and Privacy Policy now explain how this works and what it means for your data.',
+          )}
+        </p>
+      </section>
+
+      <div className="flex flex-col gap-4">
```
scazan: Since we want the gap-6 here instead, could we just eliminate this div entirely?

apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+    terms: t('Terms of Use'),
+    privacy: t('Privacy Policy'),
+    conduct: t('Code of Conduct'),
+  };
+  const Content = documentContent[document];
+
+  return (
+    <>
+      <div className="sticky top-0 z-30 flex min-h-16 items-center border-b bg-white px-4">
+        <button
```
scazan: Can we use a pre-existing component here? Perhaps the same one used on Back arrows in our header when you are on the proposal editor for instance

apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+// it sits inside (stop the press from reaching the surrounding React Aria
+// Checkbox).
+const PolicyLink = ({
+  onOpen,
+  children,
+}: {
+  onOpen: () => void;
+  children: ReactNode;
+}) => (
+  <button
```
scazan: Same thing here.. we should use a pre-existing Button component so we can get the consistent styles for this. The styles are correct here but we should be using a variant of our button with those styles

apps/app/src/components/PolicyReacceptanceModal/eligibility.ts
```diff
@@ -0,0 +1,29 @@
+import type { CommonUser } from '@op/api/encoders';
```
scazan, line 1: This file is only used in one place. just put these items into the file that uses them in that case.

apps/app/src/lib/i18n/dictionaries/en.json
```diff
@@ -35,6 +35,14 @@
   "Exit": "Exit",
   "Effective Date: March 15, 2025": "Effective Date: March 15, 2025",
   "Privacy Policy": "Privacy Policy",
+  "Terms of Use": "Terms of Use",
+  "Code of Conduct": "Code of Conduct",
+  "We've updated our policies.": "We've updated our policies.",
+  "Review the changes and accept to keep using Common.": "Review the changes and accept to keep using Common.",
+  "What's changed": "What's changed",
```
scazan, line 42: You are missing the other translations

scazan (changes requested):

apps/app/src/components/PolicyReacceptanceModal/index.tsx
op-bot-common: Done — dropped the className and rely on the Modal defaults.

apps/app/src/components/PolicyReacceptanceModal/index.tsx
op-bot-common: Done — now uses `<Surface variant="filled">`, which keeps the neutral-offWhite background.

apps/app/src/components/PolicyReacceptanceModal/index.tsx
op-bot-common: Done — removed that wrapper; the checkbox row and button are now direct children of ModalBody so they use its gap-6.

apps/app/src/components/PolicyReacceptanceModal/index.tsx
op-bot-common: Done — switched to our `Button` (link variant) with the LuArrowLeft icon, matching the proposal editor header's back control.

apps/app/src/components/PolicyReacceptanceModal/index.tsx
op-bot-common: Done — these are now `Button` with `variant="link" size="inline"`. I also moved them next to the checkbox instead of inside its label, so no manual press-propagation handling is needed.

apps/app/src/components/PolicyReacceptanceModal/eligibility.ts
op-bot-common: Moved out of utils/ and colocated next to the component as `eligibility.ts`. I kept it as a tiny pure module (rather than inlining into index.tsx) so the unit test can import it without pulling the client component — and next-intl/next-navigation — into the node test env. Happy to inline fully if you'd prefer dropping the unit test.

apps/app/src/lib/i18n/dictionaries/en.json
op-bot-common: Done — added the strings to every dictionary (ar, bn, fr, pt, so) in addition to en/es.

scazan (approved):

apps/app/src/components/PolicyReacceptanceModal/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+    terms: t('Terms of Use'),
+    privacy: t('Privacy Policy'),
+    conduct: t('Code of Conduct'),
+  };
+  const Content = documentContent[document];
+
+  return (
+    <>
+      <div className="sticky top-0 z-30 flex min-h-16 items-center border-b bg-white px-4">
+        <Button variant="link" size="inline" className="gap-1" onPress={onBack}>
```
nourmalaeb: This is wrapping, maybe you want a `ghost` button here instead? The `inline` button is best for text content inside other text content. 

Image

nourmalaeb (approved):
Would be good to update the back button on the `PolicyDocumentView` component before we ship this.

PR #1612, opened 2026-07-17
fix(profile): migrate profile image uploads to signed-URL storage flow

nourmalaeb:
Profile avatar/banner uploads sent the file as base64 inside the tRPC mutation body, which Vercel rejects over ~4.5MB — users saw a bare "failed to fetch" and a stuck edit form with no way to recover (reported in Columbus, 4 affected users in PostHog since July 11). Uploads now use the same signed-URL flow as the decision hero image: sign → PUT the binary straight to storage → record the path, so the file never rides in a request body.

- new `profile.signProfileImageUploadUrl` / `profile.saveProfileImage` endpoints, parameterized by `profileId` — personal and org profiles both hang images off a `profiles` row, so one endpoint pair serves both (own personal profile allowed; otherwise requires profile `UPDATE` access, i.e. org admins)
- validation + persistence live in `@op/common` profile services with the shared storage trust-boundary check (profile-scoped path prefix, stored MIME vs allowlist, 5MB image cap); personal-profile avatar mirrors onto the `users` row inside a transaction
- shared `useProfileImageUpload` hook (error toast, preview rollback, stale-request guard) replaces the duplicated FileReader/base64 logic in `BaseUpdateProfileForm` and `PersonalDetailsForm`
- old base64 `account.uploadImage` / `account.uploadBannerImage` endpoints removed
- gating + behavior tests (org-admin path, cross-profile authz + prefix rejection, MIME mismatch, non-image, persistence)

Stacked on top: #1614 migrates the organization forms (still on base64 `organization.uploadAvatarImage`) onto a draft-space variant of this flow and extracts the shared upload hook. Remaining follow-up: delete replaced storage objects (pre-existing leak, hero flow precedent).

Asana: https://app.asana.com/1/1108348036672214/project/1210594608104944/task/1216624731912034


packages/common/src/services/profile/saveProfileImage.ts
```diff
@@ -0,0 +1,99 @@
...
+  await db.transaction(async (tx) => {
+    await tx
+      .update(profiles)
+      .set(
+        imageType === 'avatar'
+          ? { avatarImageId: storageObjectId }
+          : { headerImageId: storageObjectId },
+      )
+      .where(eq(profiles.id, profileId));
+    if (isPersonalProfile && imageType === 'avatar') {
```
scazan, line 90: This is actually surfacing a real issue, which is that the Profile should probably always carry this image, regardless of it being an organization or a personal profile. This is fine for now because the issue already exists, but we should actually eliminate this sooner than later and have a single way of handling it. I believe right now we could potentially be uploading an image onto the user and then displaying a different image from the profile as there is some confusion between the two. An old sin we need to correct.

packages/common/src/services/profile/signProfileImageUploadUrl.ts
```diff
@@ -0,0 +1,81 @@
...
+ * `saveProfileImage` knows to mirror the avatar onto the users row.
+ */
+export async function assertProfileImageAccess({
+  user,
+  profileId,
+}: {
+  user: User;
+  profileId: string;
+}): Promise<{ isPersonalProfile: boolean }> {
+  const caller = await db.query.users.findFirst({
```
scazan: Hmm.. I think I see why you did this for the isPersonalProfile specifically but we might as well do this is parallel with the below since assertProfileAccess doesn't need the caller here.. We can save a step there by getting it whether we will need it or not.

scazan (approved):

packages/common/src/services/profile/saveProfileImage.ts
nourmalaeb: agreed — filed it in eng triage: https://app.asana.com/1/1108348036672214/project/1211387328265612/task/1216662549242970. Plan is to make `profiles.avatarImageId` the single source of truth, repoint anything still reading `users.avatarImageId`, then drop the column and this mirror write.

Also noted the deeper version of the same sin in that task: personal-profile ownership being implicit (the `users.profileId` pointer) instead of a real access-zones grant is why this code needs an `isPersonalProfile` branch at all. If we grant an explicit owner role when the profile is created (+ backfill), `assertProfileImageAccess` collapses into plain `assertProfileAccess` and both special cases die together.

packages/common/src/services/profile/signProfileImageUploadUrl.ts
nourmalaeb: done in 0bf48d89f. One wrinkle: couldn't put `assertProfileAccess` itself in the Promise.all — it throws when the caller has no roles, and personal-profile owners have no role row on their own profile (ownership is just the `users.profileId` pointer), so every personal upload would 403. Used `getProfileAccessRoles` (the non-throwing primitive it wraps) in parallel with the users lookup, then `checkPermission` after the own-profile check. Same authz, one round trip.

PR #1628, opened 2026-07-21
feat(sense): visual effects + icons — Confetti, LogoLoop, CheckIcon, MegaphoneIcon

nourmalaeb:
Final P2 family:

- **Confetti** — canvas burst copied as-is (internal dep of the Modal→Dialog swap in P3)
- **LogoLoop** — reactbits marquee copied as-is (1 use, ComingSoon)
- **icons/** — CheckIcon + MegaphoneIcon inline SVGs, exported via `@op/sense/icons`
- Stories: replayable confetti burst, marquee with placeholder logos, icon sheet. No comparison stories — these are pixel-identical copies with no styling to diff.

This completes the P2 composite ports: every entry in the inventory's rebuild/copy columns now exists in @op/sense.

Asana: https://app.asana.com/1/1108348036672214/project/1216361907129487/task/1216401654627076


nourmalaeb:
@greptileai review

PR #1629, opened 2026-07-21
feat(ai): bootstrap @op/ai package with a provider-agnostic OpenAI-compatible client

op-bot-common:
Bootstraps a new workspace package `@op/ai` exposing `getAIClient()`, a lazily-initialized OpenAI SDK client that works with any OpenAI-compatible provider. `AI_BASE_URL` selects the provider endpoint (unset falls back to the SDK default) and `AI_API_KEY` supplies the credential. The client throws an explicit error when the key is unset — otherwise the OpenAI SDK silently falls back to `OPENAI_API_KEY` and its error names the wrong variable. The `OpenAI` re-export is type-only so consumers can type params without gaining a bypass around the configured client or pulling the SDK into browser bundles.

Also registers `AI_API_KEY` / `AI_BASE_URL` in `turbo.json` `globalEnv` (Turbo strict env mode strips undeclared vars), adds them to both env example files, and adds the `w:ai` workspace shortcut.

No consumers yet — bootstrap only (no service integrations, model catalogs, or retry/streaming helpers).

Note: `pnpm-lock.yaml` picked up a few benign in-range re-resolutions from `pnpm install` (magic-string, tinyexec, sourcemap-codec) alongside the new `openai@6.48.0` entry.

Multica issue: ONE-608


op-bot-common:
Superseded by #1631, which now targets the base branch directly (no stack). The raw OpenAI client this PR introduced was made redundant by Mastra's built-in model router, and #1631 already contains this branch's commits plus the Mastra layer. Closing per discussion on ONE-608.

PR #1631, opened 2026-07-21
feat(ai): bootstrap @op/ai package with a Mastra agent layer

op-bot-common:
Bootstraps a new workspace package `@op/ai` — our single, provider-agnostic path to LLM inference, built around [Mastra](https://mastra.ai) used purely as a self-hosted library (no Mastra Cloud; usage telemetry disabled by default).

A very basic use-case is supported for a standard chat-style agent. No memory storage or anything yet.

- `createAIAgent({ name, instructions, id?, model })` returns a Mastra `Agent`. `model` is `{ modelId, baseURL?, apiKey? }`, resolved through Mastra's built-in model router, which natively speaks any OpenAI-compatible endpoint — so there's no `openai` SDK dependency.
- Model choice is per-agent, in code (no `AI_MODEL` env var), so the app can run several models side by side.
- The inference endpoint is runtime config: `baseURL`/`apiKey` are passed by the caller (e.g. from end-user settings), with `AI_BASE_URL`/`AI_API_KEY` as deploy-level fallbacks. `apiKey` is optional for keyless self-hosted endpoints (Ollama, vLLM). Model resolution is deferred to first use so module-scope agents don't read env at import time.
- `resolveAIModelConfig()` is exported for direct Mastra workflow usage.

Also registers `AI_API_KEY`/`AI_BASE_URL` in `turbo.json` `globalEnv` and both env example files, and adds the `w:ai` workspace shortcut. No consumers yet — bootstrap only.


packages/ai/src/model.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export interface ResolvedAIModel {
+  providerId: string;
+  modelId: string;
+  url: string;
+  apiKey: string | undefined;
+}
+
+export const resolveAIModel = ({
```
scazan: let's call this resolveAIModelConfig

packages/ai/src/agent.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    const contents = body.messages.map(
+      ({ content }: { content: string }) => content,
+    );
+    expect(contents).toContain('Answer concisely.');
+    expect(contents).toContain('What is the capital of France?');
+  });
+});
+
+describe('holding a multi-turn conversation', () => {
+  it('threads prior turns into later completions so the model remembers', async () => {
```
scazan, line 129: This test primarily serves as an example really

valentin0h (approved):
Let's go 🤖

PR #1634, opened 2026-07-21
fix(account): let existing accounts log in regardless of the invite allow-list

valentin0h:
Accounts created outside the allow-list gate (anon claim flow, pre-#1507 rejected-OAuth orphans) have no allowList entry and were locked out of login entirely once their session ended. The allow-list is a network-membership concept, not an authentication one — an existing account may always log back in (the OTP/OAuth exchange proves email ownership; isNetworkMember gates walled-garden access separately), so it now only decides whether an email with no account may create one. Existence is checked against auth.users because public.users.email is drifted (broken sync_user_email trigger, tracked separately).

### DIFF
--- a/services/api/src/routers/account/login.test.ts
+++ b/services/api/src/routers/account/login.test.ts
@@ -1,9 +1,21 @@
-import { expect } from 'vitest';
+import { randomUUID } from 'crypto';
+import { describe, expect, it } from 'vitest';
 
+import { appRouter } from '..';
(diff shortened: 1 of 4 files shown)


scazan (approved):
Looks good but it would be nice to clarify some of the logic for easier reading. I don't have a big suggestion on that though. Maybe that's more a sign that we should away from the DB trigger user creation soon.

PR #1638, opened 2026-07-23
fix(onboarding): unblock promote-flow users who did not finish onboarding

valentin0h:
Returning non-members who abandoned onboarding mid-session were permanently 403'd: every route redirected them to /start, which then rejected them once the transient ?promote=1 was gone. Onboarding now admits any real account and branches on account state, and onboarding redirects preserve where the user was headed.

### DIFF
--- a/apps/app/src/components/Onboarding/PromoteOnboardingFlow.tsx
+++ b/apps/app/src/components/Onboarding/PromoteOnboardingFlow.tsx
@@ -3,6 +3,7 @@
 import { analyzeError, useConnectionStatus } from '@/utils/connectionErrors';
 import { trpc } from '@op/api/client';
 import { isSafeRedirectPath } from '@op/common/client';
+import { logger } from '@op/logging/client';
 import { LoadingSpinner } from '@op/ui/LoadingSpinner';
 import { StepperProgressIndicator } from '@op/ui/Stepper';
 import { toast } from '@op/ui/Toast';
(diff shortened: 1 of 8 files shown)


apps/app/src/components/Onboarding/PromoteOnboardingFlow.tsx
```diff
@@ -77,6 +80,14 @@ export const PromoteOnboardingFlow = ({
...
       await completeOnboarding.mutateAsync({ tos: true, privacy: true });
       await trpcUtils.account.getMyAccount.invalidate();
+      if (!isSafeRedirectPath(redirectParam)) {
+        // `/` sends a non-member into the walled garden (403) — track how often
+        // onboarding dead-ends here.
+        logger.warn(
+          'Promote onboarding finished without a safe redirect; landing may be gated',
+          { redirectParam },
+        );
+      }
```
nourmalaeb, line 90: This works for now, but would be great to have a fallback at `/` for users without walled garden access. A design question really.

nourmalaeb: Makes me wonder if we should hard-code `/columbus` as the fallback redirect destination until we have some better solution.

nourmalaeb (approved):
This unblocks the onboarding but the user will still land on a 403 page once they complete onboarding if they don't have walled garden access and they don't have a valid redirect destination in the url search params. A fine solution for now.

apps/app/src/components/Onboarding/PromoteOnboardingFlow.tsx
valentin0h: Yeah, I was thinking that we need a landing for non-network users.

PR #1640, opened 2026-07-23
chore: drop obsolete @typescript/native-preview dependency

valentin0h:
TypeScript 7 ships the native compiler in the canonical typescript package (pinned at 7.0.2), so the preview channel is redundant and unused.


valentin0h:
Closing: the premise was wrong. `@typescript/native-preview` is not redundant — Next.js's build-time TypeScript step (verify-typescript-setup) requires it as the marker that permits the canonical TS7 native compiler (which ships no lib/typescript.js JS API). Removing it makes `next build` fail in CI with 'do not have the required package(s) installed'. Both packages are needed: `typescript@7.0.2` for `tsc`, native-preview for `next build`.

PR #1685, opened 2026-07-30
feat(sense): TanStack forms on @op/sense — fields, validation, modal hosting

nourmalaeb:
Migrate the shared TanStack `useAppForm` field factory (`components/form/utils.tsx`) and all its consumers from `@op/ui` to `@op/sense`, then make the forms behave correctly in their (now sense) modal hosts.

**Stacked on `sense/auth-onboarding` (#1674)** — it shares the onboarding form files, so this branch builds on top of it (base = `sense/auth-onboarding`, not `sense-migration`).

### Field factory → sense, context-wired
The bound field components now read value/change/blur/error straight from the TanStack field via `useFieldContext()`, so call sites pass only presentation props (label, placeholder, options, isRequired) — sense-native APIs, no RAC shims.
- `TextField` / `TextArea` (Field + Input/Textarea/InputGroup-with-icon)
- `Select` — `options` array; value→label `items` map for base-ui's `SelectValue`
- `Switch` (was `ToggleButton`), `Checkbox`
- `Button` / `SubmitButton` → sense `Button`
- Dropped the unused `MultiSelectComboBox` binding

All 10 consumers rewritten to the slim API (onboarding ×5, profile ×3, decisions ProcessBuilder, coming-soon). Inverted/side-effecting toggles use a raw `<Switch>`.

### Validation UX
- **Native validation off**: `aria-required` instead of `required`, `noValidate` on every `<form>` — the browser's "fill out this field" / email bubbles are gone; TanStack + zod own it.
- **Submit surfaces errors**: `getFieldErrorMessage` shows errors once a submit is attempted, so an untouched invalid form no longer "does nothing" on Submit.
- **Single source**: prefer the live `onChange` result over merging the flat error list (was double-printing "Invalid email address, Invalid email").
- **Blur validation**: forms add an `onChange` validator; field components gate display to after-blur (errors appear when you leave a field, update live as you fix).
- **Select doesn't error on open**: marks blurred on popup *close*, not the trigger's DOM blur (opening moved focus into the listbox and flipped `isBlurred`).
- **Friendly, translated messages** across every modal form (`createValidator(t)` / `createOrganizationFormValidator(t)`); Update-org had no validator before. Added the `Enter a valid email address` key to all 7 dictionaries.

### Modal hosting
The `@op/ui` (React Aria) modals `inert` everything outside themselves, killing the base-ui Select popup (portaled to `<body>`) — dropdowns did nothing. Migrated the modals that host these forms to sense `Dialog` (base-ui Dialog cooperates with base-ui Select), with the shadcn layout: pinned `DialogHeader`, scrollable form body, pinned `DialogFooter` (was `@op/ui` `ModalFooter`, which floated over the content):
- `UpdateOrganizationModal`, `UpdateUserProfileModal`, `UpdateProfileModal`, `CreateOrganizationModal`, PlatformAdmin `UpdateProfileModal`

### Out of scope (separate per-area cleanup)
Consumers' own remaining `@op/ui`: `AvatarUploader`/`BannerUploader`, the `TermsMultiSelect`/`GeoNamesMultiSelect` wrappers, and `@op/ui/Toast` (`toast.error({title, message})` — only the sense Toaster is mounted, so those toasts should move to `@op/sense/Sonner`).

typecheck + format pass.


nourmalaeb:
Thanks @greptile — all three addressed in 20cc64ca9:

1. Removed the leftover `console.log('submit')` in PersonalDetailsForm.
2. CreateOrganizationModal: inlined `setIsSuccessOpen(false)` in the `isNew` effect and dropped the unstable `onSuccess` dependency (also removed the now-unused `onSuccess`).
3. Swapped the `as any` on PersonalDetailsForm's validators for `// @ts-expect-error`, consistent with the other migrated forms.

PR #1721, opened 2026-08-04
fix(decisions): paginate invite modal members server-side per role

valentin0h:
Asana: https://app.asana.com/1/1108348036672214/project/1211387328265612/task/1217093803809976

## Problem

The invite modal fetched only the **first page** of `profile.listUsers` (limit 25) and filtered members by role client-side — so for large processes (251-member stress test) the per-role "People with access" lists and tab badge counts were silently wrong, and loading all members client-side doesn't scale.

## Fix

Server-side per-role pagination:

- **`listProfileUsers`** gains an optional `roleId` filter — a purely additive `IN`-subquery over `profileUser_to_access_roles`. **No cursor/keyset/orderBy code touched**, so this composes cleanly with #1657's cursor rewrite (whichever lands second rebases trivially).
- **`profile.listUsers`** router accepts `roleId`.
- **`getRoles` / `profile.listRoles`** gain opt-in `includeMemberCounts` — one grouped count query per call (junction ⋈ profileUsers, `excludeGlobalUsers` applied, matching `listProfileUsers` semantics) so tab badges no longer require loading all members.
- **`ProfileInviteModal`**: per-role `useInfiniteQuery` (React Query caches per input key → each role tab paginates independently, switching tabs preserves loaded pages), bounded scrollable list with `useInfiniteScroll` trigger, skeleton rows while the first page loads. Badge = server `memberCount` + staged + pending invites.
- Client-side duplicate-exclusion (search dropdown, paste, add-email) now only sees loaded pages — acceptable since `inviteUsersToProfile` dedupes authoritatively server-side.

## Tests

- `users/listUsers.test.ts`: roleId filtering, empty for unheld role, 28-member cursor pagination across 2 pages (no dupes/gaps).
- `listRoles.test.ts`: memberCount correctness in both query branches, absent without the flag, global-user sentinel excluded.

44/44 pass locally; `pnpm typecheck` + `pnpm format:check` clean.

## Out of scope

- `listProfileInvites` remains unpaginated (invite volume is small).
- `ProfileUsersAccess` (Manage Participants table) unchanged.

### DIFF
--- a/apps/app/src/components/decisions/ProfileInviteModal.tsx
+++ b/apps/app/src/components/decisions/ProfileInviteModal.tsx
@@ -18,7 +18,6 @@ import { Skeleton } from '@op/ui/Skeleton';
 import { toast } from '@op/ui/Toast';
 import Image from 'next/image';
 import {
-  Key,
   type ReactNode,
   Suspense,
   useEffect,
@@ -113,8 +112,7 @@ function ProfileInviteModalContent({
   const utils = trpc.useUtils();
   const [selectedItemsByRole, setSelectedItemsByRole] =
     useState<SelectedItemsByRole>({});
-  const [selectedRoleId, setSelectedRoleId] = useState<string>('');
-  const [selectedRoleName, setSelectedRoleName] = useState<string>('');
+  const [requestedRoleId, setRequestedRoleId] = useState<string>();
   const [searchQuery, setSearchQuery] = useState('');
   const [debouncedQuery] = useDebounce(searchQuery, 200);
   const [isSubmitting, startSendTransition] = useTransition();
@@ -126,21 +124,17 @@ function ProfileInviteModalContent({
     width: 0,
   });
 
-  // Fetch roles with decisions zone permissions to identify admin roles
-  const { data: rolesWithPerms } = trpc.profile.listRoles.useQuery(
-    { profileId, zoneName: 'decisions' },
-    { enabled: isDraft },
-  );
-  const adminRoleIds = useMemo(() => {
-    if (!rolesWithPerms) {
-      return new Set<string>();
-    }
-    return new Set(
-      rolesWithPerms.items.filter((r) => r.permissions?.admin).map((r) => r.id),
-    );
-  }, [rolesWithPerms]);
-
-  const showDraftBanner = isDraft && !adminRoleIds.has(selectedRoleId);
+  const [{ items: roles }] = trpc.profile.listRoles.useSuspenseQuery({
+    profileId,
+    zoneName: 'decisions',
+    limit: 100,
+  });
+  const selectedRole =
+    roles.find((role) => role.id === requestedRoleId) ?? roles[0];
+  const selectedRoleId = selectedRole?.id ?? '';
+  const selectedRoleName = selectedRole?.name ?? '';
+  const showDraftBanner =
+    isDraft && !!selectedRole && !selectedRole.permissions?.admin;
 
   // Fetch existing pending invites
   const [serverInvites] = trpc.profile.listProfileInvites.useSuspenseQuery({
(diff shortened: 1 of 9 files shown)


apps/app/src/components/decisions/ProfileInviteModal.tsx
```diff
@@ -370,15 +400,51 @@ function ProfileInviteModalContent({
...
+  }: {
+    id: string;
+    roles: Array<{ id: string }>;
+  }) => {
     startOptimisticTransition(async () => {
-      dispatchRemoveUser(profileUserId);
+      dispatchRemoveUser(id);
+      const removedRoleIds = new Set(memberRoles.map((role) => role.id));
+      const updateMemberCounts = (delta: number) =>
+        utils.profile.listRoles.setData(rolesQueryInput, (rolesData) =>
```
valentin0h, line 414: Ideally, we use real-time channels for this. Potentially separate PR.

packages/common/src/services/access/getRoles.ts
```diff
@@ -66,17 +71,21 @@ type RoleCursor = { value: string; id: string };
  * - If no profileId: returns only exposable global roles (profileId IS NULL
  *   and named in EXPOSABLE_GLOBAL_ROLE_NAMES)
  * - If zoneName is provided: includes permission for that zone
+ * - If includeMemberCounts is set (and profileId is present): each role gains
+ *   a memberCount — the number of profile members holding that role
  */
 export const getRoles = async (params?: {
   profileId?: string;
   zoneName?: string;
+  includeMemberCounts?: boolean;
```
valentin0h, line 81: Would it make sense to just remove this flag and include it in all cases?

packages/common/src/services/access/getRoles.ts
```diff
@@ -114,6 +123,19 @@ export const getRoles = async (params?: {
...
+  // The role junction's accessRoleId index bounds the correlated lookup.
+  const memberCountExpr = (accessRoleCols: typeof accessRoles) => sql<number>`(
+    SELECT COUNT(*)::int
+    FROM ${profileUserToAccessRoles}
+    INNER JOIN ${profileUsers}
+      ON ${profileUsers.id} = ${profileUserToAccessRoles.profileUserId}
+    WHERE ${profileUserToAccessRoles.accessRoleId} = ${accessRoleCols.id}
+      AND ${profileUsers.profileId} = ${profileId}
+      AND ${excludeGlobalUsers(profileUsers.authUserId)}
+  )`;
```
valentin0h: if possible let's use drizzle queries

valentin0h (commented):
did a quick pass, lgtm

iwillig (approved):

packages/common/src/services/access/getRoles.ts
iwillig: We can remove this flag, but the following components would then pay the counting costs (a correlated COUNT subquery per role row) for a value they never read:

- `ShareProposalModal.tsx`
- `ProfileUsersAccess.tsx`
- `ProcessBuilder/stepContent/participants/RolesSection.tsx`
- `organization.getRoles` (used by `AddUserToOrgModal` / `InviteToOrganizationButton`) — these don't even pass a `profileId`, so a member count would be meaningless here anyway

Only `ProfileInviteModal` reads `memberCount` today. This flag was actually added in an earlier review pass on this branch for that reason (see efc4d0fd7) — happy to revisit if we want to change that tradeoff, but wanted to surface the cost.

apps/app/src/components/decisions/ProfileInviteModal.tsx
iwillig: Agreed — filed as a follow-up rather than blocking this PR: https://app.asana.com/1/1108348036672214/task/1217915210633272 (subtask of the parent invite-modal-pagination task).

PR #1725, opened 2026-08-05
refactor(sense): migrate proposal editor, view, and review/decide to @op/sense

nourmalaeb:
Completes the decisions area's move off `@op/ui`: the proposal editor and its collaborative fields, the proposal read view, the review flow, and the vote / selection / results screens. `@op/ui` is now gone from all of `components/decisions/`. Merges into `sense-migration` (#1662), not the default branch.

- **63 files swapped**, plus the Figma redesign items that were safe to apply alongside: editor fields rebuilt on sense `Field`/`Input` with labeled always-visible controls, share modal rebuilt on `Dialog` + `Combobox` (dropping a hand-rolled portal and its z-index hacks), Like/Follow moved from the header into the proposal's engagement row, status states as `StatusBadge`s, and the proposal header restructured to match its frame.
- **Review surfaces rebuilt too**: the review rubric form and the submitted-review view are composed from one sense `Field`/`FieldTitle`/`FieldDescription` set, so a criterion looks the same whether you're answering it or reading it back. Multiple-choice criteria are boxed radios, as in the editor.
- **Editor chrome**: the shared top formatting toolbar is replaced by a per-field bubble menu on the selection; the version-history aside is non-modal so a version can be previewed and scrolled beside the document, and previewing renders through the same read-only components a past version uses rather than a disabled collaborative editor. The bubble menu carries an image picker, so inserting an image is an explicit control again (paste and drop still work). The proposal title field starts as one line, grows as it wraps, and shows a live character count against a 50-character cap from the template frame; the server's ceiling stays at 200, since tightening it would make every existing longer title unsaveable.
- **Read view**: TipTap's `details` node renders through sense `Collapsible` as a small client island, so a collapsible animates and starts closed for readers instead of relying on native `<details>`. The editor keeps its own node view, so the two surfaces differ for now — one shared node view is filed separately.
- **`@op/sense` additions and spec fixes**: `FieldLabel variant="box"` for selectable option boxes, a `ghost` `Toggle` variant, `showOverlay` on `SheetContent`, a `render` prop on `FieldTitle`, and audits that brought `Badge` and `StatusBadge` back in line with their component sets. `--secondary` is now defined in the sense theme — it previously resolved through the legacy `@op/ui` theme and would have lost its background when that package is removed. (Correcting `--color-gray-100` to Figma's `#f1f2f2` alongside it made every border too light, so that part is reverted pending a design answer on the token.)
- **Accessibility**: criterion prompts are `h4` again — `@op/ui`'s `FieldHeader` gave them heading semantics for free and the sense rewrite had quietly dropped heading navigation from a long rubric. `FieldDescription` uses logical alignment, so descriptions no longer sit flush-left on RTL pages while their labels sit flush-right. Tooltips are provided once from the root layout: base-ui's `Provider` is the grouping primitive, and a provider per icon button made every icon a group of one. Toasts render above dialog overlays, split panes and profile links have real sense focus rings, and the engagement row's stats are only buttons when they actually do something.
- **Fixes found along the way**: the collaborative editor was importing `@op/ui/RichTextEditor` while the app's config already pulled sense defaults, putting two TipTap packages on one page; its placeholder never reached TipTap; `ReviewTabs`' force-mount maps to base-ui `keepMounted` (a naive swap would have dropped in-flight review answers); review analytics needed a guard against base-ui's automatic selection changes; `restoreVersion` was adding a duplicate entry at the top of the history; admins couldn't advance a phase before its start date; and a prose field kept its bubble menu after losing focus, so two menus sat on screen and a click aimed at one field applied formatting to the other.

The e2e suite is updated for the redesign and green. Most of it was renamed copy and controls that changed shape (single-selects are boxed radios, scored scales are comboboxes, confirm dialogs carry `role="alertdialog"`), and it caught the two real bugs above; five of the failures were inherited from the base branch and are fixed here as well.

Deferred deliberately, not forgotten: the remaining structural redesign items on shared contracts (`SelectionConfirmShell`, `PostUpdate`), two endpoint changes needed for the Results and review-selection filter/sort controls, and several items that need a design answer rather than code. RTL bidi handling for `Input`/`Textarea` didn't survive the migration from `@op/ui` and is tracked separately — it touches every form in the app. The process builder's rubric and proposal-template previews are removed rather than migrated: each forked the review form and had already drifted from it, and a shared preview is worth more than two more copies.

Worth a look during review: `StatusBadge` is 8px taller across its six call sites, the bubble-menu changes also affect the process builder's overview editor since that component is shared, and the root tooltip provider means hover delay is 100ms everywhere except the two action rows that opt into 500ms.


nourmalaeb:
@greptile-apps review

PR #1731, opened 2026-08-06
refactor(decisions): extract the review answer resolver into @op/common

valentin0h:
Answer-to-label resolution (opaque single_select ids, yes/no literals, the overall recommendation) lived inside SubmittedReviewView in apps/app, where the upcoming review-receipt Inngest sender cannot reach it. Move the pure rubric helpers into @op/common and add resolveSubmittedReview with an injected t so both the app view and the email sender resolve answers from one place. Pure refactor, no user-visible change.


valentin0h:
Superseded: this refactor only existed to render the rubric inside the email in #1732, which was replaced by the link-only email in #1716.

PR #1732, opened 2026-08-06
feat(reviews): email the reviewer a receipt on submit and on every edit

valentin0h:
Reviewers lose every link to their own reviews once the process advances past the review phase; a receipt email on each submit and edit hands them a durable copy and a URL that still resolves. Debounced per assignment so a quick edit after submitting yields one receipt, worded via timestamps only so a collapsed burst can never lie, and threaded per assignment through a synthetic References root. Stacked on #1731.


valentin0h:
Superseded by #1716: the review email is link-only now, so the rubric rendering in this PR is no longer needed.

PR #1750, opened 2026-08-10
feat(export) Export infra and UI in place

iwillig:
**Note**

1. Need to confirm playwright tests before merging. 
2. Review unit tests. 

Wires the proposal CSV export end to end: an admin-gated export control on the proposals list, behind the `export-feature` PostHog flag for staged rollout. The backend pipeline already existed but nothing ever called it; connecting it surfaced most of what the rest of these commits fix.

Two things to know before reading the diff. **Exports write to the shared public `assets` bucket**, not a dedicated private one — a private bucket needs manual provisioning in every hosted environment and nothing in this repo does that. The unguessable file name is therefore the access control, the signed URLs are link expiry rather than a boundary, and these CSVs carry submitter names and email addresses. That trade is deliberate and documented in `exports/constants.ts`. Separately, **every submitted proposal was exporting a blank description**: the reader asked for a `default` fragment, but templated documents key their fragments by field name (`summary`), so it worked for legacy proposals and silently failed for every current one.

Scope: `8d05e96d7` also carries a Turbopack workspace-root pin and a stale-cache fix in the Create menu, and `006d46a5b` carries two setState-in-render fixes in the ProcessBuilder. All four came out of debugging this feature. Flagging rather than unbundling.

## How the export runs

Four services share this flow, and no single file states it. These diagrams describe the code on this branch.

### 1. The export run, end to end

The mutation returns an id and finishes long before the file exists. The workflow does the work and the status cache holds the result.

```mermaid
sequenceDiagram
    autonumber
    actor Admin
    participant Button as ExportProposalsButton
    participant API as tRPC API
    participant Cache as Export status cache
    participant Inngest
    participant Workflow as exportProposals workflow
    participant Storage as Supabase Storage
    participant Realtime as Supabase Realtime

    Admin->>Button: Press Export
    Button->>API: decision.export(processInstanceId, filters)
    API->>API: assertProfileAccess(decisions ADMIN)
    API->>Cache: seed the record, status pending
    API->>Inngest: send proposalExportRequested
    API-->>Button: exportId
    Button->>API: decision.getExportStatus(exportId)
    API->>API: registerQueryChannels(proposalExport:{id})
    API-->>Button: status pending, channel proposalExport:{id}

    Inngest->>Workflow: run
    Workflow->>Cache: status processing
    Workflow->>Workflow: listProposals(the same filters, limit 1000)
    Workflow->>Workflow: generateProposalsCsv
    Workflow->>Storage: upload the CSV
    Workflow->>Storage: createSignedUrl
    Storage-->>Workflow: signed URL
    Workflow->>Cache: status completed, fileName, signedUrl
    Workflow->>Realtime: publish on proposalExport:{id}
    Realtime-->>Button: broadcast
    Button->>API: decision.getExportStatus(exportId)
    API-->>Button: status completed, signedUrl
    Button-->>Admin: Download CSV link
```

### 2. How the client learns the run finished

The button does not poll. The status query carries its channel back to the client, and the client re-reads on two cues: a confirmed join, and a broadcast. Supabase does not replay a broadcast, so the join cue is what covers a run that finishes first.

```mermaid
sequenceDiagram
    autonumber
    participant Button as ExportProposalsButton
    participant TrpcLink as tRPC channel link
    participant Registry as queryChannelRegistry
    participant Subscriber as QueryInvalidationSubscriber
    participant Manager as RealtimeManager
    participant Supabase as Supabase Realtime
    participant Workflow as exportProposals workflow

    Button->>TrpcLink: decision.getExportStatus(exportId)
    TrpcLink-->>Button: data plus the channel from _meta
    TrpcLink->>Registry: registerQuery(queryKey, channels)
    Registry->>Subscriber: query:added
    Subscriber->>Manager: subscribe(channel, handler, onSubscribed)
    Manager->>Supabase: join proposalExport:{id}

    alt The run finishes before the join lands
        Workflow->>Supabase: publish
        Note over Supabase: No subscriber yet.<br/>Supabase does not replay, so this message is lost.
        Supabase-->>Manager: SUBSCRIBED
        Manager->>Registry: notifyChannelSubscribed(channel)
        Registry->>Subscriber: channel:subscribed
        Subscriber->>Button: invalidate the channel's queries
        Button->>TrpcLink: re-read, status completed
    else The run finishes after the join lands
        Supabase-->>Manager: SUBSCRIBED
        Manager->>Registry: notifyChannelSubscribed(channel)
        Registry->>Subscriber: channel:subscribed
        Subscriber->>Button: invalidate, status still pending or processing
        Workflow->>Supabase: publish
        Supabase-->>Manager: broadcast, mutationId export:{id}
        Manager->>Subscriber: handler
        Subscriber->>Button: invalidate, status completed
    else The socket drops and rejoins
        Supabase-->>Manager: CLOSED
        Note over Manager: Everything published while the socket was down is gone.
        Supabase-->>Manager: SUBSCRIBED on rejoin
        Manager->>Registry: notifyChannelSubscribed(channel)
        Registry->>Subscriber: channel:subscribed
        Subscriber->>Button: invalidate, status completed
    end
```

### 3. Failure and the silence timeout

The workflow reports a failure the same way it reports success. A workflow that dies reports nothing, so the timer is the only thing that ends the wait.

```mermaid
sequenceDiagram
    autonumber
    actor Admin
    participant Button as ExportProposalsButton
    participant API as tRPC API
    participant Cache as Export status cache
    participant Workflow as exportProposals workflow
    participant Realtime as Supabase Realtime

    Note over Button: The timer restarts on each reported state.<br/>It bounds silence, not total run time.

    alt A step throws
        Workflow->>Cache: status failed, errorMessage
        Workflow->>Realtime: publish on proposalExport:{id}
        Realtime-->>Button: broadcast
        Button->>API: decision.getExportStatus(exportId)
        API->>Cache: read the record
        API-->>Button: status failed
        Button-->>Admin: Report the error, return to idle
    else The run dies without a terminal write
        Note over Workflow,Realtime: No status write and no broadcast.
        Button->>Button: 2 minutes pass without a reported state
        Button-->>Admin: Report the timeout, return to idle
    end
```

### 4. The channel machinery this uses

Every channel-backed query follows this path. The export adds a run-scoped channel and the confirmed-join cue; the rest already existed.

```mermaid
sequenceDiagram
    autonumber
    participant Query as Any query component
    participant TrpcLink as tRPC channel link
    participant Registry as queryChannelRegistry
    participant Subscriber as QueryInvalidationSubscriber
    participant Manager as RealtimeManager
    participant Supabase as Supabase Realtime
    participant Writer as A mutation or a workflow

    Query->>TrpcLink: call a query
    TrpcLink-->>Query: data plus the channels from _meta
    TrpcLink->>Registry: registerQuery(queryKey, channels)
    Registry->>Subscriber: query:added
    Subscriber->>Manager: subscribe(channel)
    Manager->>Supabase: join the channel
    Supabase-->>Manager: SUBSCRIBED
    Manager->>Registry: notifyChannelSubscribed(channel)
    Registry->>Subscriber: channel:subscribed
    Subscriber->>Query: invalidate the channel's queries

    Writer->>Supabase: publish on the channel
    Supabase-->>Manager: broadcast with a mutation id
    Manager->>Subscriber: handler
    Subscriber->>Subscriber: drop a mutation id it already saw
    Subscriber->>Query: invalidate the channel's queries

    Note over Query,Subscriber: A local mutation takes the same path without the socket.<br/>The link registers the mutation's channels and the registry emits mutation:added.

    Query->>Registry: TanStack removes the query
    Registry->>Subscriber: channel:removed once the last query leaves
    Subscriber->>Manager: unsubscribe
    Manager->>Supabase: leave the channel
```

## Follow-ups (out of scope)

- **1000-proposal silent truncation:** the workflow passes `limit: 1000` with no warning past it. Blocks the Columbus GIS export, which plausibly exceeds it — `listProposals` already supports keyset pagination.
- **Export status is cache-only:** no backing table, so there is no export history and a completed export reads as `not_found` once the record expires.
- **Attachment signed URLs gate nothing:** `getProposalAttachmentsWithSignedUrls.ts` and `getProposal.ts` sign against the same public bucket. Same exposure, wants its own ticket.
- **Stale TODO:** `ProposalsList.tsx:338` proposes extracting an export button; this PR added one by a different route.
- **`Generating...` almost never shows:** the workflow writes `processing` to the cache and broadcasts only at a terminal state. The label therefore changes only when the join re-read happens to land after that write, so most runs read `Preparing...` for the whole wait. Broadcasting on the `processing` write would make the two states visible; the timeout logic already tells them apart.

## LLM Usage

- opus

## Risk Justification

- CSV access rests on an unguessable UUID in the object key. Deliberate, recorded on the Asana task.
- Behind a feature flag, so rollout is staged and reversible without a deploy.
- Not yet verified end to end against real TipTap credentials — the description fix is covered by unit tests but wants one export from the review build before merge.

Asana: https://app.asana.com/0/1215095416548894/1215120685966991

## Feature Flag

Feature flag disabled on Thru Aug 13, 2026

image


apps/app/src/components/decisions/ExportProposalsButton.tsx
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }, [isStale]);
+
+  // Surface a failed workflow once, then return the button to its idle state so
+  // the admin can retry.
+  useEffect(() => {
+    if (!isFailed) {
+      return;
+    }
+    const message =
```
iwillig, line 170: Should this display anything to the user?

iwillig: never mind, we are using the toast object here.

apps/api/next.config.mjs
```diff
@@ -28,6 +32,12 @@ dotenv.config({
 const config = {
   serverExternalPackages: ['sharp', 'onnxruntime-node'],
   turbopack: {
+    // Pin the workspace root. Turbopack otherwise infers it by walking up for
```
iwillig: not sure if this is required.

packages/common/src/services/decision/exports/generateProposalsCsv.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  documentContent: undefined,
+};
+
+const makeProposal = (
+  overrides: Partial<ProposalFromList> = {},
+): ProposalFromList => ({ ...baseProposal, ...overrides });
+
+/**
+ * Minimal reader for the fully-quoted CSV `csv-stringify` emits under
+ * `{ quoted: true }`. Hand-rolled because the repo has no CSV *parser*
```
iwillig: We should include a CSV parser.

apps/app/src/components/decisions/ProcessBuilder/stepContent/rubric/ReviewSettingsContent.tsx
```diff
@@ -85,21 +85,24 @@ export function ReviewSettingsContent({
...
-        saveChanges({ phases: phasesWithScope(updated.scope) });
-      }
-      if (updates.reviewsAllowRevisions !== undefined) {
-        saveChanges({
-          config: { reviewsAllowRevisions: updated.reviewsAllowRevisions },
-        });
-      }
-      return updated;
-    });
+    const updated = { ...settings, ...updates };
```
scazan: This feels a bit unrelated to this PR, so I'm wondering if it should happen in a separate PR instead rather than this one. There are a few others like this in the PR that we should watch out for.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -518,6 +525,27 @@ const ProposalsListContent = ({
...
+          exportControl={
+            canExportProposals ? (
+              <ExportProposalsButton
+                processInstanceId={queryParams.processInstanceId}
+                isEmpty={total === 0}
+                // Forwarded straight from the list's own query params (minus
+                // pagination) so the export can't scope differently from the
+                // list it was launched off.
+                filters={{
+                  categoryId: queryParams.categoryId,
```
scazan: Seems like we could just spread in the filters? I'm thinking that if we add filters in the future we'll need to update it here as well which makes me think that we should actually pull this out into a specific type or something that will expand automatically so we don't have to maintain both.

iwillig:
Feedback.


1. Use the real time features for the download button.

apps/app/src/components/SiteHeader/CreateMenu.tsx
```diff
@@ -37,8 +37,16 @@ export const CreateMenu = () => {
...
-      const { processes: templates } =
-        await utils.decision.listProcesses.ensureData({});
+      // `fetch`, not `ensureData`: template ids are regenerated on every seed
+      // (decision_processes.id is an autoId, and seed-access-control dedupes on
+      // name), so a cached list can hand out an id the server no longer has and
+      // createInstanceFromTemplate 404s. `ensureData` returns cached data
+      // without revalidating, and with gcTime 24h + refetchOnWindowFocus off a
+      // stale entry survives all day. `fetch` respects staleTime (0 here), so
+      // the id is always checked against the server.
+      const { processes: templates } = await utils.decision.listProcesses.fetch(
```
scazan: Interesting.. I wonder if it matters much since they won't change but we might get new ones. Or if the problem is further up where we query by name instead of id (something else we can fix). Just flagging it but not a blocker.

apps/app/next.config.mjs
```diff
@@ -66,6 +66,12 @@ const config = {
     authInterrupts: true,
   },
   turbopack: {
+    // Pin the workspace root. Turbopack otherwise infers it by walking up for
+    // lockfiles, so a stray package-lock.json above the monorepo (e.g. an `npm
+    // install` run in $HOME) silently moves the root and emits externals
+    // symlinks under .next/dev/node_modules with the wrong relative depth —
+    // producing "Cannot find module '@swc/helpers-<hash>/...'" at runtime.
+    root: path.resolve(__dirname, '../..'),
```
scazan: yeah, not sure if we need this (or why it specifically was needed in this PR). Is this a legit issue ? curious because we haven't run into it before.

packages/common/src/services/decision/schemaValidator.ts
```diff
@@ -269,6 +294,29 @@ export class SchemaValidator {
       }
       case 'format':
         return `${friendlyName} has an invalid ${error.params?.format} format`;
+      case 'oneOf':
```
scazan: I'm a little suspect of the need for these changes. Are these updates necessary for this PR or should they exist as a follow-up.

apps/app/src/components/SiteHeader/CreateMenu.tsx
scazan: I'm also thinking that this might not belong in this PR. Wondering if it is anything to do with the exports.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
```diff
@@ -119,17 +119,22 @@ function PhaseDetailForm({
...
+  // Safe to read `phase` from the closure because every caller updates once
+  // per event — if that ever changes, this needs a ref, not an updater.
   const updatePhase = (updates: Partial<PhaseDefinition>) => {
-    setPhase((prev) => {
-      if (!prev) {
-        return prev;
-      }
-      const updated = { ...prev, ...updates };
-      const phasesPayload = toPayload(
+    if (!phase) {
```
scazan, line 154: Is this change also maybe better in a separate PR? Or does it affect the exports specifically?

apps/app/src/components/decisions/ExportProposalsButton.tsx
```diff
@@ -0,0 +1,182 @@
...
+    },
+  });
+
+  const { data: status, refetch: refetchStatus } =
+    trpc.decision.getExportStatus.useQuery(
+      { exportId: exportId ?? '' },
+      { enabled: Boolean(exportId) && !hasTimedOut },
+    );
+
+  // The export finishing is announced on its own channel, so there is no poll.
```
scazan: yes! nice to have this on the realtime channel rather than polling :)

apps/app/src/components/decisions/ExportProposalsButton.tsx
```diff
@@ -0,0 +1,182 @@
...
+  // confirmed covers exactly that gap: anything that settled earlier is visible
+  // in this read, and anything later arrives as a broadcast.
+  useEffect(() => {
+    if (!exportId) {
+      return;
+    }
+
+    const channel = Channels.proposalExport(exportId);
+
+    return queryChannelRegistry.on('channel:subscribed', (event) => {
```
scazan: I'm feeling there is a bit of a smell here in that this sounds like an issue that could be true of any channel. Should it be fixed higher up potentially? Practically I wouldn't imagine that this would be so quick since the export workflow is likely to kick off a bit after the moment the query is subscribed, but maybe it's a bigger issue we should solve in our realtime service? If so we can submit that separately maybe.

apps/app/src/components/decisions/ExportProposalsButton.tsx
```diff
@@ -0,0 +1,182 @@
...
+    logger.error('Proposals export failed', { error: message });
+    toast.error({ message: message || t('Export failed') });
+    setExportId(null);
+  }, [isFailed, status, t]);
+
+  useEffect(() => {
+    if (!isRunning) {
+      return;
+    }
+    const timer = setTimeout(() => {
```
scazan, line 198: A little curious about this one. I imagine the case here is that our channel has been disconnected and we missed a broadcast. Since status is dictated by the server end we'd want that timeout to mostly be communicated from there but I guess in a low-connectivity situation this could be helpful (although adds some complexity, especially if we want to support long-running exports of unknown length). Maybe we should increase the timeout at least to make it the extreme case.

Just a very quick glance on this part... but looking for where we clear this timer once it's running and we get an update of failure or success. Is this potentially triggering no matter what?

packages/common/src/services/decision/exports/generateProposalsCsv.test.ts
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, doing background research on the open questions in this PR. I'm deliberately *not* answering; I'm gathering context so the humans can decide.

**Status: already actioned on this branch.** `csv-parse` was added as a devDependency of `@op/common` and the hand-rolled reader was replaced (`import { parse } from 'csv-parse/sync'`; `parseCsv = (csv) => parse(csv)`), removing ~48 lines.

**Why the original hand-rolled reader was likely written:** the repo depended on `csv-stringify` but had no CSV *parser*, so adding one meant adding a dependency — and the assertion only needed to read a single row. The JSDoc on the original said as much. The likely reasoning was "one small helper beats a new dependency for one test."

**What that reasoning missed:** `csv-parse` is the sibling package of the `csv-stringify` the exporter already writes with, so it isn't really a new dependency family. And a hand-written reader puts untested code in the assertion path — if it's wrong, the suite can pass over a malformed file or fail over a correct one, and either way blames the exporter.

One detail worth a human eye: the replacement is left **strict** (no `relax_column_count`), so a row whose width disagrees with the header now fails the test rather than silently shifting every column.

apps/api/next.config.mjs
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching the open questions on this PR. Not answering, just gathering evidence for the humans.

**Where it came from:** commit `30366632c` ("fix: stop local dev failing in ways that look like app bugs"). The identical pin was added to `apps/app/next.config.mjs` in the same commit.

**Why the LLM likely added it:** Turbopack infers the workspace root by walking up for lockfiles. If any lockfile exists *above* the monorepo (e.g. an `npm install` run in `$HOME`), the inferred root moves and Turbopack emits externals symlinks under `.next/dev/node_modules` at the wrong relative depth — surfacing at runtime as `Cannot find module '@swc/helpers-<hash>/...'`. That reads as an application bug, which is the class of problem that commit was addressing. Pinning `turbopack.root` is Next's documented remedy.

**Evidence on whether it's needed here:**
- `pnpm-lock.yaml` **is** at the repo root, so in a clean checkout inference already resolves correctly — the pin is redundant in that case.
- It's insurance against a stray lockfile *above* the monorepo, which is machine-specific and can't be determined from the repo.
- In `apps/api/next.config.mjs`, `path` and `__dirname` are used **only** by this pin, so removing it makes those imports dead too.
- The `.env.local` loading just above it reaches up three levels for git worktrees, which is a layout where root inference is genuinely more fragile.

The one check that would settle it is machine-local: whether a `package-lock.json` / `pnpm-lock.yaml` exists above the repo on the machine where the failure was seen.

apps/app/next.config.mjs
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR's open questions. Not answering; leaving the call to you.

Same change as the one on `apps/api/next.config.mjs` — both were added together in `30366632c` ("fix: stop local dev failing in ways that look like app bugs").

**On "why was it needed in this PR":** the commit is one of several on this branch that came out of debugging local dev while building the export, rather than from the export code itself. The PR body flags that bundling explicitly. So the honest framing is that it's a local-dev fix that travelled with the feature work, not something the export requires.

**On "is this a legit issue / why haven't we hit it before":** the failure mode is real but environment-dependent, not repo-dependent — it only triggers when a lockfile exists *above* the monorepo (classically from running `npm install` in `$HOME`). That's why it can appear on one machine and never on anyone else's, and why it wouldn't have shown up before. The symptom is `Cannot find module '@swc/helpers-<hash>/...'` at runtime, which looks like a broken build rather than a misresolved workspace root.

**Counter-evidence:** `pnpm-lock.yaml` is at the repo root, so inference resolves correctly in a clean checkout — meaning the pin is defensive rather than load-bearing for most people. Whether it's load-bearing for anyone here can only be checked on the affected machine.

apps/app/src/components/decisions/ProcessBuilder/stepContent/rubric/ReviewSettingsContent.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching the open questions here. Not answering; the scope call is yours.

**Provenance:** this change came in via `cacf35dd6` ("fix(decision): get proposal exports writing to storage again"), together with the `PhaseDetailPage.tsx` change flagged separately. Both are setState-in-render fixes, and the PR body lists them as bundled rather than required:

> `006d46a5b` carries two setState-in-render fixes in the ProcessBuilder. All four came out of debugging this feature. Flagging rather than unbundling.

(The hashes in the body predate a rebase; the content is the same.)

**Why the LLM likely bundled them:** they were almost certainly hit while exercising the ProcessBuilder to create a decision to export from — i.e. encountered *en route* to testing the feature, not sought out. An LLM working a single task tends to fix what blocks it and carry the fix along, because unbundling costs a branch, a PR, and a review cycle for a change it already has in hand. The commit message and PR body disclose the bundling rather than hiding it, which suggests the model recognised the scope problem and chose disclosure over separation.

**What that trades away:** it makes this PR's diff wider than its title, and it couples an unrelated fix's review to the export feature's. Your comment notes there are several of this shape, which is the accumulation cost of that choice.

Worth noting these are genuinely small and independently revertible, so splitting them out later is cheap if you'd rather they land separately.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR. Not answering; flagging evidence only.

**Does it affect exports?** No — this is a setState-in-render fix in the ProcessBuilder, with no code path to the export feature. It came in via `cacf35dd6` ("fix(decision): get proposal exports writing to storage again"), the same commit as the `ReviewSettingsContent.tsx` change you flagged.

**Why the LLM likely included it:** the same pattern as the sibling comment — these look like problems met while driving the ProcessBuilder to set up a decision to export from. React logs setState-in-render loudly in dev, so it would have been visible in the console while testing, and an LLM mid-task tends to fix what it trips over rather than defer it, since deferring means a new branch and another review cycle for something already diagnosed.

The PR body does disclose this ("two setState-in-render fixes in the ProcessBuilder … Flagging rather than unbundling"), so the bundling was a stated decision rather than an oversight.

**For your decision:** it's self-contained and independently revertible, so extraction is cheap. The argument for leaving it is that it's already reviewed here; the argument against is the one you're making — the diff no longer matches the PR title, and it sets a precedent for the next feature branch.

apps/app/src/components/decisions/ProposalsList.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching the open questions. Not answering; the design call is yours.

**Partly actioned since you commented.** The prop is no longer hand-enumerated — it's now the shared type:

```ts
filters: ProposalExportFilters;
```

`proposalExportFiltersSchema` (in `services/events/src/types.ts`) is the single definition the event payload, the request input, and the status response all derive from, so adding a filter updates the component's contract automatically. That's the "specific type that will expand automatically" you asked for.

**What is still hand-maintained:** the *call site* in `ProposalsList.tsx` still lists each key when building the object from `queryParams`.

**Why the LLM likely kept that explicit rather than spreading:** `queryParams` is a superset — it also carries `limit`, and pagination must not travel to the export (the job pages independently, and a `limit` in the payload would silently cap the CSV). A bare `{...queryParams}` would forward it. The enumeration is the crude way to exclude it.

**The gap that leaves:** a filter added to the list and forgotten here still silently widens the export — which is the exact regression `5eb45b163` fixed, where three filters were dropped. There's a compile-time guard for filters that *aren't* valid query inputs, but nothing catches a valid filter that simply isn't forwarded.

A destructure-and-rest (`const { limit, ...exportFilters } = queryParams`) would make new filters flow automatically while still excluding pagination — noting it as the shape your comment points at, not as a recommendation.

apps/app/src/components/SiteHeader/CreateMenu.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR's open questions. Not answering; recording findings.

**Your "further up" instinct matches what the code now says.** The comment added here names exactly that:

> template ids are regenerated on every seed (`decision_processes.id` is an autoId, and `seed-access-control` dedupes on **name**), so a cached list can hand out an id the server no longer has

So the root cause is upstream: the seed dedupes on `name` while the id is regenerated, meaning a logically identical template gets a new id on every seed. The client-side change treats the symptom.

**Why the LLM likely fixed it at the call site:** the observed failure was `Template with ID '<uuid>' not found` when creating a decision, and the proximate cause was `ensureData` returning a cached list without revalidating (gcTime 24h, `refetchOnWindowFocus` off, so a stale entry survives all day). Switching to `fetch` — which respects `staleTime` — makes the id always checked against the server. That's a one-line change at the point of failure. Changing the seed's identity strategy is a data-layer change with migration implications, well outside an export PR.

**On "I wonder if it matters much since they won't change but we might get new ones":** in a deployed environment they don't change, so the exposure is mostly local dev and CI, where seeds re-run. The cost of the current fix is one uncached request per decision creation.

The durable fix your comment gestures at — seeding with stable ids, or resolving templates by a stable key rather than a regenerated one — would make the client-side revalidation unnecessary.

iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR. Not answering; evidence only.

**Does it have anything to do with exports?** No. It came in via `30366632c` ("fix: stop local dev failing in ways that look like app bugs"), which is the local-dev commit on this branch rather than export code. Creating a decision is a *precondition* for testing an export — you need an instance with proposals — so this is a blocker encountered on the way to the feature, not part of it.

**Why the LLM likely bundled it:** the failure presents as `Template with ID '<uuid>' not found` when clicking Create, which stops any manual verification of the export dead. An LLM driving a task to completion tends to clear whatever blocks verification and carry the fix along, since deferring means the feature can't be exercised at all. The PR body does disclose it as bundled.

**Argument for extracting it:** it's independent of the export, it touches a shared header component every page renders, and — per the sibling thread — the real cause is upstream in how templates are seeded. Reviewing it beside export code makes it easy to wave through.

**Argument for leaving it:** it's a one-line change (`ensureData` → `fetch`) with an explanatory comment, and it fixes a failure anyone re-seeding locally will hit.

packages/common/src/services/decision/schemaValidator.ts
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR's open questions. Not answering; the scope call is yours.

**Provenance:** commit `6659365f7` ("fix(decision): say which field a rejected selection belongs to"). Unlike the ProcessBuilder changes flagged elsewhere, this one has a concrete reproduction recorded.

**What it fixes:** submitting a proposal could fail with the message **`0 is invalid`**, naming neither the field nor the problem. The `0` is an array index leaking out: AJV reports a failing array element at `/category/0`, the validator took the last path segment as the field name, `properties["0"]` doesn't exist, so the display name fell through to the index. Any array-valued field reported this way.

The specific trigger found was an instance configured with **two categories sharing the same label**. The category schema emits one `oneOf` branch per option, so a value matching two branches satisfies neither "exactly one" nor "wrong" — the submitter picked a real option, and the ambiguity is in the configuration behind it. Every proposal choosing that option fails on every attempt until an admin removes the duplicate, so reporting it as an invalid *selection* sends the one person who cannot fix it back to re-pick a value that can never validate.

**Why the LLM likely included it here:** it was hit while trying to create a proposal to export — same "blocked on a precondition" pattern as the other bundled changes. It's diagnosis-only: duplicate labels are still accepted at configuration time, and the underlying validation gap is untouched.

**For your decision:** it's self-contained (one validator + its tests) and has no dependency on the export code, so extraction is clean. Against extraction: it's covered by tests added in the same commit, and the error it replaces is genuinely unactionable for the user who sees it.

apps/app/src/components/decisions/ExportProposalsButton.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR's open questions. Not answering; this one is a genuine design fork and yours to settle.

**Your instinct is right that it generalises.** The gap is not export-specific: a client only subscribes after a response carrying the channel, so anything broadcast before the socket join is lost, and Supabase does not replay. The same is true on **reconnect** — if the socket drops and returns, every invalidation published in the interval is lost permanently, for every feature, with nothing to recover it.

**Why the LLM likely solved it locally anyway.** The signal itself was added to shared infrastructure — `RealtimeManager.subscribe()` gained an `onSubscribed` callback, and `queryChannelRegistry` gained a `channel:subscribed` event — but **nothing consumes it by default**; only the export button opts in. That was a deliberate narrowing, recorded in the registry JSDoc:

> Nothing invalidates on this by default — broadcasts published before the join are lost, but for most queries the data was fetched at about the same moment the channel was registered, so the window holds nothing worth re-reading. It is opt-in for the callers where it does: those waiting on a background job that can settle before the join lands.

The reasoning was that making it app-wide would add a refetch per channel subscription across every realtime query — a behaviour change to features the PR doesn't touch, and a load increase — which felt like more than a feature branch should decide unilaterally.

**On "practically I wouldn't imagine this would be so quick":** agreed for hosted environments, where Inngest has real scheduling latency. It's most plausible locally, where `inngest-cli dev` picks the event up immediately and a single-proposal instance finishes in well under a second. The consequence differs by design, though — under polling a lost broadcast self-healed on the next tick; with polling removed it would hang until the timeout and report failure on a successful export.

So the mechanism is already in shared code and the only question is who subscribes to it. Promoting it to app-wide (and thereby fixing the reconnect gap) is a one-place change in `QueryInvalidationSubscriber` — plausibly the separate PR you suggest.

apps/app/src/components/decisions/ExportProposalsButton.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, researching this PR. Not answering; but your last question is factual, so here is what the code does.

**"Is this potentially triggering no matter what?"** — No. The timer is cleared by the effect's own cleanup, and the effect is keyed on `isRunning`:

```ts
useEffect(() => {
  if (!isRunning) return;
  const timer = setTimeout(/* … */, EXPORT_WAIT_TIMEOUT_MS);
  return () => clearTimeout(timer);
}, [isRunning, t]);
```

`isRunning` is derived from the status, and goes false on `completed`, `failed`, timeout, or a stale filter set. Any of those re-runs the effect, the cleanup fires, and the pending timer is cleared before the branch that would fire the toast. So a successful or failed export cancels it; it only reaches the toast if neither terminal state ever arrives.

**Why the LLM likely added it at all:** it predates the realtime work, when this was a poll loop and the timeout bounded the polling. Going realtime-only removed the poll, which changed the timeout's role from "stop polling" to "the only thing that can end the wait" — a workflow that dies without writing a terminal status now leaves nothing else to notice.

**On your point that status should be dictated server-side:** the record is written by the workflow, so the server *is* authoritative for `completed`/`failed`. What it cannot report is its own disappearance — a run that dies before writing a terminal status, or a broadcast lost to a dropped socket. The client timeout only covers that silence.

**On the value and the long-running case:** it's currently 2 minutes, chosen when generation was expected in seconds and capped at 1000 proposals. Once that cap is lifted (tracked as a follow-up), the "unknown length" problem you raise becomes real — a large export could legitimately exceed any fixed client-side bound, and the timeout would then abort a healthy run. Raising it is the cheap mitigation; a server-reported progress or heartbeat signal is the one that removes the guess.

apps/app/src/components/decisions/ProposalsList.tsx
iwillig: 🤖 **LLM research note — correction to my previous reply.**

I suggested `const { limit, ...exportFilters } = queryParams` as the shape that would let new filters flow automatically. That destructure is incomplete: `queryParams` carries **two** keys that are not export filters, not one.

```ts
type ProposalQueryParams = {
  processInstanceId: string;   // ← passed separately as its own prop
  categoryId?, submittedByProfileId?, votedByProfileId?,
  status?, dir, phase?, excludeAssignedForReview?,
  limit: number;               // ← pagination, must not reach the export
};
```

Everything except `processInstanceId` and `limit` is exactly `ProposalExportFilters`, so the correct form excludes both:

```ts
const { processInstanceId, limit, ...exportFilters } = queryParams;
```

The underlying point stands and is arguably stronger for it: the exclusion list is precisely two well-understood keys, which is what makes a rest-spread viable rather than fragile. Still not a recommendation — the human call is whether the automatic-expansion benefit is worth taking `queryParams` as the source of truth for a second consumer.

apps/app/src/components/decisions/ProcessBuilder/stepContent/rubric/ReviewSettingsContent.tsx
iwillig: Going to break this change into its own PR.

apps/app/src/components/decisions/ProposalsList.tsx
iwillig: Addressed in commit 0b5bd75abe24421a7016ad040feb7f68a81f7d6e

apps/app/src/components/SiteHeader/CreateMenu.tsx
iwillig: Moved into its own PR. https://github.com/oneprojectorg/common/pull/1783

packages/common/src/services/decision/schemaValidator.ts
iwillig: Fixed and moved to its own PR.

apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
iwillig: Moved to its own PR.

apps/app/src/components/decisions/ExportProposalsButton.tsx
iwillig: Addressed in commit 62b89416983a95777d7b276f83386cd404af8920

iwillig:
> Feedback.
> 
>     1. Use the real time features for the download button.

This is done.

apps/app/src/components/decisions/ExportProposalsButton.tsx
iwillig: 🤖 **LLM research note** — Claude again, reporting what changed on the branch since the note above.

Two commits landed after that reply, and both touch this line.

**`7b082e73d` re-aimed the timer.** It now bounds silence rather than total run time: the effect restarts whenever the run reports a new state. The dependency list I quoted earlier is stale — it reads `[isRunning, reportedState, t]` now. The clearing behaviour is unchanged, so the answer to "is this triggering no matter what" is still no.

**Your long-running-export point stands, and the silence bound does not yet fix it.** The workflow broadcasts only when the run reaches `completed` or `failed`. It writes `processing` to the cache record and tells nobody. After the first read the client therefore hears nothing until the run ends, so a 2-minute silence bound behaves as a 2-minute bound on total duration — the case you raised. With `limit: 1000` and a document fetch per proposal, a large export can pass that and get reported as timed out while it is still working.

Same root cause as the label: `Generating...` almost never appears, because `processing` never reaches the client.

Tracked in Asana as "Export button never reports Generating..." (`1217471411621878`). Three options, cheapest first:

1. Broadcast on the `processing` write. One line in the workflow. Buys one restart, not liveness.
2. Heartbeat per step, or per N proposals. The silence bound then tracks real liveness, and no ceiling on total run time is needed.
3. Raise `EXPORT_WAIT_TIMEOUT_MS`. Your suggestion, and still the cheapest thing that helps today.

One cost worth naming: the timeout ends only the client's wait. The workflow keeps running and may write the file a second later. Export status is cache-only with no history, so that file is then unreachable and the admin re-runs the whole export. Filed separately as `1217471543979321`.

Leaving the thread unresolved — that call is yours.

apps/app/src/components/decisions/ProposalsList.tsx
```diff
@@ -518,6 +541,15 @@ const ProposalsListContent = ({
           hasLocationField={hasLocationField}
           effectiveView={effectiveView}
           onViewChange={handleViewChange}
+          exportControl={
+            canExportProposals ? (
+              <ExportProposalsButton
```
scazan, line 530: Let's make sure we have a good error boundary around this. We can also use APIErrorBoundary.

packages/common/src/services/decision/exportProposals.ts
```diff
@@ -26,52 +35,76 @@ export const exportProposals = async ({
...
+    phase: input.phase,
+    excludeAssignedForReview: input.excludeAssignedForReview,
+  } satisfies AppliedExportFilters;
+
+  // Seed the full status record — not just the id and state — so the frontend
+  // can poll immediately and read the same shape it gets once the workflow
+  // takes over. A partial seed would answer the first poll with a record
+  // missing `format`/`filters`, which no longer satisfies the status contract.
+
+  // Deferred: this record exists because the client polls. Once export state
```
scazan: Are we still polling? Just curious if the comment is stale (in which case just remove the comment or make it way more concise)

packages/common/src/services/decision/exportProposals.ts
```diff
@@ -82,18 +115,11 @@ export const exportProposals = async ({
...
       userId: user.id,
       format: input.format,
-      filters: {
-        categoryId: input.categoryId,
-        submittedByProfileId: input.submittedByProfileId,
-        status: input.status,
-        dir: input.dir,
-        proposalFilter: input.proposalFilter,
-      },
+      filters,
```
scazan: Yeah, let's remove the filters so that we don't need to pass them around. I think we should just have a standard template for this export rather than having it react to how it's currently filtered. I think users will also be confused about that.

scripts/ensure-e2e-redis.mjs
```diff
@@ -0,0 +1,94 @@
...
+const deadline = Date.now() + 30_000;
+while (Date.now() < deadline) {
+  if (await isListening()) {
+    console.log(`[e2e] Redis ready on ${HOST}:${PORT}`);
+    process.exit(0);
+  }
+  await new Promise((resolve) => setTimeout(resolve, 250));
+}
+
+console.error(`[e2e] Redis did not accept connections on ${HOST}:${PORT}`);
```
scazan, line 93: Was this a real problem you run into or is this something that Claude was just suggesting as a whole to fill? If the latter I say let's not have it in this PR because we haven't run into this at all before. But it is true that we're not specifically testing our Redis and probably don't need to much because it's just acting currently as a simple cash key value basically.

package.json
```diff
@@ -21,7 +21,7 @@
     "docker:down": "docker compose -f docker-compose.dev.yml down",
     "docker:seed": "docker compose -f docker-compose.dev.yml exec api sh -c 'cd /app && /app/node_modules/.bin/tsx services/db/seed-access-control.ts'",
     "docker:migrate": "docker compose -f docker-compose.dev.yml exec api sh -c 'cd /app && pnpm w:db migrate'",
-    "dev:e2e": "cross-env E2E=true NEXT_PUBLIC_SUPABASE_URL=http://127.0.0.1:56321 DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:56322/postgres S3_ASSET_ROOT=http://127.0.0.1:56321/storage/v1/object/public/assets TIPTAP_SECRET=e2e NEXT_PUBLIC_TIPTAP_APP_ID=e2e turbo dev:e2e",
+    "dev:e2e": "node ./scripts/ensure-e2e-redis.mjs && cross-env E2E=true REDIS_URL=redis://127.0.0.1:6380 NEXT_PUBLIC_SUPABASE_URL=http://127.0.0.1:56321 DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:56322/postgres S3_ASSET_ROOT=http://127.0.0.1:56321/storage/v1/object/public/assets TIPTAP_SECRET=e2e NEXT_PUBLIC_TIPTAP_APP_ID=e2e turbo dev:e2e",
```
scazan, line 24: Also, I'm not totally against spinning up redis in order to do the testing, but just wondering if the payoff makes sense.

scazan (approved):
I proved, especially since it's behind a feature flag, but I do think we should definitely get rid of the filters. I think they'll cause more confusion, but it'll also remove a lot of complexity from this.

Just a flag, lets also just make sure that we are never exporting much PII so we should limit that as much as possible. Let's discuss what that shape looks like together.

apps/app/src/components/decisions/ProposalsList.tsx
iwillig: 🤖 **LLM research note** — I'm Claude, an LLM, doing background research on the open questions in this PR. Deliberately *not* answering; gathering context so the humans can decide.

**What a boundary would catch here today: almost nothing.** The button uses non-suspense `useQuery` (line 75) and `useMutation` (line 60). In react-query v5 `throwOnError` defaults to `false`, and there is no global `defaultOptions` overriding it in `apps/app` — the only `throwOnError` usages in the app are two explicit `false`s elsewhere. So neither hook throws during render, and an `APIErrorBoundary` wrapped around line 546 would sit there catching render-time throws that don't currently exist.

**The unhandled path is real, but it fails as a wrong message rather than as a throw.** The mutation already has one (`onError` logs and toasts, lines 66–69). The status *query* has none, and its failure is not inert:

- `status` stays `undefined` → `isResolved` and `isFailed` stay false → `isRunning` stays true (line 85)
- the silence timer (line 127) then fires `Export timed out. Please try again.`

So a failing `getExportStatus` is reported to the admin as a timed-out export — while the workflow may have written the file perfectly well. Reaching a boundary from that path needs `throwOnError: true` on the query, which is a behaviour change (boundary fallback instead of the timeout toast), not just a wrapper.

**Two details about `APIErrorBoundary` that shape the choice.** It keys fallbacks on `error.data?.httpStatus` and **re-throws** when neither a matching code nor a `default` is supplied — so a bare wrap escalates to the page boundary rather than containing anything. The nearby precedents split: `ReviewProgressStats` and `ReviewAssignmentsList` use `{ default: () => null }` (drop the widget silently); `ProposalsList:584` renders an inline message.

**One thing that looks wrong but isn't:** `exportControl` is threaded into `ProposalsStickyFilterBar` as a `ReactNode` prop, so it's easy to read a boundary at 546 as being in the wrong place. Boundaries are positional in the rendered tree, not lexical in the file, so wrapping there does contain the button's subtree and does leave the filter bar mounted.

packages/common/src/services/decision/exportProposals.ts
iwillig: 🤖 **LLM research note** — Claude again, researching this PR's open questions. Not answering; recording what the code says.

**Stale, yes — and it's two comments, not one.** Polling was removed on this branch in `10fc3e749` ("announce a finished export instead of polling for it"). The client query (`ExportProposalsButton.tsx:75`) has no `refetchInterval`; it carries only `enabled`. Both comment blocks at this site still describe the old arrangement:

- **lines 86–89** — "so the frontend can poll immediately"
- **lines 91–95** — "this record exists because the client polls"

**They're stale in different degrees, which matters for whether you delete or rewrite.**

*86–89 is a wording slip over a claim that survives.* The seed still has to be complete, because the client's **first read** — now driven by the confirmed-join re-read rather than a poll tick — has to satisfy the same contract: `format` and `filters` are required on `ExportStatusData`, and `getExportStatus`'s output schema requires `filters`. Only the word "poll" is wrong.

*91–95 is stale in substance.* It says the record exists *because* the client polls, and proposes moving export state onto a realtime channel as future work — but that move already landed on this branch. What actually keeps the record now is that it's the only store of export state: the broadcast carries no payload (`mutationId` only, workflow line 54), subscribers re-read `getExportStatus`, and there is no table behind it. So the deferred note argues for something already done while omitting the reason the record is load-bearing.

**A third one nobody has flagged.** `services/workflows/src/functions/exports/exportProposals.ts:42–45`, the `notifyExportFinished` JSDoc, still reads:

> a dropped broadcast costs latency, not correctness, and the client's poll still resolves the run on its own. That is the reason polling stays in place rather than being replaced

That's now false in the way that matters. With polling gone, a lost broadcast costs correctness — a healthy export reported as timed out — which is precisely the gap the `channel:subscribed` opt-in was added to cover.

**Context worth keeping if these get rewritten rather than cut:** the CI failure this branch just spent three runs on (`325924399`) came from the same "cache is the only record" property — status written three times, read back absent, no error anywhere. That property is the durable reason for the seed; the polling framing was only ever the proximate one.

packages/common/src/services/decision/exportProposals.ts
iwillig: 🤖 **LLM research note** — Claude, researching the open questions on this PR. Not answering; this is a product call and it's yours. Here's the blast radius and the parts that are decisions rather than deletions.

**What comes out.** `proposalExportFiltersSchema` (`services/events/src/types.ts:21`) is the single definition, so removing it drops the field from the event payload (`types.ts:68`), the mutation input (`export.ts:9`), the status response (`getExportStatus.ts:22`) and `ExportStatusData`. Then:

- `exportProposals.ts` — the `filters` object (74–84) and the `AppliedExportFilters` guard (24–25). That guard exists only to stop a filter travelling that `listProposals` can't apply; moot once nothing travels.
- workflow `exportProposals.ts:92–107` — `listProposals` keeps `processInstanceId`, `limit`, `skipAccessCheck`, `includeDocumentContent` and loses the other seven.
- `ExportProposalsButton` — the `filters` prop, and with it the entire stale-file mechanism: `filterKey` (57), `startedFilterKey` (58), `isStale` (84) and the reset effect (89–95). That is the bulk of the complexity you're pointing at, and it exists *only* because the file describes a filtered view.
- `ProposalsList` — the rest-spread at 511–515 and the prop at 549. Worth noting that spread landed three commits ago in `f4e847067`, in response to your earlier comment on this same line, so this supersedes that thread rather than building on it.

**Four things that are decisions, not deletions:**

1. **`dir` is not optional** in the schema (the router defaults it to `'desc'`). Row *order* is inherited today, not just the row set. A standard-template export still needs a stated sort order — it just moves from URL state to a constant.
2. **`isEmpty`** (`ProposalsList.tsx:548`) is `total === 0`, the *filtered* count. If the export stops following filters, the button still disables on a filtered-empty view while the instance has rows to export. That prop wants to become `totalProposalCount`.
3. **Couples to the 1000-row follow-up.** Every export becomes instance-wide, so the silent truncation at `limit: 1000` stops depending on what the admin narrowed to and becomes reachable by any instance over 1000 proposals.
4. **Points opposite to your PII note on the approval.** The columns are fixed (`Submitted By`, `Submitter Email`, `Profile ID`), so filters are currently the only thing bounding how many names and addresses leave in one file. Removing them makes every export the maximal-PII export. Not an argument against — the column set is the better lever, and you've said you want to discuss that shape — but the two changes pull in opposite directions, which is an argument for landing the column decision alongside rather than after.

**Test impact is structural, not cosmetic.** `tests/e2e/tests/proposals-export.spec.ts` is built around the filter: two proposals seeded, list narrowed by category, one row asserted. Its own comment (197–200) says that narrowing is what makes the row count an assertion rather than a coincidence. Remove filters and the spec loses its discriminator — it needs a new one (two proposals, two rows), not an edit. `exportProposals.test.ts:118` also asserts the forwarded filters `.strict()`-parse and goes with them.

scripts/ensure-e2e-redis.mjs
iwillig: 🤖 **LLM research note** — Claude, researching this PR. Not answering; but this question is factual, so here's the trail.

**Real problem, and the branch records it run by run.**

1. `ce5e61048` added the export e2e spec. It passed on every developer machine and failed on every CI run.
2. `edd3b0bec` fixed a *different* CI-only cause first — the workflow step starting the job runner couldn't take effect on the branch that added it, because the e2e workflow triggers on `pull_request_target` and GitHub runs the workflow file from the base branch. Still failing.
3. `a3b4737cf` added socket/status tracing to the spec, explicitly temporary ("comes out once the cause is known"). It showed the channel join, the reply and the broadcast all arriving on time, with the status still reading absent — which ruled out the realtime path everyone was suspecting.
4. `325924399` added this script.
5. `325a89379` removed the tracing.

**The mechanism.** `@op/cache`'s `get`/`set` are Redis-only: `set` returns silently when `REDIS_URL` is unset, `get` answers null, and neither errors. Developers have `REDIS_URL` in `.env.local` and a container already up; CI had neither. Export state is cache-only — no table behind it — so the status was written three times and read back as absent, the browser waited out its timeout, and the admin was told the export failed while the file sat in storage.

So it's the opposite of speculative filler: it's the specific fix for a specific red run, found by instrumenting for it, and it's the first thing on this branch that made the export spec pass in CI.

**Your reading of the intent is right, though.** This isn't testing Redis — nothing asserts on it. It's *providing* one, the same category as Supabase: a precondition of the feature under test.

**Two placement details that aren't arbitrary.** It runs before the servers because `@op/cache` builds its client at module load and stops reconnecting after three attempts, so a Redis that appears later is never picked up. And it runs from `start:e2e` rather than from the workflow for the `pull_request_target` reason above — a service container in the workflow is the more idiomatic form and would reduce this to a no-op (it exits early when the port already answers), but it can't fix the branch that adds it.

**What it doesn't address**, per its own commit message: `set` swallows a dropped connection and a command timeout as readily as a missing URL, so the same silent failure is available in production — and the export is the worst caller to lose it on, because the cache is its only record.

package.json
iwillig: 🤖 **LLM research note** — Claude, researching this PR's open questions. Not answering; the cost/benefit call is yours. Here are both sides measured rather than estimated.

**Cost.** One `docker run redis:7-alpine` mapped to 6380, plus a TCP readiness poll capped at 30s. It exits at line 51–54 when something already answers on the port, so a stack that grows a real service container makes it a no-op. Docker is already a hard requirement here for Supabase, so it adds no new dependency. Port 6380 rather than 6379 keeps a local run from reading or evicting the keys of the stack on 6379, matching how the databases are already kept apart. The one genuine tax: it's on `start:e2e` and `dev:e2e`, so every e2e run pays it — including the a11y scan, which has no use for it and can't be scoped out of it.

**Payoff.** It isn't buying Redis coverage; it's buying the export spec's ability to assert anything at all. Export status is cache-only — there is no table — so without Redis every status write is discarded, every read answers absent, and the spec can never reach its terminal assertion (`Download CSV` visible → fetch the signed URL → parse the bytes). The spec covers the two things this PR is actually about:

- the filter reaches the file (two proposals seeded, one row asserted)
- `Description` is non-empty for a templated document — the regression `78edaaa50` fixed, where **every submitted proposal exported a blank description** because the reader asked for a `default` fragment while templated documents key by field name

That second one is the case unit tests were passing over before this branch, and it's the failure mode the PR body calls out as having silently affected every current proposal.

**So the question underneath yours may be a different one.** Whether the container is worth it *given* the spec is one thing; whether the export deserves an e2e spec at all is another. The script has no other caller — if the spec goes, it goes with it.

**One change would remove the cost without removing the coverage:** giving export status a table, already on your follow-up list. The e2e stack would then need no Redis for this feature, and the production silent-failure mode the commit message flags (`set` swallows a dropped connection as readily as a missing URL) would close at the same time.

packages/common/src/services/decision/exportProposals.ts
iwillig: We are no longer polling. Updating the comments.

package.json
iwillig: 🤖 **LLM research note** — Claude again. Unlike my earlier notes on this PR, this one has an answer in it as well as research.

**Short version: yes, it has to stay — but you were right to ask, and the reason the payoff is narrow is worth more than the container is.**

**Why it can't just come out.** `package.json:24` and `:34` are the only source of `REDIS_URL` in the e2e stack — nothing in `.github/workflows/e2e-tests.yml` or `tests/e2e` supplies one. Without it the status writes vanish silently, the `Download CSV` link never appears, and `proposals-export.spec.ts` fails on its 90-second wait. That is precisely the CI failure `325924399` was written to fix, so removing the preflight takes the export spec with it.

**But the payoff really is as narrow as you suspected, and not for the reason I'd have guessed.** About fifty files import `@op/cache`, which looks like broad justification. It isn't:

- ~40 sites call `cache({ fetch })` — a read-through wrapper with an in-process LRU in front of Redis and a `fetch` fallback behind it. Without Redis these are *slower*, not wrong.
- `invalidate` / `invalidateMultiple` no-op harmlessly when nothing is cached.
- The proposals export is the **only** caller using raw `get`/`set`, and those bypass the memory layer entirely — straight to Redis or nothing.

So every other caller treats the cache as an optimisation. The export treats it as a system of record. The container is buying exactly one call site: the one where the cache isn't a cache.

**Which turns your question into a better one.** It isn't really "is one e2e spec worth a container" — it's that the export keeps its only record in a cache. Giving export status a table (already on the follow-up list) would delete this script and its 94 lines, drop the cost from *both* e2e jobs — the a11y-baseline job also runs `start:e2e` at `.github/workflows/e2e-tests.yml:451` and has no use for Redis — and close the matching production failure, where `set` swallows a dropped connection or a command timeout exactly as readily as a missing URL, on the one caller that cannot survive it.

**Two things that limit the cost meanwhile.** The script exits at line 51 when anything is already listening, so it is built to become a no-op rather than to need removing: once this merges to `dev`, a `services:` block can be added to the base branch, which is the more idiomatic form and was only impossible here because the workflow triggers on `pull_request_target` and runs the workflow file from the base. And Docker is already a hard requirement of this stack for Supabase, so nothing new is being asked of a developer machine.

iwillig: 🤖 *Written by Claude.*

To state the decision plainly: the Redis preflight stays, because without it there is no way to test the download at all.

The export's status lives only in the cache. With no Redis the workflow's status writes are discarded and every read comes back empty, so the button never reaches `Download CSV`. That link is what the rest of the test hangs off — the signed URL, the fetched bytes, the parsed CSV, the row count, and the Description column are all reached through it. Lose the status and none of them are observable.

So the choice is Redis or no end-to-end coverage of the download. Unit tests still cover the request that starts an export, but nothing else would exercise the workflow, the upload to storage, or the file that comes out the other end.

PR #1207, opened 2026-05-21
fix(decisions): include vote counts in My Ballot results page

op-bot-common:
The My Ballot tab queried `listProposals` with `votedByProfileId` but nothing ever requested the vote-count subquery, so every card fell through to `voteCount ?? 0` and reported zero while the funded-proposals tab showed real counts. Ballot reads now always carry a `voteCount`, gated on a successful `decision_process_results` row so a live tally can't leak mid-vote: `null` until results are published, a real count afterwards — where `0` unambiguously means zero recorded votes. `MyBallot` hides the Total Votes footer while the count is `null` rather than rendering a misleading zero, matching how the funded tab already gates its footer on result stats.

Rebased onto `dev`; the earlier client-side approach (passing `includeVoteCounts` from `MyBallot` and exposing it on `proposalFilterSchema`) is gone, so `services/api/src/encoders/decision.ts` is untouched.

## Follow-ups (out of scope)

- **Vote counts on My Ballot aren't subject to the `cowop` carve-out:** `ResultsList` hides its Total Votes footer for `slug === 'cowop'`; My Ballot doesn't, so that profile now sees counts on one tab and not the other. Worth deciding whether the carve-out should be shared (or dropped).

Asana: https://app.asana.com/0/1213315744811204/1215015720055366

PR #1765, opened 2026-08-12
feat(decisions): admin-only phase reversal

op-bot-common:
Phase advancement is one-way, so an operator who advances a decision prematurely has no way back. This adds a platform-admin action that undoes the most recent advancement — the button lives only on the backend admin screen (`/admin/decisions/[instanceId]`), behind `withAuthenticatedPlatformAdmin`. There is deliberately no user-facing equivalent; advancement stays one-way in the product.

The reversal undoes what the advance wrote, in one transaction: the review assignments generated on entry to the phase, the `decision_transition_history` row (which cascades to `decision_transition_proposals`, so the proposals carried into the phase stop belonging to it), the `completed_at` stamp on the scheduled transition, and `current_state_id`.

Decisions worth calling out:

- **It deletes the history row rather than appending a compensating `B → A` row.** Phase membership is derived from transitions, and the most recent transition *into* a phase defines its inbound window (`getProposalsForPhase`). A compensating row would restart the returned-to phase's window at `now` with an empty attachment set and silently hide every proposal submitted before the reversal.
- **It fails closed on reviewer work.** Only assignments still at `PENDING` are deleted — the same `PRUNABLE_STATUS` rule `reconcileReviewAssignments` already uses. If a reviewer has started one, the reversal is rejected with a message naming the count rather than cascade-deleting their draft or submitted review.
- **Votes are kept.** `decisions_vote_submissions` is scoped to the instance rather than a phase, and its unique `(instance, voter)` constraint means a re-advance reuses the same ballots instead of double-counting them.

### Retiring the results run

The results record has to survive a reversal so there's an audit trail that a result occurred. But several reads treat "a successful `decision_process_results` row exists" as "this process is finished" — most sharply `listProposals`, which uses it to decide whether ballot queries may carry real `voteCount`s (#1207). Leaving the run live meant reverting out of the results phase republished a real tally into the re-opened vote, and `votedByProfileId` is client-settable, so that is reachable from the API and not just the results screen.

So the reversal retires the run: `success` goes false with an `errorMessage` explaining why. The row survives with its `executed_at`, counts and selections, so the audit trail is intact, and every "are results published?" read already keys on `success`. That means `listProposals` and `getLatestSelectionForProposal` need no change at all. `getSelectedProposalIds` gains the `success` filter it never had (it would otherwise report a failed run's selections as `isSelected`), and `getResults` / `getResultsStats` now skip unsuccessful runs and resolve to "no results" instead of throwing a raw `Error`.

An earlier revision of this PR added a `reverted_at` column for this. That made the feature unexercisable on a preview deployment — `services/db/migrate.ts` skips migrations for any Vercel preview branch that isn't `dev`, so the column never exists there and both the write and the reads filtering on it fail. No schema change is needed to express this, so there isn't one.

The scheduled transition is re-opened only while it is still future-dated. Re-opening an already-due row would hand it straight back to the date cron, which would re-advance within minutes and silently defeat the reversal.

Not reversible: the `phaseTransitioned` event the advance emitted has already gone out as member notifications. The confirm dialog says so.

`submitManualSelection` picked up the extracted `lockProcessInstance` helper — the instance `SELECT … FOR UPDATE` was duplicated between it and the new service.

Asana: https://app.asana.com/0/1213315744811204/1217366745512650
Multica: ONE-672


iwillig (approved):

packages/common/src/services/decision/getResults.ts
```diff
@@ -62,20 +62,21 @@ export const getLatestResultWithProposals = async ({
...
 
-  // Get the latest result (without loading all selections)
+  // Get the latest result (without loading all selections). Failed and
+  // reverted runs are kept as audit records but never describe the instance's
+  // current results, so they resolve to "no results" rather than an error.
   const result = await db._query.decisionProcessResults.findFirst({
-    where: eq(decisionProcessResults.processInstanceId, processInstanceId),
+    where: and(
+      eq(decisionProcessResults.processInstanceId, processInstanceId),
+      eq(decisionProcessResults.success, true),
```
scazan, line 71: This is a bit of an override to this functionality.. but to keep this simple I am using success: false for ones that are not the real ones. Will revisit a status down the road perhaps. Either way the reason for no success is recorded

PR #1774, opened 2026-08-13
feat(decisions): resolve proposal review screens from one proposal-keyed URL

valentin0h:
Clicking a review card's title and its reviewer count opened two different screens behind two different URL keys with opposite auth: reviewers could only open `/decisions/[slug]/reviews/[assignmentId]`, admins only `/decisions/[slug]/proposal/[profileId]/reviews`. This makes the proposal-keyed URL the one canonical screen, resolved per viewer: instance admins get the Review Progress summary, assigned reviewers get their own review (the latest assignment when several phases assigned one), everyone else gets `forbidden()` — anonymous SSR included, mapped to forbidden rather than a 500. The lookup is an optional `proposalProfileId` filter on the existing `listReviewAssignments`; with no explicit `phaseId` it skips the current-phase default so assignments pinned to earlier phases still resolve, ordered `assignedAt DESC NULLS LAST` in SQL. `/decisions/[slug]/reviews/[reviewId]` stays untouched for stable email links.

### DIFF
--- a/apps/app/src/components/decisions/ReviewAssignmentsList.tsx
+++ b/apps/app/src/components/decisions/ReviewAssignmentsList.tsx
@@ -69,8 +69,19 @@ export function ReviewAssignmentsList({
     parseAsStringLiteral(REVIEW_ASSIGNMENT_SORTS).withDefault('leastReviewed'),
   );
 
+  // cached from the page-level suspense query — no extra request
+  const [instance] = trpc.decision.getInstance.useSuspenseQuery({
+    instanceId: processInstanceId,
+  });
+  const phases = instance.instanceData?.phases ?? [];
+  const currentPhase = phases.find(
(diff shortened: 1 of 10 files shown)


apps/app/src/components/decisions/ReviewAssignmentsList.tsx
```diff
@@ -69,8 +69,19 @@ export function ReviewAssignmentsList({
     parseAsStringLiteral(REVIEW_ASSIGNMENT_SORTS).withDefault('leastReviewed'),
   );
 
+  // cached from the page-level suspense query — no extra request
+  const [instance] = trpc.decision.getInstance.useSuspenseQuery({
+    instanceId: processInstanceId,
+  });
+  const phases = instance.instanceData?.phases ?? [];
+  const currentPhase = phases.find(
```
scazan, line 77: Don't we have a util for this already?

PR #1783, opened 2026-08-13
fix(app): stop the process builder saving from inside a state updater

iwillig:
Split out of #1750 at review request — these were found while setting up a decision to export from, and were carried along rather than separated.

Both the phase editor and the review settings saved by calling into the autosave store from inside a state updater. Updaters must be pure and React runs them during render (twice in dev), so each edit wrote to that store mid-render and scheduled an update on other subscribers while the editing component was still rendering — the "cannot update a component while rendering a different component" warning, with the write itself able to run more than once per edit. Both now compute, set, then save.

Worth a reviewer's attention: each screen now reads its current value from the closure rather than the updater's argument, so two edits arriving in the same tick would lose one. That holds today because every caller changes a single field per user event, and both places say so — but it is the assumption this change rests on, and it is why these are better reviewed here than inside an export PR.

## LLM Usage

- opus

## Risk Justification

- Behaviour-preserving for single-field edits, which is every current caller; no API, schema, or data changes.
- The closure read replaces an updater's `prev`, so a future caller batching two field changes into one event would drop one. Called out in the code comments at both sites.

### DIFF
--- a/apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
+++ b/apps/app/src/components/decisions/ProcessBuilder/stepContent/general/PhaseDetailPage.tsx
@@ -150,15 +150,17 @@ function PhaseDetailForm({
   const allPhasesRef = useRef(allPhases);
   allPhasesRef.current = allPhases;
 
+  // The next phase is computed here rather than inside a `setPhase` updater:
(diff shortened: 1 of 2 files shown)


scazan (approved):
Looks good to me. Are we able to verify that this was an issue that was observed? It's a good fix either way but of course we should make sure to test a bunch for any regressions. Nice!

PR #1784, opened 2026-08-13
ci: probe the design-system job (throwaway, do not merge)

nourmalaeb:
Throwaway. Adds a temporary `pull_request` trigger so the `design-system` job from #1782 actually executes — `pull_request_target` reads the workflow from the base branch, so a new job can never run on the PR that introduces it.

Delete this branch and PR once the job is proven green.


nourmalaeb:
Probe complete — the design-system job is green on a fresh runner (run 31726154526). It caught a real failure first: @op/styles resolves to a generated dist/styles.css that does not exist on a clean checkout, fixed in #1782 by driving the build through turbo. Closing and deleting the branch.

PR #1794, opened 2026-08-14
fix(sense): isolate user-authored text direction in RTL

nourmalaeb:
Two unrelated quick fixes, split into their own commits.

- **RTL bidi.** User-authored text (proposal titles and previews, category tags, phase and category names, decision names) rendered without isolation, so its punctuation, digits, and the clamp ellipsis resolved against the *page* direction — an English title on an Arabic page showed as `??Hell yeah` instead of `Hell yeah??`. Blocks get `dir="auto"` + `[text-align:match-parent]`, inline names get `<bdi>`. Fixed in `@op/sense` where possible (`ProposalCard`, `ProfileItem`, `TagGroup`) so every card and list surface inherits it, plus the app spots that render raw elements.
- **Docs.** Both `CLAUDE.md` files now carry the rule. The two halves do different jobs and both are needed: `dir="auto"` takes the ordering and the ellipsis side from the *content*, `match-parent` keeps the alignment on the *page* — without it an English row in an Arabic list aligns left while its neighbours align right, which is what the active-decisions panel was doing.
- **Layout polish.** Platform highlights spacing and number size at small widths, decision action bar button widths, edit-banner button size, summary row gap.


nourmalaeb:
Superseded by four granular PRs rebased onto the current base branch: #2038 (sense bidi), #2039 (app-side bidi call sites), #2040 (mobile button widths), #2041 (platform highlights). The notification-panel commit is dropped, since the same fix already landed upstream.

PR #1798, opened 2026-08-14
fix(decisions): translate user content on proposal and review screens

iwillig:
Machine translation reached the decision overview but not the proposal list, the proposal page, or the rubric review screen, so a COWOP reviewer who reads Spanish had no way to review. Three independent causes:

- Language detection sampled only body text, so a proposal with a short or empty body produced no sample and the banner never rendered for a foreign-language title.
- The review screen sat outside every translation provider, and no endpoint could translate rubric criteria at all.
- The reviewer's queue tab rendered outside the provider its sibling tab mounts internally, so the two tabs disagreed.

Translation state lives in two React contexts — one at the decision-view layout, one inside `ProposalsList` — so a new proposal-card surface renders untranslated until it is wired in explicitly. That is why three screens drifted. Rubric translation moves display copy only: option `const` values, bounds and the `required` list stay as authored, so stored answers and validation are unchanged.

## Testing the preview build

Needs a decision whose proposal content is in a language other than your UI locale (switch locale with the header language chooser), plus a review phase with a rubric and at least one assignment. `DEEPL_API_KEY` must be set in the preview environment — translation throws without it.

1. **Proposal list** — on `/decisions/<slug>/current`, with a proposal that has a Spanish title and an empty or one-line body. The "Translate to English" banner appears. On `dev` it does not.
2. **Proposal page** — open that proposal and translate. Title, body and field labels switch; "Translated from Spanish · View original" puts them back.
3. **Review screen** — open `/decisions/<slug>/reviews/<assignmentId>`. One banner. Translating switches the proposal pane *and* the rubric criteria, descriptions and option labels. "View original" reverts both panes together.
4. **Answers survive translation** — while translated, pick a rubric option and type a note, then hit "View original". The selection and the note persist, and the form still submits.
5. **Reviewer queue** — in a review phase, "Proposals to review" offers the banner and translates its cards, matching the "Other proposals" tab.
6. **No regressions** — the decision overview still translates, and an English proposal viewed in English shows no banner.

## Follow-ups (out of scope)

- **Snapshot divergence:** the review pane renders the assignment's pinned proposal snapshot, while `translation.translateProposal` fetches the live proposal. Resubmission re-anchors only the requesting reviewer's assignment, so on a multi-reviewer phase a second reviewer can be shown a translation of newer text than the original beside it. Translate the assignment payload rather than the proposal id.
- **Peer reviews:** the "Other reviews" tab renders a separately-fetched rubric template plus reviewer-authored notes, which have no translation path at all.

Other untranslated surfaces (comments, results, ballot, vote review, review summary) are listed on the Asana task.

Asana: https://app.asana.com/1/1108348036672214/project/1211387328265612/task/1217153730617467

### DIFF
--- a/apps/app/src/components/decisions/Review/ReviewLayout.tsx
+++ b/apps/app/src/components/decisions/Review/ReviewLayout.tsx
@@ -16,6 +16,7 @@ import { ReviewNavbar } from './ReviewNavbar';
 import { ReviewProposalPane } from './ReviewProposalPane';
 import { ReviewRubricForm } from './ReviewRubricForm';
 import type { PreviousReviewPhase } from './ReviewTabs';
+import { ReviewTranslationProvider } from './ReviewTranslationContext';
 
 interface ReviewLayoutProps {
   decisionSlug: string;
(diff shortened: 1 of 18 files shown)


apps/app/src/components/decisions/Review/ReviewLayout.tsx
```diff
@@ -90,27 +91,34 @@ export async function ReviewLayout({
...
+            both come from the assignment that provider loads. */}
+        <ReviewTranslationProvider assignmentId={assignmentId}>
+          <div className="flex h-dvh flex-col overflow-hidden bg-white">
+            <ReviewNavbar decisionSlug={decisionSlug} />
 
-          <SplitPane className="mx-auto max-w-6xl" defaultMobileTabId="review">
-            <SplitPane.Pane
-              id="proposal"
-              label={<TranslatedText text="Proposal" />}
+            <SplitPane
```
scazan, line 100: I wonder why the formatting shifted here. It probably should've been caught in CI if there was an issue prior or here. Obv. not a blocker but curious.

apps/app/src/components/decisions/Review/ReviewTranslationContext.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const [translated, setTranslated] = useState<{
+    proposal?: TranslatedFields;
+    rubric?: TranslatedFields;
+    sourceLocale: string;
+  } | null>(null);
+
+  // Flipped true on Translate, false on View Original. A response that lands
+  // after the reviewer reverts checks this before writing state, so the revert
+  // can't be undone by an in-flight request (as in useTranslateDecision).
+  const translatingRef = useRef(false);
```
scazan, line 85: Check useTranslateDecision for similar logic. We can definitely pull out some of the duplication to DRY it up a bit. Not a blocker though and we can ship without it for now since it's a bug

scazan (approved):
Nice

PR #1799, opened 2026-08-14
test(e2e): pin the UGC translation gaps on proposal and review screens

iwillig:
A COWOP facilitator reported a translation failure. The browse list, the proposal page, and the rubric review screen keep the source language. The decision overview does translate. A non-English reviewer therefore cannot read the proposals that the process assigns to them.

The suite covered no part of translation. The report had no reproduction, and a fix could not prove itself.

These tests pin two of the reported failures. Each test also asserts a positive control, so a missing Translate button cannot pass a test for the wrong reason:

- A reviewer opens a Spanish proposal on the rubric review screen and gets no way to translate it. The same decision's overview shows the Translate button.
- A proposal carries a Spanish title and no body. It offers no translation. A second proposal carries a Spanish body and offers the Translate button.

Worth a reviewer's attention: these tests assert the current broken behaviour. The engineer who fixes these gaps must invert them. This is deliberate. The tests then fail when the bug goes away.

These tests skip two more reported symptoms. Both symptoms need a complete translation round trip, which calls DeepL from the suite. The review phase hero ignores its translated headline. A list translation disappears when the reader opens a proposal.


iwillig:
Notes to address

1. Use the service layers in the e2e tests.

